PR #6, opened 2020-02-07
Add testing

jwhonce:
- Add Makefile and targets: lint and unittest
 - unittest target will fail if test coverage falls below 50%
   - This number will be increased into the 90% ASAP
 - pylint configuration for project included
 - Source updated to eliminate all pylint Errors
 - Currently pylint Warning are reported but not gating
 - Renamed client to api_connection to align with HTTPConnection
 - Tests added as samples and to obtain 55% coverage


Makefile
```diff
@@ -0,0 +1,53 @@
...
+PODMAN_VERSION ?= '1.80'
+
+.PHONY: podman-py
+podman-py:
+	PODMAN_VERSION=$(PODMAN_VERSION) \
+	$(PYTHON) setup.py sdist bdist
+
+.PHONY: env
+env:
+	dnf install python3-coverage
```
rhatdan: Should we add a `-y` to these

Makefile
```diff
@@ -0,0 +1,53 @@
...
+	dnf install pylint
```
rhatdan: Ditto

Makefile
rhatdan: Also should we support apt versions for this, probably do this in a different PR.

baude: im actually wondering if we should detect the presense of python3-coverage and bail if not?

Makefile
```diff
@@ -0,0 +1,53 @@
+PYTHON ?= $(shell command -v python3 2>/dev/null || command -v python)
+DESTDIR ?= /
+PODMAN_VERSION ?= '1.80'
```
cevich: Make this `export PODMAN_VERSION ?= 1.80`

Makefile
```diff
@@ -0,0 +1,53 @@
...
+
+.PHONY: podman-py
+podman-py:
+	PODMAN_VERSION=$(PODMAN_VERSION) \
```
cevich: Delete this line (export directly from Makefile does same thing)

Makefile
cevich: I'm very much against installing system-wide stuff from a makefile.  It presumes a lot about the environment it's running under, and that the developer even wants their system messed with for them :grinning:

A better approach would be to stick this stuff into a shell script.  That can be used by both humans and machines as needed.

Makefile
```diff
@@ -0,0 +1,53 @@
...
+.PHONY: uninstall
+uninstall:
+	$(PYTHON) -m pip uninstall --yes podman ||:
+
+.PHONY: clean
+clean:
+	rm -rf podman.egg-info dist
+	find . -depth -name __pycache__ -exec rm -rf {} \;
+	find . -depth -name \*.pyc -exec rm -f {} \;
+	$(PYTHON) ./setup.py clean --all
```
cevich: I would recommend prefixing all those lines with `-` so make treats them as non-fatal if any fail.  Otherwise an early failure can block complete cleanup.

podman/api_connection.py
```diff
@@ -0,0 +1,108 @@
...
+from http.client import HTTPConnection
+
+import podman.errors as errors
+import podman.images as images
+import podman.system as system
+
+
+class ApiConnection(HTTPConnection, AbstractContextManager):
+    """ ApiConnection provides a specialized HTTPConnection
+        to a Podman service."""
```
cevich: Howcome the extra whitespace on the left-hand side?  (new to me)

podman/api_connection.py
```diff
@@ -0,0 +1,108 @@
+""" Provides a Connection to a Podman service. """
+import socket
+import urllib.parse
```
cevich: Am I miss-remembering, or isn't `requests` much better to use?

Makefile
jwhonce: @rhatdan Done
@baude Done
@cevich Note that target has no dependencies so it will only run if called.

Makefile
jwhonce: @cevich the -f on the rm's ensures the same.  The last item should cause a failure if it fails.

podman/api_connection.py
jwhonce: It was not obvious to me how to do ssh tunneling etc with `requests`.  Seems too high level.

podman/api_connection.py
jwhonce: Making the docstring left aligned? Easier to read.

rhatdan:
LGTM

Makefile
```diff
@@ -0,0 +1,53 @@
+PYTHON ?= $(shell command -v python3 2>/dev/null || command -v python)
+DESTDIR ?= /
+export PODMAN_VERSION ?= '1.80'
```
TomSweeneyRedHat: It would be nice if we could manage to grab the version number from somewhere.  But that's past my Makefile skills.

PR #10, opened 2020-02-08
[WIP] Implement new API

nick-lehmann:
Currently highly work in progress. Migrate code from `docker-py` to `podman-py`.

Signed-off by: [nick@lehmann.sh](mailto:nick@lehmann.sh)


podman/images/collection.py
```diff
@@ -0,0 +1,278 @@
+class ImageCollection(Collection):
+    model = Image
+
+    def build(self, **kwargs):
+        """
+        Build an image and return it. Similar to the ``docker build``
```
rhatdan: s/docker/podman/g

podman/images/api.py
```diff
@@ -0,0 +1,580 @@
...
+
+log = logging.getLogger(__name__)
+
+
+class ImageApiMixin(object):
+
+    @utils.check_resource('image')
+    def get_image(self, image, chunk_size=DEFAULT_DATA_CHUNK_SIZE):
+        """
+        Get a tarball of an image. Similar to the ``docker save`` command.
```
rhatdan: s/docker/podman/g

podman/images/api.py
```diff
@@ -0,0 +1,580 @@
...
+
+        Example:
+
+            >>> image = cli.get_image("busybox:latest")
+            >>> f = open('/tmp/busybox-latest.tar', 'wb')
+            >>> for chunk in image:
+            >>>   f.write(chunk)
+            >>> f.close()
+        """
+        res = self._get(self._url("/images/{0}/get", image), stream=True)
```
rhatdan: We should be using the libpod API not the legacy API.

podman/images/api.py
```diff
@@ -0,0 +1,580 @@
...
+        Show the history of an image.
+
+        Args:
+            image (str): The image to show history for
+
+        Returns:
+            (str): The history of the image
+
+        Raises:
+            :py:class:`docker.errors.APIError`
```
rhatdan: s/docker/podman/g

podman/images/api.py
```diff
@@ -0,0 +1,580 @@
...
+                If the server returns an error.
+        """
+        res = self._get(self._url("/images/{0}/history", image))
+        return self._result(res, True)
+
+    def images(self, name=None, quiet=False, all=False, filters=None):
+        """
+        List images. Similar to the ``docker images`` command.
```
rhatdan: s/docker/podman/g

podman/images/api.py
```diff
@@ -0,0 +1,580 @@
...
+                Available filters:
+                - ``dangling`` (bool)
+                - `label` (str|list): format either ``"key"``, ``"key=value"``
+                    or a list of such.
+
+        Returns:
+            (dict or list): A list if ``quiet=True``, otherwise a dict.
+
+        Raises:
+            :py:class:`docker.errors.APIError`
```
rhatdan: s/docker/podman/g

podman/images/api.py
```diff
@@ -0,0 +1,580 @@
...
+        res = self._result(self._get(self._url("/images/json"), params=params),
+                           True)
+        if quiet:
+            return [x['Id'] for x in res]
+        return res
+
+    def import_image(self, src=None, repository=None, tag=None, image=None,
+                     changes=None, stream_src=False):
+        """
+        Import an image. Similar to the ``docker import`` command.
```
rhatdan: s/docker/podman/g

podman/images/api.py
```diff
@@ -0,0 +1,580 @@
...
+
+        Args:
+            src (str or file): Path to tarfile, URL, or file-like object
+            repository (str): The repository to create
+            tag (str): The tag to apply
+            image (str): Use another image like the ``FROM`` Dockerfile
+                parameter
+        """
+        if not (src or image):
+            raise errors.DockerException(
```
rhatdan: s/docker/podman/g

podman/images/api.py
```diff
@@ -0,0 +1,580 @@
...
+            if stream_src:
+                headers['Transfer-Encoding'] = 'chunked'
+            return self._result(
+                self._post(u, data=src, params=params, headers=headers)
+            )
+
+    def import_image_from_data(self, data, repository=None, tag=None,
+                               changes=None):
+        """
+        Like :py:meth:`~docker.api.image.ImageApiMixin.import_image`, but
```
rhatdan: s/docker/podman/g

podman/images/api.py
```diff
@@ -0,0 +1,580 @@
+"""
+Taken from docker.api.images
```
rhatdan: sed 's/Taken from/Based on/g'

podman/images/collection.py
```diff
@@ -0,0 +1,278 @@
...
+                If the server returns an error.
+        """
+        return RegistryData(
+            image_name=name,
+            attrs=self.client.api.inspect_distribution(name, auth_config),
+            client=self.client,
+            collection=self,
+        )
+
+    def list(self, name=None, all=False, filters=None):
```
jwhonce: List is a python reserved word.  This should be something list list_images.

nick-lehmann: That`s true. Hoewever, method names cannot clash with reserved words. The `list` method here can only be addressed using the dot notation from an instance of the class. Therefore, there should be no case in which anybody could confuse this method and the builtin `list`.

nick-lehmann:
Thanks for the reviews @rhatdan, @jwhonce 👍🏻 

My first commit was just a plain copy of the docker code and I had fixed the issues immediately after that. I committed only now because I was unable to test the code. I first had to compile the development version of `libpod` and get the new rest service up and running. @baude was so kind as to get me started on that. In my latest commit I have ported more of the `docker-py` code. 

@jwhonce is this structure of the code right in your opinion? I have kept the interface of `docker-py` to make it easier for people using the docker SDK already to switch to podman.

- Models are Python representation of the single resources on the server (e.g. images, containers, etc)
- Collections represent all entities of a single resource and can be used through `client.images`, `client.containers` etc.
- The code for talking to the API is kept separate

Those three things are located in a single python package as they are logically grouped together. You can see this structure in `podman.resources.images`. This is the endpoint that I have ported at the moment and tested the `get()`, `list()` and `prune()` method which all work as expected.

Furthermore, I have removed everything related to python2, including the use of the `six` compatibility layer and changed all type annotation to the native type annotation provided by python3.

My biggest questions at the moment are how authentication should be handled and which protocols should be used to talk to the API. In `podman.transport` there are all the adapters for the single protocols. They contain a lot of little details which I considered worth keeping as they seem not trivial. However, keeping this code would introduce a dependency on the `request` library.

Furthermore, I would need advice on what to do with `auth`, `credentials` and `tls`. How much of this applies to the podman remote API right now or in the near future? The API documentation does not say anything about those topics but I would consider them important.

After you give me some insights on how to handle the things described above, I will also fix the checks and clean up `podman.errors`. Currently, it contains the exceptions you have defined as well as all exceptions from `docker-py`.

Just another quick question: What is the equivalent to a `Dockerfile`? Do you call it a `Podmanfile` or something more generic?

rhatdan:
@nick-lehmann Can you fix the gating issues?

nick-lehmann:
@rhatdan I will very shortly. I just wanted to clarify what is actually needed from all the docker code so I don't have to fix code that will not be used later on. But all those issues are quite minor and should be easy to fix.

nick-lehmann:
Nevermind. The log for why the gating task of my previous commit failed only showed a subset of all problems. I have resolved all of them but there are actually a lot more tiny details like missing docstrings or one-letter variables. If you give me advice on what of the ported code is really necessary, I am happy to fix those issues for everything that has not been thrown out 👍🏻

rhatdan:
I think that is up to @jwhonce and @baude to decide.

rhatdan:
Personally I think making this library close to docker-py, is a good idea, just to allow users to more easily convert their existing Python apps to use podman, but I also understand that the docker-py community might have made decisions that we disagree with.

rhatdan:
@nick-lehmann Are you still working on this?

nick-lehmann:
@rhatdan No, currently not, as I am still waiting for feedback from either @jwhonce or @baude. I would really like to continue implementing, but I am uncomfortable putting time into migrating code which might be scrapped later.

jwhonce:
> Thanks for the reviews @rhatdan, @jwhonce 👍🏻
> 
> My first commit was just a plain copy of the docker code and I had fixed the issues immediately after that. I committed only now because I was unable to test the code. I first had to compile the development version of `libpod` and get the new rest service up and running. @baude was so kind as to get me started on that. In my latest commit I have ported more of the `docker-py` code.
> 
> @jwhonce is this structure of the code right in your opinion? I have kept the interface of `docker-py` to make it easier for people using the docker SDK already to switch to podman.

I like your idea.  I initially was going the same direction we did in python-podman, but I believe we'll get more traction with your ideas. 

> 
> * Models are Python representation of the single resources on the server (e.g. images, containers, etc)
> * Collections represent all entities of a single resource and can be used through `client.images`, `client.containers` etc.
> * The code for talking to the API is kept separate
> 
> Those three things are located in a single python package as they are logically grouped together. You can see this structure in `podman.resources.images`. This is the endpoint that I have ported at the moment and tested the `get()`, `list()` and `prune()` method which all work as expected.
> 
> Furthermore, I have removed everything related to python2, including the use of the `six` compatibility layer and changed all type annotation to the native type annotation provided by python3.

We need support for python 3.6 +

> 
> My biggest questions at the moment are how authentication should be handled and which protocols should be used to talk to the API. In `podman.transport` there are all the adapters for the single protocols. They contain a lot of little details which I considered worth keeping as they seem not trivial. However, keeping this code would introduce a dependency on the `request` library.

The go code does uses an ssh proxy to authenticate to the server. Which works well.  I haven't looked into doing that in Python yet.  When I reviewed the `request` library. I didn't find methods for low level connections.  Maybe I missed them.

> 
> Furthermore, I would need advice on what to do with `auth`, `credentials` and `tls`. How much of this applies to the podman remote API right now or in the near future? The API documentation does not say anything about those topics but I would consider them important.

We try not to expose the podman service to the internet by default.  So tls is not supported.  Rather we do a ssh tunnel to the unix domain socket on the host. Is that the auth/credentials concerning you? Or, from the podman service to the registries? e.g., podman login?

> 
> After you give me some insights on how to handle the things described above, I will also fix the checks and clean up `podman.errors`. Currently, it contains the exceptions you have defined as well as all exceptions from `docker-py`.
> 
> Just another quick question: What is the equivalent to a `Dockerfile`? Do you call it a `Podmanfile` or something more generic?

We use Dockerfile by default, to help ease the transition.

nick-lehmann:
> I like your idea. I initially was going the same direction we did in python-podman, but I believe we'll get more traction with your ideas.

Thank you very much. Then I can now continue working on the other endpoints 👍🏻

> The go code does uses an ssh proxy to authenticate to the server. Which works well. I haven't looked into doing that in Python yet. When I reviewed the request library. I didn't find methods for low level connections. Maybe I missed them.

Ah ok, I see. I have found the [POC](https://github.com/containers/libpod/blob/master/cmd/cli/main.go) in the podman client where you do that. Right off the bat, I do not know how to implement this in python but I will have a look at that. However, I guess that the `requests` library does not have this functionality because it is a very high-level wrapper that should work in everyday scenarios and this is probably not one of them. Since the wide variety of connection protocols offered by `docker-py` is not needed here, we can scrap all the adapters they have written and I could go back to using the builtin `http` module.

> We try not to expose the podman service to the internet by default. So tls is not supported. Rather we do a ssh tunnel to the unix domain socket on the host. Is that the auth/credentials concerning you? Or, from the podman service to the registries? e.g., podman login?

Alright. Then I can remove everything related to `tls`. And yes, it is about everything inside `auth.py` and the `credentials` module which is the part concerned with finding and authenticating against a registry.

rhatdan:
BTW Buildah and Podman build now support Containerfile syntax, which looks a lot like Dockerfile...

nick-lehmann:
Just a quick status update. The thing that concerned me the most is connecting to a remote unix socket via an ssh tunnel. It is not a standard thing to do (at least I haven't seen it in python) and I have found no ssh library for python that does that. Features requests like [this for paramiko](https://github.com/paramiko/paramiko/issues/544) or [this one for sshtunnel](https://github.com/pahaz/sshtunnel/issues/113) have remained uncommented.

However, setting up a socket forward using the command line was pretty straightforward using just `ssh -fN -L./podman.sock:/run/podman/podman.sock root@server` so I used this by creating a subprocess.

Luckily, it seems like `docker-py` allows the user to connect to a remote docker engine via ssh (even if it is not documented). I have tried it out and it works. However, I have not figured out how the code behind this works and will explore it more tomorrow.

jwhonce:
@nick-lehmann Great update.  We used the ssh binary type tunnel in the varlink API and found users had a heck of a time getting that setup.  That is why I went to the ssh proxy style for the RESTful API.  Fingers crossed we can get Python to support that style of connection.

nick-lehmann:
@jwhonce I have found out how the docker sdk is making its requests over ssh. It connects to the host via ssh and then [opens a repl](https://github.com/docker/docker-py/blob/030af62dca982e62ed418314c857da1f90ce4282/docker/transport/sshconn.py#L34) using `docker system dial_stdio`. The docker source code for this command can be found [here](https://github.com/docker/cli/blob/master/cli/command/system/dial_stdio.go). Then, it pastes the HTTP requests inside and the corresponding HTTP response will be printed back at the console.

To me, there are a couple options on how to continue:

- Follow the approach of docker described above. However, this option would require an appropriate command in `libpod`.

- Connect to the server via ssh and then start a repl using `netcat`. This is described in [this article](https://skife.org/go/2013/02/08/rpc_with_ssh_and_domain_sockets.html). This would require `netcat` to be installed on the server which it is not by default.

- Connect to the server and implement a repl using plain `sh` or other programs which are installed by default. Sadly, I have no clue how to do this.

- Use the native `ssh` client to forward the socket to the client, e.g. to `/tmp/podman.sock` and send all requests through this. I feel this might be a little bit fragile but is the fastest option.

nick-lehmann:
@jwhonce did you have any time to look at my latest comments? An answer to how ssh and authentication should be handled would grealty help me to continue working on this 👍🏻

rhatdan:
@nick-lehmann @jwhonce What is the state of this PR?

nick-lehmann:
@rhatdan I haven't worked on this since my last comment as I am still waiting for feedback from @jwhonce on how to do the SSH connection. I found no way to tunnel the SSH connection like he did in his proof-of-concept for the go client and outlined possible solutions/workarounds above.

edeediong:
@nick-lehmann maybe since there's no library that can fix this now, maybe the Python subprocess library can be used to run the commands on terminal or maybe use the os library to do same. The os module has a function called  `system` called by `os.system`. You stated [here](https://github.com/containers/podman-py/pull/10#issuecomment-591137267) that you used subprocess, can I understand why you maybe didn't continue with that?

nick-lehmann:
@edeediong sorry for the late response...

While `subprocess` is indeed an option, it can be quite unstable across multiple platforms. You would need to know the right cli command for each operating system which might also vary depending on the configuration of the users' machine. Jhon Honce has confirmed this in [this comment above](https://github.com/containers/podman-py/pull/10#issuecomment-591494927), the guys over at docker decided against this and personally, I also did not have a great experience with using subprocesses for things like this.

edeediong:
@nick-lehmann thanks for getting back to me. I am actually familiar with using subprocess. But you're right, the configuration may change for all users and the distros of Linux may change too. However, I doubt Python will introduce ssh proxy style anytime soon.

Issue #17, opened 2020-05-21
example in the README cannot run


parmentelat:
`README.md` shows an example that is broken

symbols like `system` and `images` must be either imported individually, or qualified as `podman.system`
plus, `ApiConnection` is mispelled as `APIConnection`

rhatdan:
Is this issue still true?

parmentelat:
I could not say as of now, but I did take care of that in #27 at first, so if/once it is rebased and merged this particular issue should vanish

jwhonce:
See https://pypi.org/project/podman-py/, README.md updated to use new PodmanClient()

PR #18, opened 2020-05-21
fix README.md so that the example code can run

parmentelat:
fix for #17 

just from reading it, I cannot yet test it

if you'd prefer I can have it the other way around, still `import podman` but qualify the symbols as they appear


rhatdan:
LGTM

rhatdan:
@jwhonce PTAL

parmentelat:
I close this one to create another, more extensive PR #20

PR #24, opened 2020-05-28
refactor of system/__init__.py, add unittests

bmarlow:
Refactored system to have call coverage based on the API docs
***some of these calls are not currently functional (because the API doesn't exist yet) and are noted in the code***

renamed the version call to ping as there is actually a real (planned) version endpoint

Also added some better test coverage for this.

### DIFF
--- a/podman/api_connection.py
+++ b/podman/api_connection.py
@@ -81,7 +81,7 @@ class ApiConnection(HTTPConnection, AbstractContextManager):
             )
         return response
 
-    def make_call(api, endpoint, method='GET', body=None, **kwargs):
+    def make_call(self, api, endpoint, method='GET', body=None, **kwargs):
         """A helper function to keep things DRY"""
         path = api.join(endpoint, kwargs)
         headers = {}
--- a/podman/api_connection.py
+++ b/podman/api_connection.py
@@ -81,15 +81,15 @@ class ApiConnection(HTTPConnection, AbstractContextManager):
             )
         return response
 
-    def make_call(api, endpoint, method='GET', params=None, body=None):
+    def make_call(api, endpoint, method='GET', body=None, **kwargs):
         """A helper function to keep things DRY"""
-        path = api.join(endpoint, params)
+        path = api.join(endpoint, kwargs)
         headers = {}
         data = None
         if method == 'POST':
             headers['Content-type'] = 'application/x-www-form-urlencoded'
             data = urllib.parse.urlencode(body)
-        return api.request(method, endpoint, body=data, headers=headers)
+        return api.request(method, path, body=data, headers=headers)
 
     def join(self, path, query=None):
         """Create a service URL.  Join base + path + query parameters"""
--- a/podman/api_connection.py
+++ b/podman/api_connection.py
@@ -81,6 +81,16 @@ class ApiConnection(HTTPConnection, AbstractContextManager):
             )
         return response
 
+    def make_call(api, endpoint, method='GET', params=None, body=None):
+        """A helper function to keep things DRY"""
+        path = api.join(endpoint, params)
+        headers = {}
+        data = None
+        if method == 'POST':
+            headers['Content-type'] = 'application/x-www-form-urlencoded'
+            data = urllib.parse.urlencode(body)
+        return api.request(method, endpoint, body=data, headers=headers)
+
     def join(self, path, query=None):
         """Create a service URL.  Join base + path + query parameters"""
         path = self.base + path
--- a/podman/containers/__init__.py
+++ b/podman/containers/__init__.py
@@ -272,6 +272,7 @@ def generate_kube_yaml_file(api, name, params=None):
     path = "/generate/{}/kube".format(name)
     return json.loads(make_call(api, path, params=params))
 
+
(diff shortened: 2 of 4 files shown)


rhatdan:
LGTM
@jwhonce @baude @mheon @vrothberg @giuseppe PTAL

podman/containers/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+import json
+import logging
+
+import podman.errors as errors
+
+
+def make_call(api, endpoint, method='GET', params=None, body=None):
+    """A helper function to keep things DRY"""
+    path = api.join(endpoint, params)
+    print(path)
```
mwhahaha, line 11: should drop this print

podman/containers/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+    urlify = api.quote(path)
+    response = api.request(method, urlify)
```
mwhahaha, line 13: body is unused? It seems like we should be using body in combination with {'Content-type': 'application/x-www-form-urlencoded'} to send the parameters when method='POST'

podman/containers/__init__.py
```diff
@@ -0,0 +1,315 @@
...
+    return response.read()
+
+
+def commit(api, params):
```
mwhahaha, line 17: I'm torn because I don't think we should just accept params and make the end user know what needs to get passed in for these functions, but at the same time that means we need to follow the podman api closely to ensure the params are exposed.  Should we maybe use **kwargs instead so we can support a dynamic list of parameters and pass them into make_call via body with application/x-www-form-urlencode so that you can write something like....

```
def make_call(api, endpoint, method='GET', params=None, body=None):
    path = api.join(endpoint, params)
    headers = {}
    data = None
    if method == 'POST':
        headers['Content-type'] = 'application/x-www-form-urlencoded'
        data = urllib.parse.urlencode(body)
    return api.request(method, endpoint, body=data, headers=headers)
    
def commit(api, **kwargs):
    make_call(api, "/commit", "POST", body=kwargs)
 
def container_logs(api, name, **kwargs):
    make_call(api, "/containers/{}/logs".format(name), params=kwargs)

commit(api, container='mycontainer', repo='foo', tag='bar', ...)
```

I might propose an an alternative implementation for this. I do think we need something like make_call to handle the post vs get logic since there's likely a limit for url query params (1024 or 2048 chars?) so we won't want to always use query params

podman/containers/__init__.py
```diff
@@ -0,0 +1,315 @@
+"""images provides the operations against images for a Podman service."""
+import json
+import logging
+
+import podman.errors as errors
+
+
+def make_call(api, endpoint, method='GET', params=None, body=None):
```
mwhahaha, line 8: Maybe this should go on the api class instead.

mwhahaha (changes requested):

rhatdan:
@mwhahaha Since you have taken the lead on this, I am willing to respect your judgement along with @jwhonce.

podman/containers/__init__.py
bmarlow: whoops, rookie move there.  this has been removed.

podman/containers/__init__.py
bmarlow: Fair point, it has been so long since I looked at this code I'm not quite sure why that is...  Fixed based on your subsequent comments.

podman/containers/__init__.py
bmarlow: Put your feedback for the make_call into the code, specifically the **kwargs stuff.

podman/containers/__init__.py
bmarlow: That is logical.  Moved to api class and made the appropriate reference updates.

bmarlow:
Thanks for the feedback.  Its been a while since I looked at this.  Took your feedback and integrated into the code as made since.  The latter couple of commits were just to fix the gating issues (since I forgot to lint locally).

bmarlow:
@mwhahaha please give another look, requested changes have been made.

test/unit/test_020_system.py
```diff
@@ -0,0 +1,37 @@
+import unittest
```
mwhahaha, line 1: We added a podman/tests/unit/system/test_system.py now to cover these. I think you need to rebase and bring your changes current with what we've landed over the last few days.

podman/api_connection.py
```diff
@@ -81,6 +81,16 @@ def request(self,
             )
         return response
 
+    def make_call(self, api, endpoint, method='GET', body=None, **kwargs):
```
mwhahaha, line 84: Check the current state of the apiconnection, we have a few helper functions specifically for post/get/delete

mwhahaha (changes requested):
I think you need to rebase on the current bit of master. We added get/post/delete functions on the api connection object that you can use instead of the make_call you had here.  Additionally it might be a good idea to squash your changes into a single one and just force push back into your branch.

podman/api_connection.py
bmarlow: Fair enough.  A little frustrating that this sat for 5+ months, but now I have to rebase...


I'll close this PR and resubmit.

PR #26, opened 2020-07-17
Organizing Responsibilities between Clients, Endpoints and Services

velvetkeyboard:
1. "ApiConnection" class was divided into 2 major classes:

  1.a podman.http.HTTPConnectionSocket: This one will be only
      responsible to establish and maintain the connection with the
      Podman socket.

  1.b podman.clients.lib.Client: This one will be responsible to now how
    to send/receive the connection through the HTTPConnectionSocket and
    which socket we should use with it.
    We also could encapsulate the steps to authenticate if needed.

  1.c From the base Client classes we have two clients out-of-the-box:
    Podman and PodmanRootless. The first is aimed to anyone running this
    library on a system-wide scope the second is for the ones using it
    on a more restricted environment (maybe a local dev machine with a
    virtual env). This will allow program to easily switch back and
    forth with little to none effort:

      def get_client():
          if os.environ.get('ENV') == 'DEV':
              return PodmanRootless()
          return Podman()

2. Endpoints are designed to encapsulate any common routines related to
   the requests on the ReST API scope. If you need a fixed injection at
   the headers or readjust the format of certain query params or data
   for POST. That should be handled here.
   To make it work properly we naturally create two main groups of
   classes using the podman.endpoints.lib.Endpoint class.

   2.a API Roots: These classes will extend from the Endpoint base class
       and will centralized fixed values and routines for an specific
       version of the API. You can see examples at the bottom of
       podman/endpoints/lib module.

   2.b Actual Endpoints: These classes are easy to create or spot. We
       would expect theses classes to only have at most two fields:
         - path
         - headers
       That should be enough to describe what that specific endpoint
       need to work.

  Another aspect of Endpoints is that they return the full HTTP response
  object. So this ones here are design to scope that requires validation
  of values at that level. Example: Web server pulling list of images
  will probably show the same page for a empty response but would be
  nice to distinguish between a regular 404 from a 500 so we can notify
  Sentry about a possible outrage on Podman. (not best example)

3. Services do almost the same as the Endpoint classes but present a
   high-level abstraction to get the JSON data (as dict). Here the user
   will deal with either success request returning a JSON (dict) or
   specially crafted Exceptions.

   Right now all the examples on the code are simple but we could:

   3.a Create a new class: GetOnlyFedoraImages. Inherited from the
       GetAllImages class we could easily override the json method and
       filter locally for images where the tag contains "fedora" in it.

   3.b Create a new class: GetOnlySmallImages. Inherited from the
       GetAllImages class we could easily override the json method and
       filter locally for images where the size needs to smaller than
       200 MB

  I think that this kind of level of abstraction might be more suited
  for a second library consuming this one but maybe we could thought
  about some handy services.


rhatdan:
@jwhonce PTAL

PR #29, opened 2020-10-04
Typos style

danvergara:
## Description

The purpose of this PR is improve the code quality. At this time I used [black](https://github.com/psf/black) as code formatter, keeping the maximum length of the line as 79.

# Changes
Just 4 files were modified:

* podman/api_connection.py
* podman/system/__ init __.py
* podman/images/__ init __.py
* podman/errors/__ init __.py

### DIFF
--- a/podman/api_connection.py
+++ b/podman/api_connection.py
@@ -88,8 +88,12 @@ class ApiConnection(HTTPConnection, AbstractContextManager):
         data = params
         if not headers:
             headers = {}
-        if 'content-type' not in headers and params:
-            headers['content-type'] = 'application/x-www-form-urlencoded'
+
+        if encode:
+            if "content-type" not in set(key.lower() for key in headers) and params:
+                headers["content-type"] = "application/x-www-form-urlencoded"
+
+            data = urllib.parse.urlencode(params)
         return self.request('POST',
                             self.join(path),
                             body=data,
--- a/podman/api_connection.py
+++ b/podman/api_connection.py
@@ -14,19 +14,25 @@ import podman.system as system
 
 
 class ApiConnection(HTTPConnection, AbstractContextManager):
-    """ApiConnection provides a specialized HTTPConnection
-    to a Podman service."""
+    """
+    ApiConnection provides a specialized HTTPConnection
+    to a Podman service.
+    """
 
-    def __init__(self, url, base="/v1.24/libpod", *args, **kwargs):  # pylint: disable-msg=W1113
+    def __init__(
+        self, url, base="/v1.24/libpod", *args, **kwargs
+    ):  # pylint: disable-msg=W1113
         if url is None or not url:
             raise ValueError("url is required for service connection.")
 
         super().__init__("localhost", *args, **kwargs)
         supported_schemes = ("unix", "ssh")
         uri = urllib.parse.urlparse(url)
-        if uri.scheme not in ("unix", "ssh"):
+        if uri.scheme not in supported_schemes:
             raise ValueError(
-                "The scheme '{}' is not supported, only {}".format(uri.scheme, supported_schemes)
+                "The scheme '{}' is not supported, only {}".format(
+                    uri.scheme, supported_schemes
+                )
             )
         self.uri = uri
         self.base = base
@@ -38,7 +44,9 @@ class ApiConnection(HTTPConnection, AbstractContextManager):
             sock.connect(self.uri.path)
             self.sock = sock
         else:
-            raise NotImplementedError("Scheme {} not yet implemented".format(self.uri.scheme))
+            raise NotImplementedError(
+                "Scheme {} not yet implemented".format(self.uri.scheme)
+            )
 
     def delete(self, path, params=None):
         """Basic DELETE wrapper for requests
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@jwhonce @baude PTAL

jwhonce:
@danvergara CI seems unhappy let me take a look.  I'm okay with using the `black` formatter as long as no-one has objections.

danvergara:
Thanks for the response. Yes, something is failing, but I don't know why.It seems the problem is related with /usr/local/bin/entrypoint.sh, but I don't know what it does.

rhatdan:
I am fine with the formatter.
@cevich PTAL

cevich:
looking.

cevich:
It seems automation here isn't even using VMs, so the failing task and lots of other stuff can simply go away: #31

jwhonce:
@danvergara Please rebase, thanks

danvergara:
@jwhonce yes sir! Thanks!

rhatdan:
@mwhahaha PTAL

podman/api_connection.py
```diff
@@ -13,20 +13,26 @@
 
 
 class ApiConnection(HTTPConnection, AbstractContextManager):
-    """ ApiConnection provides a specialized HTTPConnection
-        to a Podman service."""
-
-    def __init__(self, url, base="/v1.24/libpod", *args,
-                 **kwargs):  # pylint: disable-msg=W1113
+    """
```
mwhahaha, line 16: I've been using rst example doc strings. I guess we should establish a standard. I wouldn't add the line break here and leave it as it.  could remove the leading space to be consistent with everything else.

podman/api_connection.py
```diff
@@ -13,20 +13,26 @@
...
+    ApiConnection provides a specialized HTTPConnection
+    to a Podman service.
+    """
+
+    def __init__(
+        self, url, base="/v1.24/libpod", *args, **kwargs
```
mwhahaha, line 22: I don't think we should format function function parameters like this.  Is there a specific rule this is following?  The previous method was more in line with the pep8 rules.

podman/api_connection.py
```diff
@@ -87,22 +94,16 @@ def post(self, path, params=None, headers=None):
...
 
-    def request(self,
-                method,
-                url,
-                body=None,
-                headers=None,
-                *,
-                encode_chunked=False):
+    def request(
+        self, method, url, body=None, headers=None, *, encode_chunked=False
```
mwhahaha, line 98: The previous way was awful, but I don't think we should stick to aligning with the opening delimiter rather than doing it this way.

If we must split because line length (though we're using 100 instead of <80)


```
def request(self, method, url, body=None, headers=None, *, 
            encode_chunked=False):
```

podman/api_connection.py
```diff
@@ -87,22 +94,16 @@ def post(self, path, params=None, headers=None):
...
         if headers is None:
             headers = {}
 
-        super().request(method,
-                        url,
-                        body,
-                        headers,
-                        encode_chunked=encode_chunked)
+        super().request(
+            method, url, body, headers, encode_chunked=encode_chunked
```
mwhahaha, line 105: I think we should keep it aligned with the opening delimiter or if you are going to do this, I think the ) shouldn't be separated on a new line. Seems more in line with https://www.python.org/dev/peps/pep-0008/#indentation

podman/system/__init__.py
```diff
@@ -29,7 +29,10 @@ def get_info(api):
 
 
 def show_disk_usage(api):
-    """Return information about disk usage for containers, images and volumes"""
+    """
+    Return information about disk usage for containers,
+    images and volumes
+    """
```
mwhahaha, line 35: I think you updated this because of line length? Since we're using pylint, it should be 100. I'd leave this.

podman/api_connection.py
```diff
@@ -13,20 +13,26 @@
...
         uri = urllib.parse.urlparse(url)
-        if uri.scheme not in ("unix", "ssh"):
-            raise ValueError("The scheme '{}' is not supported, only {}".format(
-                uri.scheme, supported_schemes))
+        if uri.scheme not in supported_schemes:
+            raise ValueError(
+                "The scheme '{}' is not supported, only {}".format(
+                    uri.scheme, supported_schemes
+                )
+            )
```
mwhahaha, line 35: I personally prefer something like the previous format but I'm not against this way for readability.

mwhahaha (changes requested):

podman/api_connection.py
jwhonce: @mwhahaha I'm fine with letting the `black` formatter drive the style.  It is very opinionated and is almost PEP8.

mwhahaha: Works for me if that is preferred. Just need to document it.

podman/api_connection.py
danvergara: Thanks for the review @mwhahaha. Black formatted the code this way. I thinks is more readable, tho.

podman/system/__init__.py
danvergara: You're right. I define 79 characters as maximum length of the line, in the other hand Black defaults to 88 characters per line.

podman/api_connection.py
danvergara: Got it. I can fix this ASAP and adjust to the rst example docstrings, no probs.

rhatdan:
@danvergara Needs a rebase.

danvergara:
@rhatdan It's done, sir!

mwhahaha:
LGTM

mwhahaha (approved):

PR #34, opened 2020-11-03
Documentation idea

erikbgithub:
Hi Containers team,

I was checking out this project and saw it was still missing documentation. To have a quick PR that we can use as basis for working on this I created a skeleton on ReadTheDocs (the standard place for auto-generated python docs) 
 with the documentation generation engine [mkdocs](https://www.mkdocs.org/) (no asciidoc supported afaik).

I also renamed my fork so your subdomain will stay free. The PR's location of generated content is here:
https://eb-podman-py.readthedocs.io/en/latest/

No content yet but I'll start to look at the code to think about a documentation structure.

What do you think?


rhatdan:
SGTM
@baude @jwhonce PTAL

PR #35, opened 2020-11-04
Add api helper functions and unit tests

mwhahaha:
This change adds get, post, delete helper functions to the api
connection that simplify calls and handle the query parameter or data
posts. Additionally this change covers the existing calls in proper unit
testing and increases the minimum coverage to 80% as it's brought up to
90+ with this change.


rhatdan:
LGTM
@jwhonce PTAL

PR #36, opened 2020-11-04
Docs: Covering the Essentials

erikbgithub:
I think this might be enough for now regarding docs. The project structure is pretty intuitive so I just want to give hints. If it gets more complex down the road we have a nice foundation to build upon (and also still enough flexibility to switch to another solution if needed, agile ftw).

Demo view is here: https://eb-podman-py.readthedocs.io/en/latest/

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,17 +1,39 @@
-# Welcome to MkDocs
+# Welcome to Podman-Py
 
-For full documentation visit [mkdocs.org](https://www.mkdocs.org).
+This python package is a set of bindings to use the new RESTful API in
+[podman](https://github.com/containers/podman).  It is currently under
+development and contributors are welcome!
 
-## Commands
-
-* `mkdocs new [dir-name]` - Create a new project.
-* `mkdocs serve` - Start the live-reloading docs server.
-* `mkdocs build` - Build the documentation site.
-* `mkdocs -h` - Print help message and exit.
 
 ## Project layout
 
-    mkdocs.yml    # The configuration file.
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@jwhonce @baude PTAL

rhatdan:
@TomSweeneyRedHat PTAL

docs/index.md
```diff
@@ -1,17 +1,39 @@
-# Welcome to MkDocs
+# Welcome to Podman-Py
 
-For full documentation visit [mkdocs.org](https://www.mkdocs.org).
+This python package is a set of bindings to use the new RESTful API in
+[libpod](https://github.com/containers/libpod).  It is currently under
```
TomSweeneyRedHat, line 4: I think pointing at Podman rather than libpod is better here.  @mheon do you have an opinion?

```suggestion
[podman](https://github.com/containers/podman).  It is currently under
```

TomSweeneyRedHat:
One question, otherwise LGTM
Thanks @erikbgithub

docs/index.md
erikbgithub: Done. Should I also change the same line in the README? It's copied from there.

erikbgithub:
Happy to help. Great that there's a project that combines systemd, podman and Python!

docs/index.md
TomSweeneyRedHat: @erikbgithub If you wouldn't mind doing a separate PR, that would be great.  Thanks!

PR #37, opened 2020-11-04
Start networks implementation

mwhahaha:
This change also includes a renaming of the api's static function to
raise a not found error to allow for alternative types of not found
exceptions to be raised (e.g. ImageNotFound, NetworkNotFound,
ContainerNotFound, etc)


rhatdan:
LGTM
@jwhonce PTAL

PR #38, opened 2020-11-05
Clean up code, list containers

jwhonce:
* Reformat project with black formatter/style + isort
* Rename `__ALL__` to `__all__`
* Implement list_containers() + unit tests

Fixes #21 

Signed-off-by: Jhon Honce <jhonce@redhat.com>

PR #39, opened 2020-11-05
Use black formatter

### LINKED ISSUE #28: PEP8 style guide
As a python developer and a big fan of podman, I might find this package useful in order to automate some container processes. But now, I see there's a lack of features and I think this is a perfect moment to improve code quality. [PEP8](https://www.python.org/dev/peps/pep-0008/) is the official style guide for python and this library could follow those principles in a more strictly way.

i. e.

Before PEP8
```python
def __init__(self, url, base="/v1.24/libpod", *args,
    **kwargs):  # pylint: disable-msg=W1113
```

After PEP8
```python
def __init__(
    self, url, base="/v1.24/libpod", *args, **kwargs
):  # pylint: disable-msg=W1113
```

mwhahaha:
This change documents the usage of black as the prefered formatter for
this project. Additionally it adds two tox targets that can be used to
have black be invoked.

* `tox -e black` will print out what would be changed
* `tox -e black-format` will actually update the files with the code
  formatted

Black is configured via pyproject.toml with a line-length of 100 and
will skip string normalization

Closes #28

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,45 +1,83 @@
 # How to contribute
 
-Thank you in your interest in the podman-py project.  We need your help to make it successful. 
+Thank you in your interest in the podman-py project.  We need your help to make
+it successful.
 
 You may also want to look at:
- - https://github.com/containers/libpod
- - https://podman.readthedocs.io/en/latest/Reference.html
- 
-## Reporting Issues
-Before reporting an issue, check our backlog of open issues to see if someone else has already reported it. If so, feel free to add your scenario, or additional information, to the discussion. Or simply "subscribe" to it to be notified when it is updated.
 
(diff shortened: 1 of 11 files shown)


mwhahaha:
Should hold off on this until #38 lands, then it'll just be a tox env and documentation

CONTRIBUTING.md
```diff
@@ -1,45 +1,75 @@
...
+- [pylint](https://www.pylint.org/)
+- [black](https://github.com/psf/black)
+- [tox](https://tox.readthedocs.io/en/latest/)
+- You may need to use [virtualenv](https://virtualenv.pypa.io/en/latest/) to
+  support Python 3.6
+
 ## Testing
-Depending on the size of your PR we will expect at a minimum unit tests. 
+
+Depending on the size of your PR we will expect at a minimum unit tests.
```
jwhonce, line 37: @mwhahaha Thanks for this work.

Given your coverage work and we're in the high 80's on coverage. How about putting a number on the unit test/coverage results? Won't merge if PR coverage falls below 85%?

mwhahaha: Sure we can raise that and document it

rhatdan:
LGTM

PR #33, opened 2020-11-03
Adding tox to run unit tests and create venv

mwhahaha:
These changes adds tox to:
- allow us to run unit tests locally without efforts
- allow us to create python virtual env embbed in tox
- updates the Makefile to prefer tox if available but fall back to
  previous pylint/coverage method

The venv mechanism afford shortcuts to quickly setup the required python env
to run podman-py:
```
$ tox -e venv -- python -c 'import podman; print(podman.ApiConnection("unix:///"))'
venv develop-inst-noop: /home/hberaud/dev/perso/podman-py
venv installed: extras==1.0.0,fixtures==3.0.0,linecache2==1.0.0,nose==1.3.7,pbr==5.5.0,-e git+git@github.com:4383/podman-py@911121d43ca2707b2ff5aa7a22342aa74f5b2036#egg=podman,python-mimeparse==1.6.0,six==1.15.0,testtools==2.4.0,traceback2==1.4.0,unittest2==1.1.0
venv run-test-pre: PYTHONHASHSEED='2558994106'
venv run-test: commands[0] | python -c 'import podman; print(podman.ApiConnection("unix:///"))'
<podman.api_connection.ApiConnection object at 0x7f5eac25a750>
_____________ summary _____________
  venv: commands succeeded
  congratulations :)
```

Also the `test` directory is moved to the podman directory to allow
distros packagers to deliver tests if needed.

This change replaces  #27

Co-Authored-By: Hervé Beraud <hberaud@redhat.com>
Co-Authored-By: Alex Schultz <aschultz@redhat.com>
Signed-off-by: Alex Schultz <aschultz@redhat.com>

PR #40, opened 2020-11-05
Revert Makefile tox usage

mwhahaha:
The change in #33 seems to have broken failures in unitests/linting from
stopping the check.


mwhahaha:
@jwhonce PTAL

PR #41, opened 2020-11-10
Initial pods implementation

mwhahaha:
Implementation for pods and disabling too many methods lint rule because
the tests trigger on them.


jwhonce:
bors +r

PR #42, opened 2020-11-13
Bump API version and increase unittest coverage

jwhonce:
* Change to libpod API 2.0.0
* Remove example code from coverage reports
* Add APIConnection ContextManager test

### DIFF
--- a/podman/api_connection.py
+++ b/podman/api_connection.py
@@ -19,9 +19,7 @@ class ApiConnection(HTTPConnection, AbstractContextManager):
     to a Podman service.
     """
 
-    def __init__(
-        self, url, base="/v1.24/libpod", *args, **kwargs
-    ):  # pylint: disable-msg=W1113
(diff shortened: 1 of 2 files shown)


jwhonce:
@mwhahaha PTAL

podman/api_connection.py
```diff
@@ -90,14 +90,16 @@ def post(self, path, params=None, headers=None, encode=False):
             headers = {}
 
         if encode:
-            if "content-type" not in set(key.lower() for key in headers) and params:
+            if (
+                "content-type" not in set(key.lower() for key in headers)
+                and params
+            ):
```
mwhahaha, line 96: Is this an old formatting thing? I just ran black and I'm not getting this being updated. Also we switched to linelength 100 in #39 so this shouldn't be updated

jwhonce: @mwhahaha I updated my preferences for formatting.

mwhahaha (approved):
LGTM

PR #46, opened 2021-01-19
Additional container calls implementation

mwhahaha:
This change adds additional calls for the other container related calls.

### DIFF
--- a/podman/containers/__init__.py
+++ b/podman/containers/__init__.py
@@ -3,7 +3,125 @@
 
 import json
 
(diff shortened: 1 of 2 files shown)


mwhahaha:
Yea i'd like to get some integration tests written. To be honest these are just straight implementations from the API docs so I don't even know if they fully work yet. I've done some spot checking here and there but more tests would be good. It's on my list of things to do still

mwhahaha:
bors retry

Issue #47, opened 2021-01-22
Error handling in api.request


ngaller:
Hello, and thank you for making this package!

I noticed that in the `api.request` method, the error message returned by the server is not captured.  For example we might get:

    podman.errors.InternalServerError: Request POST:/v2.0.0/libpod/containers/create failed: Server got itself in trouble

Instead of:

    Decode(): json: cannot unmarshal string into Go struct field SpecGenerator.cni_networks of type []string

Is that on your roadmap, or would you be open to a PR adding that handling?

mwhahaha:
Was that message part of the actual response in some form? I believe this gets thrown because it's a 500

ngaller:
- The first message is the message from the exception
- The second message comes from the actual response and you can still access it with `e.response.read()`, so I use this code for now to wrap the API calls:

```
            try:
                response = api.post("/containers/create", json.dumps(params))
            except http.client.HTTPException as e:
                error_content = e.response.read()
                if error_content:
                    error_content_parsed = json.loads(error_content)
                    raise Exception(error_content_parsed["message"])
                raise
```

mwhahaha:
Seems reasonable. Feel free to submit a pull request

PR #50, opened 2021-01-26
api_connection: extract message from server response

ngaller:
# Behavior prior to this commit:

In the api.request method, the error message returned by the server (for a request that generated an error 500, Internal Server Error) is not captured.
For example we might get:

    podman.errors.InternalServerError: Request POST:/v2.0.0/libpod/containers/create failed: Server got itself in trouble

Instead of:

    Decode(): json: cannot unmarshal string into Go struct field SpecGenerator.cni_networks of type []string

This makes troubleshooting issues in such requests difficult.

# Behavior after this commit:

If the response body is in JSON and includes a "message" property, it is
extracted and included in the message of the exception raised.

closes issue #47

Signed-off-by: Nicolas Galler <ngl@odoo.com>

### DIFF
--- a/podman/api_connection.py
+++ b/podman/api_connection.py
@@ -123,12 +123,19 @@ class ApiConnection(HTTPConnection, AbstractContextManager):
                 response,
             )
         elif response.status >= HTTPStatus.INTERNAL_SERVER_ERROR:
+            try:
+                error_body = response.read()
+                error_message = json.loads(error_body)["message"]
+            except:
+                error_message = (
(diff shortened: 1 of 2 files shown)


mwhahaha (approved):
LGTM

rhatdan:
LGTM

mwhahaha:
Just noticed this was targeted to staging, will need to pull it into master

ngaller:
Got it - I'll make a new PR against master

PR #59, opened 2021-03-12
prune networks method with tests and typo fixes

jmguzik:
Adding one missing method with ut, skipped filters as they are not implemented yet in podman


jwhonce:
LGTM

PR #48, opened 2021-01-22
Start of integration tests

mwhahaha:
This change adds a base integration class and a podman launcher that can
be used to test the api calls and their expected results. Currently the
podman calls that will work should be ones that can be performed as
rootless. Eventually calls that require privilege (e.g. network) can be
implemented but skipped if the tests are not invoked as root.  It should
be noted that all integration tests will be skipped if a podman
executable is not in the path.

Signed-off-by: Alex Schultz <aschultz@redhat.com>

PR #62, opened 2021-03-17
Flesh out Images integration testing 

jwhonce:
* Update Makefile with integration target
* Update client to support unix:// and http+unix:// schemes
* Add Images integration tests
* Refactor PodmanLauncher to log service output to log file to capture debugging events.
* Add option to allow podman service to log debugging events during integration test
* Refactor Image and ImagesManager as needed now integration tests running.
* Refactor unittests where as-built differs from swagger

Included from https://github.com/containers/podman-py/pull/48:
> This change adds a base integration class and a podman launcher that can
> be used to test the api calls and their expected results. Currently the
> podman calls that will work should be ones that can be performed as
> rootless. Eventually calls that require privilege (e.g. network) can be
> implemented but skipped if the tests are not invoked as root.  It should
> be noted that all integration tests will be skipped if a podman
> executable is not in the path.

### DIFF
--- a/podman/config.py
+++ b/podman/config.py
@@ -1,0 +1,110 @@
+"""Read containers.conf file."""
+import os
+import pathlib
+import urllib.parse
+from functools import lru_cache
+from pathlib import Path
+from typing import Dict, Optional
+
+try:
+    import toml
+except ImportError:
+    import pytoml as toml
+
+
+class ServiceConnection:
+    """ServiceConnection defines a connection to the Podman service."""
+
+    def __init__(self, name: str, attrs: Dict[str, str]):
+        """Create a Podman ServiceConnection."""
+        self.name = name
+        self.attrs = attrs
+
+    def __repr__(self) -> str:
+        return f"""<{self.__class__.__name__}: '{self.id}'>"""
+
+    def __hash__(self) -> int:
+        return hash(tuple(self.name))
+
+    def __eq__(self, other) -> bool:
+        if isinstance(other, ServiceConnection):
+            return self.id == other.id and self.attrs == other.attrs
+        return False
+
+    @property
+    def id(self) -> str:  # pylint: disable=invalid-name
+        """Returns identifier for service connection."""
+        return self.name
+
+    @property
+    @lru_cache(1)
+    def url(self) -> urllib.parse.ParseResult:
+        """Returns URL for service connection."""
+        return urllib.parse.urlparse(self.attrs.get("uri"))
+
+    @property
+    @lru_cache(1)
+    def identity(self) -> Path:
+        """Returns Path to identity file for service connection."""
+        return pathlib.Path(self.attrs.get("identity"))
+
+
+class PodmanConfig:
+    """PodmanConfig provides a representation of the containers.conf file."""
+
+    def __init__(self, path: str = None):
+        """Read Podman configuration from users XDG_CONFIG_HOME."""
+
+        if path is None:
+            home = Path(os.environ.get("XDG_CONFIG_HOME", Path.home()))
+            self.path = home / ".config" / "containers" / "containers.conf"
+        else:
+            self.path = pathlib.Path(path)
+
+        self.attrs = dict()
+        if self.path.exists():
+            with self.path.open() as file:
+                buffer = file.read()
+            self.attrs = toml.loads(buffer)
+
+    def __hash__(self) -> int:
+        return hash(tuple(self.path.name))
+
+    def __eq__(self, other) -> bool:
+        if isinstance(other, PodmanConfig):
+            return self.id == other.id and self.attrs == other.attrs
+        return False
+
+    @property
+    def id(self):  # pylint: disable=invalid-name
+        """Returns Path() of container.conf."""
+        return self.path
+
+    @property
+    @lru_cache(1)
+    def services(self) -> Dict[str, ServiceConnection]:
+        """Returns list of service connections."""
+        services: Dict[str, ServiceConnection] = dict()
+
+        engine = self.attrs.get("engine")
+        if engine:
+            destinations = engine.get("service_destinations")
+            for key in destinations:
+                connection = ServiceConnection(key, attrs=destinations[key])
+                services[key] = connection
+
+        return services
+
+    @property
+    @lru_cache(1)
+    def active_service(self) -> Optional[ServiceConnection]:
+        """Returns active connection."""
+
+        engine = self.attrs.get("engine")
+        if engine:
+            active = engine.get("active_service")
+            destinations = engine.get("service_destinations")
+            for key in destinations:
+                if key == active:
+                    return ServiceConnection(key, attrs=destinations[key])
+        return None
--- a/podman/domain/images.py
+++ b/podman/domain/images.py
@@ -1,32 +1,39 @@
 """Model and Manager for Image resources."""
+import logging
 from typing import Any, Dict, Iterator, List, Optional, Union
 
+import requests
+
+from podman import api
 from podman.domain.manager import PodmanResource
 from podman.errors.exceptions import APIError, ImageNotFound
 
+logger = logging.getLogger("Images")
+
 
 class Image(PodmanResource):
-    """Details and configuration for an Image managed by the Podman service.
-
-    Attributes:
-        labels: Image labels
-        tags: Image tags
-    """
+    """Details and configuration for an Image managed by the Podman service."""
 
     def __repr__(self) -> str:
-        return "<%s: '%s'>" % (self.__class__.__name__, "', '".join(self.tags))
+        return f"""<{self.__class__.__name__}: '{"', '".join(self.tags)}'>"""
 
     @property
     def labels(self) -> Dict[str, str]:
         """Returns labels associated with Image."""
-        return self.attrs.get("Labels", {})
+        image_labels = self.attrs.get("Labels")
+        if image_labels is None or len(image_labels) == 0:
+            return dict()
+
+        return image_labels
 
     @property
     def tags(self) -> List[str]:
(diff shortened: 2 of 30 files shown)


jwhonce:
@mwhahaha If you have no objections, I've included your commit and built on top of it in this PR. Once I have build() included in the tests I would like to merge this PR and close https://github.com/containers/podman-py/pull/48

cevich:
Looking...I may be a bit slow, since it's all new code to me...

podman/domain/images.py
```diff
@@ -46,7 +52,7 @@ def history(self) -> List[Dict[str, Any]]:
         raise APIError(body["cause"], response=response, explanation=body["message"])
 
     def save(
-        self, chunk_size: Optional[int] = 2097152, named: Union[str, bool] = False
+        self, chunk_size: Optional[int] = 2 * 1024 * 1024, named: Union[str, bool] = False
```
cevich, line 55: Thanks for spelling out the value-math, it really helps people understand where the value came from.

podman/domain/images_manager.py
```diff
@@ -270,7 +265,6 @@ def pull(
                 config for this request. auth_config should contain the username and password
                 keys to be valid.
             platform (str) – Platform in the format os[/arch[/variant]]
-            all_tags (bool) – Pull all image tags
```
cevich, line 273: Was this intentional?  I see this is still a keyword arg in the method signature.

podman/domain/images_manager.py
```diff
@@ -306,22 +300,25 @@ def pull(
...
+            body = response.json()
             raise APIError(response.url, response=response, explanation=body["error"])
 
-        if not all_tags:
-            return self.get(body["images"][0])
+        for item in response.iter_lines():
+            body = json.loads(item)
+            if all_tags and "images" in body:
+                images: List[Image] = []
+                for name in body["images"]:
```
cevich, line 315: Is `body["images"]` guaranteed to always exist in the JSON and always be a list?  If it's a string or some other iteratable, the loop will still function but any error eventually given to the user may not make much sense.  Maybe it's worth type-checking and throwing an exception.  This will help direct any debugging effort closer to the actual problem area (JSON input processing) here.

podman/tests/integration/base.py
```diff
@@ -0,0 +1,46 @@
...
+
+class IntegrationTest(fixtures.TestWithFixtures):
+    """Base Integration test case"""
+
+    def setUp(self):
+        super().setUp()
+
+        command = os.environ.get("PODMAN_BINARY", "podman")
+        if shutil.which(command) is None:
+            self.skipTest(f"{command} is not installed, skipping integration tests")
```
cevich, line 33: I'm assuming a test-skip will ultimately result in a *PASS*.  Just as a warning: This can lead to a out-of-sight-out-of-mind situation at a high-level.  For example, PR's getting merged because "The tests are all green" where in reality, they're being skipped and nobody looked.  Granted may be a a corner-case, but it might be worth considering having this fail instead of skip.  If higher-layers shouldn't be running integration tests, they can exclude it at the harness-level more obviously/explicitly, possibly limiting surprises.

podman/tests/integration/utils.py
```diff
@@ -0,0 +1,98 @@
...
+        """start podman service"""
+        print(f"Launching {' '.join(self.cmd)}\n\tLogging to {self.log_file.name}")
+        self.proc = subprocess.Popen(self.cmd, stdout=self.log_file, stderr=self.log_file)
+
+        # wait for socket to be created
+        timeout = time.monotonic() + 30
+        while not os.path.exists(self.socket_file):
+            if time.monotonic() > timeout:
+                raise subprocess.TimeoutExpired("podman service ", timeout)
+            time.sleep(0.2)
```
cevich, line 82: Is there some positive indicator that could be waited upon instead of sleeping?  Two tenths could either be an eternity on a busy CI system, potentially obscuring another/worse problem.

cevich (commented):
Okay, took a more close gander through the changes here.  It's a good start.

podman/tests/integration/utils.py
mwhahaha: It's waiting for the socket file to be available. I found that even on local systems a bit of delay occurred prior to the service being ready to accept connections.  While you could lower the time to like 0.01, you definitely don't want to let it spin uncontrolled as it'll just eat CPU.

podman/tests/integration/utils.py
```diff
@@ -0,0 +1,98 @@
...
+                f"--time={timeout}",
+                socket_uri,
+            ]
+        )
+
+        self.log_file = tempfile.NamedTemporaryFile(
+            prefix="podman_integration_", suffix=".log", delete=False
+        )
+        self.log_file.write(f"Podman command {' '.join(self.cmd)}\n".encode())
+        self.log_file.flush()
```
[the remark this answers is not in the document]
mwhahaha: This really should use the logging module rather than creating a file like this.

podman/domain/image_build.py
```diff
@@ -155,3 +158,18 @@ def build(self, **kwargs) -> Tuple[Image, Iterator[bytes]]:
...
+            if isinstance(dockerfile, io.BytesIO):
+                info = tarfile.TarInfo("Dockerfile")
+                info.size = len(dockerfile.getvalue())
+                dockerfile.seek(0)
+            else:
+                info = tarball.gettarinfo(fileobj=dockerfile, arcname='Dockerfile')
+            tarball.add(info, dockerfile)
+
+        file.seek(0)
+        return file
```
TomSweeneyRedHat, line 175: PITA time, can we mostly convert Dockerfile to Containerfile throughout?  Perhaps not in this PR, but at some point in the not too distant future?  Not only messaging but also variables within?  Or perhaps make it secondary to Containerfile?   (Tom quickly digs fox hole to hide from Jhon in :bomb: )

podman/domain/images.py
```diff
@@ -60,10 +66,15 @@ def save(
 
         Raises:
             APIError: when service returns an error.
+
+        Notes:
+            tarball format is set to docker-archive, this allows load() to import tarball.
```
TomSweeneyRedHat, line 71:
```suggestion
            tarball format is set to docker-archive, this allows load() to import a tarball.
```

podman/tests/integration/utils.py
TomSweeneyRedHat: Should we have a max tries that would fail the test after 3 or so tries?

mwhahaha: I wouldn't put a loop cap for the exact reasons cevich mentioned, in CI this could be a while so if you looped 3 times you'd only have waited for ~.6 seconds.  There is a cap here of 30 seconds with the timeout var.  30 seconds should be plenty of time to wait for the podman api instance to become available for the integration tests.

podman/domain/images_manager.py
jwhonce: The check `"images" in body` ensures the key exists, the get() below will fail if the type isn't a list.

podman/tests/integration/utils.py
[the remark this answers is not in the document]
jwhonce: I will before duplicating, currently this is a one off.

podman/tests/integration/utils.py
jwhonce: @mwhahaha is waiting on the socket to be created on the filesystem. I added the timeout so we don't wait forever and the small sleep should help with CPU usage.

podman/domain/image_build.py
jwhonce: I can change variables and default values, I cannot change the keyword arguments as that would break developers expectations.

podman/domain/images_manager.py
jwhonce: Nope. Deleted.

podman/tests/integration/utils.py
cevich: Okay, I think I understand what's happening here.  The sleep is to prevent checking for the existence of the file, from happening repeatedly as fast as possible.  I don't see it in the code (I didn't look) but a max of 30-seconds for this loop is probably fine.

podman/config.py
```diff
@@ -0,0 +1,80 @@
...
+
+class PodmanConfig:
+    """PodmanConfig provides a representation of the containers.conf file."""
+
+    def __init__(self, path: str = None):
+        """Read Podman configuration from users XDG_CONFIG_HOME."""
+        # Python 3.6 does not include cached_property therefore must work is done in __init__.
+
+        if path is None:
+            path = os.environ.get("XDG_CONFIG_HOME", f"{Path.home()}/.config/containers")
```
cevich, line 49: I think this might be wrong.  Did you intend something like this:

```python
path = Path(os.environ.get("XDG_CONFIG_HOME", f"{Path.home()}/.config"))
path = path.joinpath("containers")
```

jwhonce: Good catch, next push has no platform dependencies:

```
            home = Path(os.environ.get("XDG_CONFIG_HOME", Path.home().name))
            self.path = home / ".config" / "containers" / "containers.conf"
```

podman/tests/integration/utils.py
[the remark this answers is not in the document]
jwhonce: @mwhahaha I went with the file because capturing stderr/stdout via logging conflicted with unittest capturing stdout/stderr. I'll keep playing with it for a future PR.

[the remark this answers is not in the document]
mwhahaha: K. I think there are ways around that and we might just need to include a test friendly logging configuration

mwhahaha (approved):
LGTM

TomSweeneyRedHat:
LGTM

PR #67, opened 2021-03-30
Build podman from master for testing

jwhonce:
Need the latest podman from master to enable the co-development between
the two repositories.

### DIFF
--- a/contrib/cirrus/latest_podman.sh
+++ b/contrib/cirrus/latest_podman.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -xeo pipefail
+
+mkdir -p "$GOPATH/src/github.com/containers/"
+cd "$GOPATH/src/github.com/containers/"
+
+dnf erase podman -y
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM other then the build failure.

contrib/cirrus/latest_podman.sh
```diff
@@ -0,0 +1,16 @@
...
+
+set -eo pipefail
+
+cd "$GOPATH"/src
+
+dnf erase podman -y
+go get github.com/containers/podman
+
+cd "$GOPATH/src/github.com/containers/podman/"
+make binaries
```
cevich, line 11: I would recommend doing a `make install PREFIX=/usr`.  This might be important as it installs configs and sets SELinux stuffs up properly.  That also puts it onto `$PATH`, so no changes to tests for `$PODMAN_BINARY` are needed.

cevich (approved):
LGTM other than one recommendation.  Nice job Jhon.

PR #69, opened 2021-04-12
Implement Manifest and ManifestsManager

jwhonce:
* New code for working with manifests stored by a Podman service
* Add unit and integration tests
* Update image files as needed

### DIFF
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -333,7 +333,7 @@ class ImagesManager(BuildMixin, Manager):
         return self.resource()
 
     def remove(
-        self, image: str, force: Optional[bool] = None, noprune: bool = False
+        self, image: Union[Image, str], force: Optional[bool] = None, noprune: bool = False
     ) -> List[Dict[str, Union[str, int]]]:
         """Delete image from Podman service.
 
@@ -353,10 +353,15 @@ class ImagesManager(BuildMixin, Manager):
         """
         _ = noprune
 
+        if isinstance(image, Image):
+            image = image.id
+
         response = self.client.delete(f"/images/{image}", params={"force": force})
         body = response.json()
 
         if response.status_code != requests.codes.ok:
+            if response.status_code == requests.codes.not_found:
+                raise ImageNotFound(body["cause"], response=response, explanation=body["message"])
             raise APIError(body["cause"], response=response, explanation=body["message"])
 
         # Dict[Literal["Deleted", "Untagged", "Errors", "ExitCode"], Union[int, List[str]]]
--- a/podman/domain/manifests.py
+++ b/podman/domain/manifests.py
@@ -1,0 +1,246 @@
+"""Model and Manager for Manifest resources."""
+import urllib.parse
+from typing import List, Optional, Union
+
+import requests
+
+from podman import api
+from podman.api import APIClient
+from podman.domain.images import Image
+from podman.domain.manager import Manager, PodmanResource
+from podman.errors.exceptions import APIError, ImageNotFound, NotFound
+
+
+class Manifest(PodmanResource):
+    """Details and configuration for a manifest managed by the Podman service."""
+
+    @property
+    def id(self) -> str:
+        """Returns the identifier of the manifest."""
+        try:
+            return self.attrs["manifests"][0]["digest"]
+        except (KeyError, TypeError):
+            return ""
+
+    @property
+    def name(self) -> str:
+        """Returns the identifier of the manifest."""
+        try:
+            if len(self.names[0]) == 0:
+                raise ValueError("Manifest attribute 'names' is empty.")
+            return self.names[0]
+        except (TypeError, IndexError) as e:
+            raise ValueError("Manifest attribute 'names' is missing.") from e
+
+    @property
+    def quoted_name(self) -> str:
+        return urllib.parse.quote_plus(self.name)
+
+    @property
+    def names(self) -> List[str]:
+        """Returns the identifier of the manifest."""
+        return self.attrs.get("names")
+
+    @property
+    def media_type(self) -> Optional[str]:
+        """Returns the Media/MIME type for this manifest."""
+        return self.attrs.get("mediaType")
+
+    @property
+    def version(self) -> int:
+        """Returns the schema version type for this manifest."""
+        return self.attrs.get("schemaVersion")
+
+    def add(self, images: List[Union[Image, str]], **kwargs) -> None:
+        """Add Image to manifest list.
+
+        Args:
+            images: List of Images to be added to manifest.
+
+        Keyword Args:
+            all (bool):
+            annotation (Dict[str, str]):
+            arch (str):
+            features (List[str]):
+            os (str):
+            os_version (str):
+            variant (str):
+
+        Raises:
+            ImageNotFound: When Image(s) could not be found
+            APIError: When service reports an error
+        """
+        params = {
+            "all": kwargs.get("all"),
+            "annotation": kwargs.get("annotation"),
+            "arch": kwargs.get("arch"),
+            "features": kwargs.get("features"),
+            "images": list(),
+            "os": kwargs.get("os"),
+            "os_version": kwargs.get("os_version"),
+            "variant": kwargs.get("variant"),
+        }
+        for item in images:
+            if isinstance(item, Image):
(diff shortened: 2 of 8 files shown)


podman/domain/images_manager.py
```diff
@@ -333,7 +333,7 @@ def pull(
         return self.resource()
 
     def remove(
-        self, image: str, force: Optional[bool] = None, noprune: bool = False
+        self, image: Union[Image, str], force: Optional[bool] = None, noprune: bool = False
```
cevich, line 336: Maybe this is my ignorance, but shouldn't the "optional" `force` parameter appear last?

jwhonce: @cevich Normally I would agree, but the argument order is set for compatibility. While a bit smelly, it is legal because the following arguments have defaults as well.

podman/domain/manifests.py
```diff
@@ -0,0 +1,247 @@
...
+class Manifest(PodmanResource):
+    """Details and configuration for a manifest managed by the Podman service."""
+
+    @property
+    def id(self) -> str:
+        """Returns the identifier of the manifest."""
+        try:
+            return self.attrs["manifests"][0]["digest"]
+        except (KeyError, TypeError):
+            return ""
```
cevich, line 23: Possibly my ignorance...Should an empty-string `ID` be allowed?  In my (limited) mind, you should allow an exception to propagate if ever referencing an item w/o an ID.

podman/domain/manifests.py
```diff
@@ -0,0 +1,247 @@
...
+
+    @property
+    def name(self) -> str:
+        """Returns the identifier of the manifest."""
+        try:
+            return self.names[0]
+        except TypeError:
+            raise ValueError("Manifest 'names' missing or empty.")
```
cevich, line 31: The "empty" case is not checked by the code, only the "missing" case.

podman/domain/manifests.py
```diff
@@ -0,0 +1,247 @@
...
+        Args:
+            digest: Image digest to be removed. Should a full Image reference be provided
+                the digest will be parsed out.
+
+        Raises:
+            ImageNotFound: When the Image could not be found
+            APIError: When service reports an error
+        """
+        if "@" in digest:
+            _, digest = digest.split("@")
```
cevich, line 139: I realize having multiple `@` in the image unexpected, but I would encourage consideration of some defensive "input-validation" here. Consider what happens if someone hand-crafts an image for malicious purposes.  What about:

```
digest = digest.split("@", maxsplit=2)[1]
```

podman/domain/manifests.py
```diff
@@ -0,0 +1,247 @@
...
+        if images is not None:
+            params["image"] = list()
+            for item in images:
+                if isinstance(item, Image):
+                    item = item.attrs["RepoTags"][0]
+                params["image"].append(item)
+
+        if all is not None:
+            params["all"] = all
+
```
cevich, line 205: Should you check if ***both*** `all` and `images` are `None`?

podman/domain/manifests.py
```diff
@@ -0,0 +1,247 @@
...
+        response = self.client.get(f"/manifests/{quoted_key}/json")
+        body = response.json()
+
+        if response.status_code != requests.codes.okay:
+            if response.status_code == requests.codes.not_found:
+                raise NotFound(body["cause"], response=response, explanation=body["message"])
+            raise APIError(body["cause"], response=response, explanation=body["message"])
+
+        body["names"] = [key]
+        return self.prepare_model(attrs=body)
```
cevich, line 242: Nit: For consistency-sake, would you consider sticking to the pattern used elsewhere:

1. If "good", return it.
2. If bad, tailor exception to fit the crime

podman/tests/integration/test_manifests.py
```diff
@@ -0,0 +1,65 @@
...
+
+        self.client = PodmanClient(base_url=self.socket_uri)
+        self.addCleanup(self.client.close)
+
+        self.alpine_image = self.client.images.pull("quay.io/libpod/alpine", tag="latest")
+
+    def tearDown(self) -> None:
+        self.client.images.remove(self.alpine_image, force=True)
+        with suppress(ImageNotFound):
+            self.client.images.remove("quay.io/unittest/alpine:latest", force=True)
```
cevich, line 23: Should this be `"quay.io/libpod/alpine:latest"`

cevich: In fact...instead of duplicating, how about using object variables?

```python
class ManifestsIntegrationTest(base.IntegrationTest):
    TEST_IMAGE_NAME = "quay.io/libpod/alpine"
    TEST_IMAGE_TAG = "latest"
    ...
                self.client.images.remove(f"{self.TEST_IMAGE_NAME}:{self.TEST_IMAGE_TAG}", force=True)
```

cevich: oh...nevermind...I understand what you're doing here.

podman/tests/integration/test_manifests.py
```diff
@@ -0,0 +1,65 @@
...
+    def test_manifest_crud(self):
+        """Test Manifest CRUD."""
+
+        self.assertFalse(
+            self.client.manifests.exists("quay.io/unittest/alpine:latest"),
+            "Image store is corrupt from previous run",
+        )
+
+        with self.subTest("Create"):
+            manifest = self.client.manifests.create(["quay.io/unittest/alpine:latest"])
```
cevich, line 34: Would it make sense to deliberately force an invalid name here?  For example: `does.not/exist/alpine:latest`

cevich: I would also encourage use of a object variable, instead of duplicating the same test-name over and over.

cevich (commented):
This is a great start, found a few nits and made some comments/observations.

podman/domain/manifests.py
jwhonce: @cevich I'm trying to create a jira card to fix this `feature` of the manifest endpoints.

podman/domain/manifests.py
jwhonce: In this case, `all` and `names` are not mutually exclusive.

podman/tests/integration/test_manifests.py
jwhonce: In tests, I prefer to duplicate the text when it provides context to the test. I've been burnt too many times otherwise.

podman/domain/manifests.py
jwhonce: I have been putting the "shorter" path first, in this case the error path and  "go right" are the same length. So sure.

podman/domain/manifests.py
jwhonce: Good catch, I'll also improve the english.

podman/domain/images_manager.py
cevich: Okay, you certainly understand the compatibility aspects better than I.

cevich (approved):
Okay,  LGTM now, thanks.

podman/domain/manifests.py
```diff
@@ -0,0 +1,246 @@
...
+            manifest = self.get(body["Id"])
+            manifest.attrs["names"] = names
+            return manifest
+
+        if response.status_code == requests.codes.not_found:
+            raise ImageNotFound(body["cause"], response=response, explanation=body["message"])
+        raise APIError(body["cause"], response=response, explanation=body["message"])
+
+    def get(self, key: str) -> Manifest:
+        """Returns and manifest by name.
```
TomSweeneyRedHat, line 218: nit
```suggestion
        """Returns the manifest by name.
```

podman/domain/manifests.py
```diff
@@ -0,0 +1,246 @@
...
+
+        Args:
+            key: Manifest name for which to search
+
+        Raises:
+            NotFound: When manifest could not be found.
+            APIError: When service reports an error.
+
+        Notes:
+            To have Manifest more conform with other PodmanResource's we use the key that
```
TomSweeneyRedHat, line 228: nit of a nit
```suggestion
            To have Manifest more conform with other PodmanResource's, we use the key that
```

TomSweeneyRedHat:
Two minor nits that could be cleaned up later, otherwise
LGTM

PR #70, opened 2021-04-12
Remove tarfile.is_tarfile check in integration test

jwhonce:
The test uses memory in place of writing the tarball to disk. This
causes the check to fail. It is not clear why this test ever passed.


cevich:
> The test uses memory in place of writing the tarball to disk.

Yeowch!

cevich (approved):
LGTM

PR #78, opened 2021-04-22
Clean up docstrings and tests

jwhonce:
* Add unittest for exists()
* Add remove() on Resource and/or Manager for each resource
* Create SwarmNotImplementedError() to reduce duplicated strings
* Refactor RESTful error handling to reduce duplicated code
* Improve docstring formatting
* Fix volume RESTful calls
* Add integration testing for volumes
* Fix platform keyword parsing
* Add missing method return types

### DIFF
--- a/podman/api/client.py
+++ b/podman/api/client.py
@@ -1,13 +1,13 @@
 """APIClient for connecting to Podman service."""
+import json
 import urllib.parse
-from typing import IO, Any, ClassVar, Iterable, List, Mapping, Optional, Tuple, Union
+from typing import IO, Any, ClassVar, Iterable, List, Mapping, Optional, Tuple, Union, Type
 
 import requests
-from requests import Response
 
 from podman import api
 from podman.api.uds import UDSAdapter
-from podman.errors import APIError
+from podman.errors import APIError, NotFound
 from podman.tlsconfig import TLSConfig
 
 _Data = Union[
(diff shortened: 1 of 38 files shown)


jwhonce:
@mwhahaha PTAL

podman/api/client.py
```diff
@@ -24,6 +24,44 @@
...
+    ) -> None:  # pylint: disable=arguments-differ
+        """Raises exception when Podman service reports one."""
+        if self.status_code < 400:
+            return
+
+        try:
+            body = self.json()
+            cause = body["cause"]
+            message = body["message"]
+        except json.decoder.JSONDecodeError:
```
mwhahaha, line 57: Do we want to handle if there is no cause/message? The could be a KeyError thrown here

podman/api/client.py
```diff
@@ -24,6 +24,44 @@
...
+
+    def __getattr__(self, item: str):
+        """Forward any query for an attribute not defined in this proxy class to wrapped class."""
+        return getattr(self._response, item)
+
+    def raise_for_status(
+        self, not_found: Type[APIError] = NotFound
+    ) -> None:  # pylint: disable=arguments-differ
+        """Raises exception when Podman service reports one."""
+        if self.status_code < 400:
```
mwhahaha, line 50: < 300?  What happens if we get a redirect?

mwhahaha (approved):
Looks ok, have a few questions which could be followed up on in a later patch

podman/api/client.py
jwhonce: The requests library takes care of redirects.

podman/api/client.py
jwhonce: I'll add that.

mwhahaha (approved):
bors r+

PR #82, opened 2021-05-01
Finish transport adapters

jwhonce:
* Add support for ssh:// and tcp: transports
* Add minimal tests for each transport
* Update tests to reflect transport changes
* Cleaned up timeout handling, lower APIs do not test for None vs.
  non-existing keyword argument
* Introduced cached_property decorator, mostly to cache creating
  managers.
* Improved docstrings
* Improved formatting URLs for future SSL verification and
  authenication.
* Bump to __version__ 3.1.2.1 tracking to Podman
* Added pyxdg dependency

### DIFF
--- a/podman/api/adapter_utils.py
+++ b/podman/api/adapter_utils.py
@@ -1,0 +1,50 @@
+"""Utility functions for working with Adapters."""
+from typing import NamedTuple, Mapping
+
+
+def _key_normalizer(key_class: NamedTuple, request_context: Mapping) -> Mapping:
+    """Create a pool key out of a request context dictionary.
+
+    According to RFC 3986, both the scheme and host are case-insensitive.
+    Therefore, this function normalizes both before constructing the pool
+    key for an HTTPS request. If you wish to change this behaviour, provide
+    alternate callables to ``key_fn_by_scheme``.
+
+    Args:
+        key_class: The class to use when constructing the key. This should be a namedtuple
+            with the scheme and host keys at a minimum.
+        request_context: An object that contain the context for a request.
+
+    Returns:
+        A namedtuple that can be used as a connection pool key.
+
+    Notes:
+        Copied from urllib3.poolmanager._default_key_normalizer.
+    """
+    # Since we mutate the dictionary, make a copy first
+    context = request_context.copy()
+    context["scheme"] = context["scheme"].lower()
+    context["host"] = context["host"].lower()
+
+    # These are both dictionaries and need to be transformed into frozensets
+    for key in ("headers", "_proxy_headers", "_socks_options"):
+        if key in context and context[key] is not None:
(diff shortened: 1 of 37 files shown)


jwhonce:
@mwhahaha PTAL

jwhonce:
@cevich I assume, the ssh tests are going to fail...

podman/api/adapter_utils.py
```diff
@@ -0,0 +1,50 @@
...
+def _key_normalizer(key_class: NamedTuple, request_context: Mapping) -> Mapping:
+    """Create a pool key out of a request context dictionary.
+
+    According to RFC 3986, both the scheme and host are case-insensitive.
+    Therefore, this function normalizes both before constructing the pool
+    key for an HTTPS request. If you wish to change this behaviour, provide
+    alternate callables to ``key_fn_by_scheme``.
+
+    Args:
+        key_class: The class to use when constructing the key. This should be a namedtuple
```
TomSweeneyRedHat, line 14:
```suggestion
        key_class: The class to use when constructing the key. This should be a named tuple
```

podman/api/adapter_utils.py
```diff
@@ -0,0 +1,50 @@
...
+            with the scheme and host keys at a minimum.
+        request_context: An object that contain the context for a request.
+
+    Returns:
+        A namedtuple that can be used as a connection pool key.
```
TomSweeneyRedHat, line 19:
```suggestion
        A named tuple that can be used as a connection pool key.
```

podman/api/adapter_utils.py
TomSweeneyRedHat: Ignore, just realized it's a Python data type

TomSweeneyRedHat:
@jwhonce unfortunately the right call on the tests.

cevich:
@jwhonce I don't think this is correct: `ssh ... ssh://root@localhost:22`.  I think what you intend is `ssh ... -p 22 root@localhost`

podman/api/ssh.py
```diff
@@ -1,75 +1,300 @@
...
+            "-N",
+            "-L",
+            f"{self.local_sock}:{uri.path}",
+        ]
+
+        if self.identity is not None:
+            path = pathlib.Path(self.identity).expanduser()
+            command += ["-i", str(path)]
+
+        command += [f"ssh://{uri.netloc}"]
```
cevich, line 75: here.  I think you need to break out the port number into a `-p` option.  Also I'm not sure about using the `ssh://` prefix...I've never used that before (maybe it works?).

cevich: Oh okay, I see in the man page now:

> ssh connects and logs into the specified destination, which may be specified
> as either [user@]hostname or a URI of the form ssh://[user@]hostname[:port]

Hrmmmm.

PR #84, opened 2021-05-04
Introduce sphinx processing for html API web site

jwhonce:
* Add docs target to Makefile
* Started updating some docstrings
* Added support for connections from containers.conf
* Ported more uses of XDG_* envvar to xdg package


jwhonce:
@mwhahaha PTAL

Makefile
```diff
@@ -47,6 +48,14 @@ release:
 	twine upload --verbose dist/*whl dist/*zip
 	# pip install podman-py
 
+.PHONY: docs
+docs:
+	mkdir -p build/docs/source
+	cp -R docs/source/* build/docs/source
+	sphinx-apidoc -f -o build/docs/source podman podman/tests podman/api_connection.py podman/containers podman/images \
```
mwhahaha, line 55: need to add sphinx to test-requirements or something to ensure it's available. Would be nice to also have a tox docs target but i can add that later

docs/source/index.rst
```diff
@@ -0,0 +1,20 @@
...
+   :caption: Contents:
+
+
+
+Indices and tables
+==================
+
+* :ref:`genindex`
+* :ref:`modindex`
+* :ref:`search`
```
mwhahaha, line 20: I've got a warning that modules.rst isn't referenced in toctree. Not certain if it's an issue or if we should add it here.

/home/aschultz/Documents/workspace/containers/podman-py/build/docs/source/modules.rst: WARNING: document isn't included in any toctree

mwhahaha (changes requested):

Makefile
jwhonce: @mwhahaha I added it to `requirements.txt` since that is where `setup` and `wheel` were already called out. I'll let you add the tox target.

mwhahaha: I don't see the update in the latest commit

docs/source/index.rst
jwhonce: added

rhatdan:
LGTM
bors r+

PR #86, opened 2021-05-06
Sphinx configured to process docstrings

jwhonce:
* Sphinx configured to create html documentation from docstrings on API.
  Low level and deprecated classes not included in index.
* Downgraded urllib3 in setup.cfg
* Updated some docstrings after seeing the results from Sphinx


jwhonce:
@lsm5 This includes the changes to setup.cfg

lsm5:
@jwhonce just wondering, do we need to have dependencies specified in both places? Or is it possible to keep just one?

jwhonce:
@lsm5 The Python packaging ecosystem is in transition. Some tools still only know about `requirements.txt` while only have moved on to the new shiny new stuff. So for now we have four files... 

* requirements.txt (the development environment)
* test-requirements.txt (superset of requirements.txt)
* setup.py/setup.cfg (what is needed to install package)
* pyproject.toml (what is needed to build package, latest)

Hope this helps.

mwhahaha:
JFYI in OpenStack we use pbr which builds on top of setuptools but supports a requirements file in the setup.cfg so you don't have that duplication.  But yea, duplication is always fun.  I would argue that test-requirements.txt is not necessarily a super-set of requirements but rather what is necessary to run tests above and beyond requirements.txt. Usually it's assumed that requirements.txt is already included.

My understanding is that pyproject.toml + setup.cfg is the replacement for setup.py with newer setuptools but it's till relatively new.

docs/source/conf.py
```diff
@@ -34,7 +35,12 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = ['sphinx.ext.autodoc', 'sphinx.ext.napoleon']
+# isort: unique-list
+extensions = [
+    'sphinx.ext.autodoc',
+    'sphinx.ext.napoleon',
+    'sphinx.ext.viewcode',
```
mwhahaha, line 42: The width on the viewcode stuff seems to assume 80 columns.  Probably should follow up with being able to widen that (it might be viewcode_tab_width, not certain).

mwhahaha: correction, seems to maybe need a custom css. it can wait

lsm5:
@jwhonce thanks!

Can we get a new tag published?

PR #87, opened 2021-05-07
Cleanup docstrings and generated html pages for API

jwhonce:
* Overused Notes in docstrings.
* Change Literal import to not mask which Literal is being used
* Reduced number of warnings from sphinx-build
* Fixed type hints for @properties


rhatdan:
LGTM

PR #108, opened 2021-06-07
Fix volumes creating for client API

sshnaidm:


jwhonce:
LGTM

jwhonce (approved):

bguezzie:
Not sure if this is the right place to ask, but what's the release cadence typically like for this project?  My IDE doesn't automatically tell me when a new version is out, but this fix would be very helpful for me.  I can work around for now by spinning up podman directly and using command line args, but would also like to have a [rough] idea for when this will be available in a release.

PR #116, opened 2021-06-11
Add inspect capability to this container

space55:
While developing against this API, I found the need to inspect a
container and gain an insight into its internals. It appeared that
inspect had not been implemented in podman-py, so I added it,
duplicating some code from the `container.kill()` function.

Signed-off-by: Eamonn Nugent <4699217+space55@users.noreply.github.com>

### DIFF
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -214,7 +214,7 @@ class Container(PodmanResource):
         stat = api.decode_header(stat)
         return response.iter_content(chunk_size=chunk_size), stat
 
-    def inspect(self) -> None:
+    def inspect(self) -> Dict:
         """Inspect a container.
 
         Raises:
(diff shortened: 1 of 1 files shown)


podman/domain/containers.py
```diff
@@ -214,6 +214,16 @@ def get_archive(
         stat = api.decode_header(stat)
         return response.iter_content(chunk_size=chunk_size), stat
 
+    def inspect(self) -> None:
```
mwhahaha, line 217: "-> None" is probably not correct. Can you fix that to reflect what the json object is

mwhahaha (changes requested):

podman/domain/containers.py
space55: I'm looking at this, and obviously you're correct - `None` is wrong. However, it returns a dict of a lot of types - would it be best to just have it be:
```py
def inspect(self) -> Dict:
```
Or would a different approach be preferred?

mwhahaha: Yea dict would be fine.

space55: Added as requested

mwhahaha (approved):
/lgtm

PR #117, opened 2021-06-17
Workaround typing_extensions module requirement

jwhonce:
This requirement was missed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ export GOPROXY=https://proxy.golang.org
 PYTHON ?= $(shell command -v python3 2>/dev/null || command -v python)
 DESTDIR ?= /
 DESTDIR ?=
-EPOCH_TEST_COMMIT ?= $(shell git merge-base $${DEST_BRANCH:-master} HEAD)
+EPOCH_TEST_COMMIT ?= $(shell git merge-base $${DEST_BRANCH:-main} HEAD)
 HEAD ?= HEAD
 
 export PODMAN_VERSION ?= "3.2.0"
--- a/podman/api/__init__.py
+++ b/podman/api/__init__.py
@@ -33,9 +33,12 @@ VERSION: str = _api_version(version.__version__)
 COMPATIBLE_VERSION: str = _api_version(version.__compatible_version__, 2)
 
 try:
-    from typing_extensions import Literal
-except ModuleNotFoundError:
     from typing import Literal
+except (ImportError, ModuleNotFoundError):
+    try:
+        from typing_extensions import Literal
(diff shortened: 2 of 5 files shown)


jwhonce:
@lsm5 FYI, keeping this as draft until we know outcome of packaging needs.

requirements.txt
```diff
@@ -4,4 +4,5 @@ urllib3>=1.24.2
 pyxdg>=0.26
 sphinx
 wheel
+typing_extensions; python_version < '3.8'
```
lsm5, line 7: @jwhonce help me read this line, does this say: `Explicitly require typing_extensions for any python less than 3.8?`

If that is the case, we still need to include typing-extensions in rhel8.

lsm5:
@jwhonce remind me what exactly I need to test to greenlight this for 8.5?

jwhonce:
@lsm5 Do `make` and `make tests` would for your environment?

lsm5:
RE: rhel8.5, there's `python3-pytoml` and `python39-toml`. python3-pytoml would be the recommended version to use because it would be supported for the entire lifetime of rhel8. How bad is it if we depend on python3-pytoml in the rpm packaging? 


btw, rpm builds fine for now on rhel8.5 

RE: `make tests`, I see: `DEBUG:podman.ssh_adapter:Waiting on /run/user/1000/podman/podman-forward-a97c89ca551bad70c129.sock` , even after installing podman 3.2 and enabling podman.sock. Let me know if you prefer a tmate session.

jwhonce:
@lsm5 `DEBUG:podman.ssh_adapter:Waiting on...` only occurs when running with debugging on and is telling you the client is waiting on the external ssh client. This is not an error. Should the ssh client never forward the connection, you will then receive a TimeoutExpired exception/message. That is an error.

lsm5:
> @lsm5 `DEBUG:podman.ssh_adapter:Waiting on...` only occurs when running with debugging on and is telling you the client is waiting on the external ssh client. This is not an error. Should the ssh client never forward the connection, you will then receive a TimeoutExpired exception/message. That is an error.

remind me, is there anything else for me to do on this part then? 


> If you use different spec files or differentiate between RHEL8 and RHEL9 in one spec file that is fine with me. Do you need me to tweak requirements.txt for which toml package to depend on ?

I think that's not needed right now, but if brew complains when we do the actual builds, then maybe.

lsm5:
LGTM as far as build is concerned. I haven't had the chance to follow up on testing though.

jwhonce:
@mwhahaha PTAL Thanks.

PR #126, opened 2021-08-04
Add exec_run implementation for containers

JacobCallahan:
This change adds a basic implementation of exec_run for containers
Environment variable conversions are handled from dict->list[str]
DetachKeys will need an actual argument option
detach, stream, socket, and demux arguments are currently not handled.

### DIFF
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -2,6 +2,7 @@
 import io
 import json
 import logging
+import shlex
 from contextlib import suppress
 from typing import Any, Dict, Iterable, Iterator, List, Mapping, Optional, Sequence, Tuple, Union
 
@@ -133,7 +134,7 @@ class Container(PodmanResource):
         stdout: bool = True,
         stderr: bool = True,
         stdin: bool = False,
-        tty: bool = False,
+        tty: bool = True,
         privileged: bool = False,
         user=None,
         detach: bool = False,
@@ -165,16 +166,42 @@ class Container(PodmanResource):
             demux: Return stdout and stderr separately
 
         Returns:
-            TBD
+            First item is the command response code
+            Second item is the requests response content
 
         Raises:
             NotImplementedError: method not implemented.
             APIError: when service reports error
         """
-        if user is None:
-            user = "root"
-
-        raise NotImplementedError()
+        # pylint: disable-msg=too-many-locals
+        user = user or "root"
+        if isinstance(environment, dict):
+            environment = [f"{k}={v}" for k, v in environment.items()]
+        data = {
+            "AttachStderr": stderr,
+            "AttachStdin": stdin,
+            "AttachStdout": stdout,
+            "Cmd": cmd if isinstance(cmd, list) else shlex.split(cmd),
+            # "DetachKeys": detach,  # This is something else
+            "Env": environment,
+            "Privileged": privileged,
+            "Tty": tty,
+            "User": user,
+            "WorkingDir": workdir,
+        }
+        # create the exec instance
+        response = self.client.post(f"/containers/{self.name}/exec", data=json.dumps(data))
+        response.raise_for_status()
+        exec_id = response.json()['Id']
+        # start the exec instance, this will store command output
+        start_resp = self.client.post(
+            f"/exec/{exec_id}/start", data=json.dumps({"Detach": detach, "Tty": tty})
+        )
+        start_resp.raise_for_status()
+        # get and return exec information
+        response = self.client.get(f"/exec/{exec_id}/json")
+        response.raise_for_status()
+        return response.json().get('ExitCode'), start_resp.content
 
     def export(self, chunk_size: int = api.DEFAULT_CHUNK_SIZE) -> Iterator[bytes]:
         """Download container's filesystem contents as a tar archive.


JacobCallahan:
@jwhonce this is the version that is currently seeing the 500 ISE when used in my project.
Screenshot from 2021-08-03 16-29-41

podman/domain/containers.py
```diff
@@ -170,10 +170,24 @@ def exec_run(
...
+            "AttachStdout": stdout,
+            "Cmd": cmd if isinstance(cmd, list) else [cmd],
+            # "DetachKeys": detach,
+            "Env": environment,
+            "Privileged": privileged,
+            "Tty": tty,
+            "User": user,
+            "WorkingDir": workdir
+        }
+        response = self.client.post(f"/containers/{self.name}/exec", params=params)
```
jwhonce, line 188: @JacobCallahan Please review https://docs.podman.io/en/latest/_static/api.html#operation/ContainerExecLibpod, `name` is the only query parameter you may pass, all the other fields need to be passed as `data`.

JacobCallahan: addressed in current update

podman/domain/containers.py
```diff
@@ -170,10 +170,32 @@ def exec_run(
...
+            "Tty": tty,
+            "User": user,
+            "WorkingDir": workdir
+        }
+        # create the exec instance
+        response = self.client.post(f"/containers/{self.name}/exec", data=json.dumps(data))
+        response.raise_for_status()
+        exec_id = response.json()['Id']
+        # start the exec instance
+        response = self.client.post(f"/exec/{exec_id}/start", params={"Detach": detach, "Tty": tty})
```
jwhonce, line 193: @JacobCallahan Thanks for your work on this. https://docs.podman.io/en/latest/_static/api.html#operation/ExecStartLibpod shows instead of `params` you want to pass that dictionary as JSON using the `data` keyword.

JacobCallahan:
@jwhonce thanks again for the sage advice! 
Now there is a question of what to return from this method.
right now it is a dict of the exec instance info, but I think an actual object (preferably one that you can pull stdout/err/status) from would be best.

If the object route is the way to go, how would that be done? 

Screenshot from 2021-08-06 16-20-47

rhatdan:
@jwhonce @JacobCallahan what is going on with this PR?

JacobCallahan:
> @jwhonce @JacobCallahan what is going on with this PR?

Unfortunately, I've been dealing with much higher priorities since the last update. Hopefully I'll be able to revisit this soon.

One thing that would help with this would be if someone could point out an efficient way to get stdout/stderr from an exec instance container

rhatdan:
@JacobCallahan @jwhonce What should we do with this PR?

JacobCallahan:
@rhatdan @jwhonce I've revisited this today and have it to the point where it returns the original container object (not the exec container since that seems to be gone by the time the manager can get the object). I don't think this is a very useful return. An output-like object would be better, but I'm not sure how to create that. Any pointers would be appreciated.

JacobCallahan:
Alright, In the latest version, I have exec_run working with basic commands. However, since it uses entrypoint, it complicates bash-like commands being passed in. In the simple example below, the command executes correctly, the file is created, and we get the correct return code. However, we currently are unable to get any additional relevant information or output from commands.
```python
In [4]: ubi8.exec_run("touch test.txt")
{'CanRemove': True, 'ContainerID': 'a260ee5258383680de1189f603f3218a9609ddb664d68d390b5dad58481932af', 'DetachKeys': '', 'ExitCode': 0, 'ID': 'fd37335600af2af4c4f717723eda0fc80ba5767613c1b22b5e3480972179cfab', 'OpenStderr': True, 'OpenStdin': False, 'OpenStdout': True, 'Running': False, 'Pid': 0, 'ProcessConfig': {'arguments': ['test.txt'], 'entrypoint': 'touch', 'privileged': False, 'tty': True, 'user': 'root'}}
Out[4]: 0
```

msisj:
@JacobCallahan 
Actually you have almost working solution. It just need a quick tweaks :

1. [split() is a wrong method to use here](https://github.com/containers/podman-py/pull/126/files#diff-ab73663c44e617127ed59777b745050a4a6323370e57c5099a72ca192c7aeb5cR180). As this is exec is using (working similar) as the entrypoint, to allow more complex command you should use [shlex.split()](https://docs.python.org/3.6/library/shlex.html#shlex.split) . Then the complex commands works (one of those is shown in example below)
2. [rename response to something else to not lose it e.g. start_respose](https://github.com/containers/podman-py/pull/126/files#diff-ab73663c44e617127ed59777b745050a4a6323370e57c5099a72ca192c7aeb5cR193)
3. [return tuple start_response.text, response.json().get('ExitCode')](https://github.com/containers/podman-py/pull/126/files#diff-ab73663c44e617127ed59777b745050a4a6323370e57c5099a72ca192c7aeb5cR200) 
4. In example you'll see some ascii signs returned by Podman REST API. Either erase them in REST API or here to prevent further issues
5. Update method's documentation's Return section 

So it turns your the exec start is blocking whole REST and the end in text we have the output of command. I've tried it with longer commands (sleep for 90seconds and then print whatever and it worked)

Example : 
```
>>> container
<Container: 65246b3e7b>
>>> output, ec = container.exec_run('bash -c "sleep 15; cat /etc/resolv.conf"')
>>> output
'\x01\x00\x00\x00\x00\x00\x00+search dns.podman\nnameserver 192.168.194.1\n'
>>> ec
0
>>> output, ec = container.exec_run('wrong')
>>> output
'\x02\x00\x00\x00\x00\x00\x00¿Error: exec failed: container_linux.go:380: starting container process caused: exec: "wrong": executable file not found in $PATH: OCI runtime attempted to invoke a command that was not found\n'
>>> ec
127
```

Let me know what you think about it

JacobCallahan:
@msisj Thanks for the fantastic testing! I'm pretty swamped at the moment, but I will definitely check out and try your suggestions. Also, do you think it would be more useful to return the output and exit status in a single object or separate how it is now? I additionally need to consider demuxing the streams.

msisj:
@JacobCallahan 
Normally I'd say to stay as similar to docker-py (when it comes to the interface) as possible, but as I can see in docker-py there were only `exec_[create|inspect|resize|start]` methods which are available in podman REST API. There was no collective exec_run which did all of these functionalities at once. So that's a no go.

Returning an object is generally a good idea I suppose, but I'd try to ask any of the owners of this product what interface they want to have honestly.

If it comes to demuxing streams I'm wondering why do you need to consider this, is there a planned rework for Podman REST API in this area? 
As for current, fairly new version (4.0.0), I can see there is no way of getting them separate (honestly I'm not even sure if the stderr is shown now, haven't tested it). 

To be even more honest, documentation is rather lacking as the [exec start](https://docs.podman.io/en/latest/_static/api.html#operation/ExecStartLibpod) basically does not say that it returns any output.
What is worth noting is that to successfully and reliably get an output from start is to use `detach=False` as the documentation says.

msisj:
@JacobCallahan Are you planning to work on this draft? If you don't, I can take some of this code, tweak it and create PR  for it (basically I need this feature on master).

samarAvi:
@JacobCallahan @rhatdan Do we have any ETA on this feature? This feature is absolutely critical for our use-case.

JacobCallahan:
@msisj @samarAvi I'm still busy with other things, but I will try to revisit this today.

samarAvi:
@JacobCallahan Hope you got the chance you revisit this. Thanks in advance. Please share if you have any updates.

JacobCallahan:
@msisj i had a bit of time to play around with this a bit more today and try out your suggestions. However, I'm still not actually seeing the command output you're seeing. I am only getting back the Id for the exec_start instance.
Is there perhaps some other local change you made to include that command output?

```
In [4]: res = ch.exec_run('bash -c "sleep 15; cat /etc/resolv.conf"')

In [5]: res
Out[5]: 
('{"Id":"7d0d0a9219f6f932b7a85bfda7b64b3b011e33623f67e116401c9f47169ac446"}\n',
 0)
```

for reference, the meat of my local version of exec_run
```python
# create the exec instance
start_resp = self.client.post(f"/containers/{self.name}/exec", data=json.dumps(data))
start_resp.raise_for_status()
exec_id = start_resp.json()['Id']
# start the exec instance
response = self.client.post(f"/exec/{exec_id}/start", data=json.dumps(
    {"Detach": detach, "Tty": tty}
))
response.raise_for_status()
# get and return exec information
response = self.client.get(f"/exec/{exec_id}/json")
response.raise_for_status()
return start_resp.text, response.json().get('ExitCode')
```

msisj:
@JacobCallahan 
You are super close.
I think you've overlooked what I've written. In point 2, when you click it I've linked which line should have the `start_response`.
As the name states, this is response of `start` call rather than `exec`, that's weird but the output is saved to `/start` endpoint by podman REST API.
To be more precise : 
```
start_resp = self.client.post(f"/containers/{self.name}/exec", data=json.dumps(data))
start_resp.raise_for_status()
``` 
->
```
response = self.client.post(f"/containers/{self.name}/exec", data=json.dumps(data))
response .raise_for_status()
```

and
```
response = self.client.post(f"/exec/{exec_id}/start", data=json.dumps(
    {"Detach": detach, "Tty": tty}
))
response.raise_for_status()
``` 
->
```
start_resp = self.client.post(f"/exec/{exec_id}/start", data=json.dumps(
    {"Detach": detach, "Tty": tty}
))
start_resp.raise_for_status()
```

samarAvi:
Thanks @msisj , That helped. 
```
In [29]: r = cont.exec_run("ls", privileged = True, stdin = True, tty = True,stream= True)
In [31]: r[1].text
Out[31]: 'bin   core  etc   lib\t lib64\t media\topt   root  sbin  sys  usr\r\nboot  dev   home  lib32  libx32  mnt\tproc  run   srv   tmp  var\r\n'
```
@JacobCallahan  I'm still having issues with interactive-tty . Below code gets stuck since it's waiting for `stdin`. 
```
In [32]: r = cont.exec_run("ssh aviuser@10.102.66.74", privileged = True, stdin = True, tty = True, stream= True)

```
Process is getting started inside the container.
```
root@01a1aad29664:/# ps ax | grep ssh
     93 pts/4    Ss+    0:00 ssh aviuser@10.102.66.74
    102 pts/2    S+     0:00 grep --color=auto ssh
root@01a1aad29664:/#
```
Any hints?

samarAvi:
Thanks for amazing work so far @JacobCallahan. IMHO attach/attach_socket also needs to be implemented as part of this PR then, could you please take it up. @msisj what do you think?

msisj:
@samarAvi I don't have anything against the idea (if REST API supports this operation) as [this was apparently supported in docker-py](https://docker-py.readthedocs.io/en/stable/api.html#docker.api.exec_api.ExecApiMixin.exec_start), but for me it's a non-blocker as I'm not using this functionality right now

JacobCallahan:
@samarAvi I'm unable to commit the time to implement both of those. Also, that would be scope-creep for this PR, which I'd like to avoid.
If you, or anyone else following this PR, would like a potential hint on how to implement attach_socket, you may be able to steal the socket file descriptor from requests (when you make the attach call) to recreate an interactive socket.

samarAvi:
@JacobCallahan Thanks for the hint. Let me dig up a little on socket file descriptor in requests. I'll try to help in whichever way I can.

jwhonce:
@JacobCallahan I like where this is going. Thank you for your hard work on this.

docker-py returns a tuple (exit_code: Int, output: Union[generator, bytes, or tuple]) from https://docker-py.readthedocs.io/en/stable/containers.html I am comfortable with you implementing (int, bytes) and we can add other return types in the future.

JacobCallahan:
@jwhonce thanks. i've switched the return order so the return code is leading. However, I'm getting normal strings back from the API without any additional processing. In order to return bytes, I'd have to explicitly convert it (specifying an encode type like utf-8). If this is something you want, I can certainly add it.

```python
In [4]: res = ch.exec_run('hostname')

In [5]: res
Out[5]: (0, '1b122d3312b8\r\n')

In [6]: type(res[1])
Out[6]: str
```

enhaut:
@JacobCallahan  
`APIClient.post()` returns `Response` object, so instead of using property `.text`, try to use `.content` - https://requests.readthedocs.io/en/latest/user/quickstart/#binary-response-content in your `exec_run` implementation

jwhonce:
@JacobCallahan For compatibility, I agree with @enhaut you should change to .content from .text. I understand we're pushing the burden to convert to str on most developers, but should the output be binary those will need this type. From reading the documentation and walking code, I think this is it!

JacobCallahan:
@jwhonce @enhaut i've updated the PR to now return the content as requested. Hope this helps!

enhaut:
@JacobCallahan you need to remove the draft status

enhaut:
Whats going on with this PR? @rhatdan

rhatdan:
@enhaut Isn't failing a gating test?  @JacobCallahan Could you fix this to pass the test?

rhatdan:
/approve
@mwhahaha @jwhonce PTAL

jwhonce:
LGTM

PR #130, opened 2021-09-14
Update VM Images + Drop prior-ubuntu references

cevich:
Prior-Ubuntu support is being dropped everywhere.

Ref: https://github.com/containers/podman/issues/11070
     https://github.com/containers/automation_images/pull/88

Signed-off-by: Chris Evich <cevich@redhat.com>

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -17,20 +17,17 @@ env:
     FEDORA_NAME: "fedora-34"
     PRIOR_FEDORA_NAME: "fedora-33"
     UBUNTU_NAME: "ubuntu-2104"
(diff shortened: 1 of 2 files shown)


jwhonce:
@cevich I think https://github.com/containers/podman-py/blob/6532351a4c95ceae83d1387924e2a024dbdb7eb2/.cirrus.yml#L4 needs to be `main` not `master`.

cevich:
Oh, is that still there? I don't even think we use it...

cevich:
...fixed, was a rebase goof on my part.

PR #132, opened 2021-09-16
Revert pyproject.toml requires changes

### LINKED ISSUE #128: fedora package isn't picking up pyxdg and requests deps from upstream files
For whatever reason, rpm packaging isn't automatically picking up the `pyxdg` and `requests` dependencies from upstream. 

See:
1. https://kojipkgs.fedoraproject.org//work/tasks/7254/73967254/build.log
2. https://kojipkgs.fedoraproject.org//work/tasks/7700/73967700/build.log

If I specify deps explicitly in the rpm specs, it builds fine.

Not yet sure if this is an issue upstream or with fedora's python packaging.

jwhonce:
Fedora packaging stopped picking up dependencies

@lsm5 FYI

Fixes #128


lsm5:
@jwhonce wanna bump the version as well?

lsm5 (approved):
LGTM, scratch build worked  https://koji.fedoraproject.org/koji/taskinfo?taskID=75798567 .

lsm5:
@jwhonce could you also link #128 here? I don't have enough privs to set that.

.cirrus.yml
```diff
@@ -1,133 +1,186 @@
...
-    # Google-cloud VM Images
-    IMAGE_SUFFIX: "c6431352024203264"
-    FEDORA_CACHE_IMAGE_NAME: "fedora-${IMAGE_SUFFIX}"
-    PRIOR_FEDORA_CACHE_IMAGE_NAME: "prior-fedora-${IMAGE_SUFFIX}"
-    UBUNTU_CACHE_IMAGE_NAME: "ubuntu-${IMAGE_SUFFIX}"
-
-    # Container FQIN's
-    FEDORA_CONTAINER_FQIN: "quay.io/libpod/fedora_podman:${IMAGE_SUFFIX}"
-    PRIOR_FEDORA_CONTAINER_FQIN: "quay.io/libpod/prior-fedora_podman:${IMAGE_SUFFIX}"
-    UBUNTU_CONTAINER_FQIN: "quay.io/libpod/ubuntu_podman:${IMAGE_SUFFIX}"
```
cevich, line 30: I'm not opposed to an indentation change...other than it will make diffing this vs other repos hard.  Since `.cirrus.yml` shares a lot of the same names/conventions, this is a handy thing to maintain.

.cirrus.yml
```diff
@@ -1,133 +1,186 @@
...
+    - "$SCRIPT_BASE/enable_ssh.sh"
+    - go get github.com/vbatts/git-validation
+    - make
+    - make validate
 
-    # see bors.toml
-    skip: $CIRRUS_BRANCH =~ ".*\.tmp"
+gating_task:
+  name: "Gating test"
+  alias: gating
```
cevich, line 104: Let's just nuke this whole `gating_task`.  I don't think it's necessary.

.cirrus.yml
```diff
@@ -1,133 +1,186 @@
...
+  # see bors.toml
+  skip: $CIRRUS_BRANCH =~ ".*\.tmp"
 
-    # see bors.toml
-    skip: $CIRRUS_BRANCH =~ ".*\.tmp"
+  # Runs within Cirrus's "community cluster"
+  container:
+    image: "$FEDORA_CONTAINER_FQIN"
+    cpu: 4
+    memory: 12
```
cevich, line 156: Do you really want to run this in a container?  I think you're `enable_podman.sh` will fail due to lack of systemd conditions.

.cirrus.yml
```diff
@@ -1,133 +1,186 @@
...
+  name: "Total Success"
+  alias: success
+
+  # see bors.toml
+  skip: $CIRRUS_BRANCH =~ ".*\.tmp"
+
+  # N/B: ALL tasks must be listed here, minus their '_task' suffix.
+  depends_on:
+    - meta
+    - gating
```
cevich, line 182: don't forget to add `build`

contrib/cirrus/enable_ssh.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -xeo pipefail
+
+dnf install openssh-server openssh-clients -y
```
cevich, line 5: This shouldn't be necessary on the VMs, pretty sure they already have ssh installed (maybe not running though).

.cirrus.yml
jwhonce: @cevich My yaml editor got excited. I'll look into restoring indentation.

docs.old/requirements.txt
```diff
@@ -0,0 +1 @@
+mkdocs==1.1.2
```
cevich, line 1: Probably safe to just delete`docs.old`, you can rely on `git` to remember how it was, no need to keep it around unless it's actually used - but then the `old` name is kind of confusing.

.cirrus.yml
jwhonce: How were we running before? enable_podman.sh was extracted from latest_podman.sh so I could call it from two places.

make.bat
```diff
@@ -0,0 +1,35 @@
+@ECHO OFF
```
cevich, line 1: This made me cringe and then fart and then I moaned "autoexec" for some inexplicable reason.

.cirrus.yml
jwhonce: I was assuming the task dependency tree would take care of that for me. Will add it.

pyproject.toml
```diff
@@ -17,6 +17,10 @@ profile = "black"
 line_length = 100
 [build-system]
 requires = [
+    "requests>=2.24",
+    "toml>=0.10.2",
+    "urllib3>=1.24.2",
+    "pyxdg>=0.26",
```
cevich, line 23: (Just a random thought)

Is there a way I can ingest this file during the VM Image build workflow (in c/automation_images), to check if those requirements are met?

This might be worth doing, since otherwise we only find out about mismatches until after the image build is finished and any fix would require doing it again.

.cirrus.yml
cevich: ***Not*** specifying a `container` or `gce_instance` attribute will cause it to inherit the top-level value (which is a `gce_instance` a.k.a. a VM IIRC)

.cirrus.yml
cevich: It will, but that's hard for old human brains to process.  Also, there may be skips/only_if clauses in play, which makes it more hard for old brains.  Easier to just list everything here and rely on Cirrus doing the right thing.

cevich: [In case it helps, Ed made a visualization tool for this very reason.](https://github.com/containers/automation/tree/main/cirrus-task-map)

make.bat
jwhonce: I'm nuking this file as well. Sphinx cruft.

pyproject.toml
jwhonce: I will have to research.

.cirrus.yml
jwhonce: I left gating and nuked build_task since it did not work as I expected.

.cirrus.yml
jwhonce: I can play will with others. :-) All tasks listed.

contrib/cirrus/enable_ssh.sh
jwhonce: Leftover cruft from me flaying at containers vs. VMs.

lsm5 (approved):
LGTM

pyproject.toml
cevich: There's no reason the c/automation_images CI can't git-clone this repo, and do something really simple, like: use `jq` on the file, feed it to `pip` and `grep -q` for n instances of `not installing...` type messages.

.cirrus.yml
```diff
@@ -109,9 +111,29 @@ test_task:
...
+        - ${SCRIPT_BASE}/enable_ssh.sh
+        - ${SCRIPT_BASE}/enable_podman.sh
+        - ${SCRIPT_BASE}/test.sh
+
+latest_task:
+    name: "Test Podman main on $FEDORA_NAME"
+    alias: latest
+    allow_failures: true
+
+    # see bors.toml
```
cevich, line 123: not-critical: bors is dead :grin:

.cirrus.yml
```diff
@@ -109,9 +111,29 @@ test_task:
...
+    skip: $CIRRUS_BRANCH =~ ".*\.tmp"
```
cevich, line 124: In fact...I think it was bors that required this `skip`, could maybe nuke all of them as well.

PR #141, opened 2021-10-22
Bugfix/extra hosts

msisj:
It was possible to pass extra_hosts earlier on but it was on user to know what is the format (["{hostname}:{ip}"], ...) and the documentation for podman-py was wrong. To make things more consistent with docker-py (https://docker-py.readthedocs.io/en/stable/api.html#docker.api.container.ContainerApiMixin.create_host_config - because why not) I've added code to make it docker-py compliant and documentation (https://github.com/containers/podman-py/blob/main/podman/domain/containers_create.py#L70-L71) compliant, while keeping interface as easy as possible (dict over list with special formatted strings)

```
Before : 

>>> import podman
>>> c = client.containers.create(name='test', image='test:latest', extra_hosts={"host1 host3": "127.0.0.2", "host2": "127.0.0.3"})
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/podman/domain/containers_create.py", line 224, in create
  File "/usr/lib/python3.6/site-packages/podman/api/client.py", line 65, in raise_for_status
    raise APIError(cause, response=self._response, explanation=message)
podman.errors.exceptions.APIError: 500 Server Error: Internal Server Error (Decode(): json: cannot unmarshal object into Go struct field SpecGenerator.hostadd of type []string)

tox -e coverage :

Ran 277 tests in 39.548s

FAILED (SKIP=3, errors=5, failures=3)



After :

>>> import podman
>>> c = client.containers.create(name='test', image='test:latest', extra_hosts={"host1 host3": "127.0.0.2", "host2": "127.0.0.3"})
>>> c.attrs['HostConfig']['ExtraHosts']
['host1 host3:127.0.0.2', 'host2:127.0.0.3']

...

[root@host ~]# podman exec test cat /etc/hosts
127.0.0.1   localhost localhost.localdomain localhost4 localhost4.localdomain4
::1         localhost localhost.localdomain localhost6 localhost6.localdomain6
127.0.0.2 host1 host3
127.0.0.3 host2

tox -e coverage :

Ran 277 tests in 35.199s

FAILED (SKIP=3, errors=5, failures=3)
```

### DIFF
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -302,7 +302,7 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
             "expose": dict(),
             "groups": pop("group_add"),
             "healthconfig": pop("healthcheck"),
-            "hostadd": pop("extra_hosts"),
+            "hostadd": list(),
             "hostname": pop("hostname"),
             "httpproxy": pop("use_config_proxy"),
             "idmappings": pop("idmappings"),  # TODO document, podman only
@@ -366,6 +366,9 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
         for item in args.pop("exposed_ports", list()):
             port, protocol = item.split("/")
             params["expose"][int(port)] = protocol
+
+        for ip, hostname in args.pop("extra_hosts", dict()).items():
+            params["hostadd"].append(f"{ip}:{hostname}")
 
         if "log_config" in args:
             params["log_configuration"]["driver"] = args["log_config"].get("Type")
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -367,8 +367,8 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
             port, protocol = item.split("/")
             params["expose"][int(port)] = protocol
 
-        for ip, hostname in args.pop("extra_hosts", dict()).items():
-            params["hostadd"].append(f"{ip}:{hostname}")
+        for hostname, ip in args.pop("extra_hosts", dict()).items():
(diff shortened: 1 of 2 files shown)


msisj:
@jwhonce 
Could you clarify which part of @mwhahaha comments you would like me to implement? (current changes are fully compatible with docker-py https://docker-py.readthedocs.io/en/stable/api.html#docker.api.container.ContainerApiMixin.create_host_config)

I could add unit test, but a test suite with pod/container creation with different options is missing (I think there should be test case for each option, as this is crucial functionality).
As I'm not project owner I feel hesitant on adding something new to the structure since I don't have that much knowledge how you would like things to be done.

jwhonce:
@msisj For the tests to verify parameter types and formatting,  I would add a file tests/unit/test_container_create.py and tests/unit/test_pod_create.py that way you can just mock up the server code.

If you want to verify the service API results then tests/integration/test_container_create.py and tests/integration/test_pod_create.py would be my suggestions.

Does that make sense? Help? Thanks for all your work on this PR!

msisj:
@jwhonce I've added test case for extra hosts, but only for containers (since option is only added to container, not to pod).
If it comes to the type checking, I've skipped that for now, since if it comes to worst, podman api will accept the passed strings and /etc/hosts will have just some invalid entries, which shouldn't make too much harm).

Please let me know whether such solution is acceptable.

```
tox -e coverage
...
Ran 278 tests in 40.806s

FAILED (SKIP=3, errors=2)
```

mwhahaha (approved):
/lgtm

PR #143, opened 2021-10-22
Allow passing all available options to volume mount

msisj:

Right now it is possible only to pass mode to mounted volume during container creation. I have no idea what was the reasoning behind it but the Podman REST API is capable of passing more options.

I'm only concerned about backward compatibility : 

- In docker-py option was called mode but there was possibility to pass multiple arguments (e.g. `z,ro`)
- I've renamed option `mode` -> `options` to be more precise what this option is actually doing so there is actually no backward compatibility with podman-py. Maybe this option should still be named `mode` and accept strings as well as list (?) - no idea whether it has to be backward compatible, if not, it is better to rename it.

```
>>> import podman
>>> c1 = client.containers.create(name='test_noexec', image='test_image', volumes={"test_volume": {"bind": "/m", "options": ["rw", "noexec"]}}, command=['/bin/sleep', '12345'])
>>> c1.start()
>>> c2 = client.containers.create(name='test_exec', image='test_image', volumes={"test_volume": {"bind": "/m", "options": ["rw", "exec"]}}, command=['/bin/sleep', '12345'])
>>> c2.start()
>>> c1.attrs['HostConfig']['Binds']
['test_volume:/m:rw,noexec,rprivate,nosuid,nodev,rbind']
>>> c2.attrs['HostConfig']['Binds']
['test_volume:/m:rw,exec,rprivate,nosuid,nodev,rbind']

...

[root@host ~]# podman exec test_exec cp /bin/true /m/
[root@host ~]# podman exec -ti test_noexec /m/true
Error: exec failed: container_linux.go:380: starting container process caused: permission denied: OCI permission denied
[root@host ~]# podman exec -ti test_exec /m/true
[root@host ~]#
```

Ran 277 tests in 40.010s

FAILED (SKIP=3, errors=4, failures=4)

### DIFF
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -220,12 +220,12 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
                 a dictionary with the keys:
 
                 - bind: The path to mount the volume inside the container
-                - mode: Either rw to mount the volume read/write, or ro to mount it read-only.
+                - options: Options are options that the named volume will be mounted with
 
                 For example:
 
-                    {'/home/user1/': {'bind': '/mnt/vol2', 'mode': 'rw'},
-                     '/var/www': {'bind': '/mnt/vol1', 'mode': 'ro'}}
+                    {'/home/user1/': {'bind': '/mnt/vol2', 'options': ['rw']},
+                     '/var/www': {'bind': '/mnt/vol1', 'options': ['ro', 'noexec']}}
 
             volumes_from (List[str]): List of container names or IDs to get volumes from.
             working_dir (str): Path to the working directory.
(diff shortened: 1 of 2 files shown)


msisj:
> @msisj I understand your change is less smelly, but does break docker-py compatibility. Other areas of the bindings use kwargs to circumvent this issue. Maybe `extended_mode`?
> 
> /hold

I might have not understood what do you mean by `extended_mode`, could you elaborate ?
If backward compatibility is crucial, then all I could do it change `options` -> `mode` (I've renamed it from `mode` to `options` to be more precise) and support only passing options separated by comma (as it worked in docker).
Please let me know what kind of solution would you prefer.

rhatdan:
@msisj @jwhonce This needs a followup?

jwhonce:
@msisj 

Compatibility has been maintained to aid developers who have existing docker-py code use the podman service. Therefore the keyword `mode` is already defined and needed. I was suggesting adding another keyword `extended_mode` to allow you to provide the podman centric values. 

I would pass the union of the those keywords' values to the podman service.

msisj:
@jwhonce Thanks for clarification, now I fully understand what you've meant.
I'm preparing update from this PR then with `extended_mode` parameter along with a proper test case to test the functionality

msisj:
@jwhonce Can you re-check the pull request? You've marked it as hold and now it can't be merged

PR #145, opened 2021-11-05
Fix container name property

cryobry:
Get container name from list of Names

### DIFF
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -25,7 +25,10 @@ class Container(PodmanResource):
     def name(self):
         """str: Returns container's name."""
         with suppress(KeyError):
-            return self.attrs["Names"][0].lstrip("/")
+            if 'Name' in attrs:
+                return self.attrs["Name"].lstrip("/")
+            else:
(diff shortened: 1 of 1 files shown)


rhatdan:
@mwhahaha @jwhonce PTAL

mwhahaha (approved):
/lgtm

podman/domain/containers.py
```diff
@@ -25,7 +25,7 @@ class Container(PodmanResource):
     def name(self):
         """str: Returns container's name."""
         with suppress(KeyError):
-            return self.attrs["Name"].lstrip("/")
+            return self.attrs["Names"][0].lstrip("/")
```
mwhahaha, line 28: The ci is failing trying to check the name property of a container. I wonder if we need to handle both cases

e.g.

 if 'Name' in attrs:
    return self.attrs["Name"].lstrip("/")
else:
    return self.aters["Names"][0].lstrip("/")

mwhahaha (changes requested):
Actually per the CI it looks like this is not correct.  Maybe we need to handle both cases depending on version?

PR #147, opened 2021-11-09
Add possibility to expose device

msisj:

This change is temporary solution which allows exposing devices to container. But since podman REST API bases on minor/major numbers (which does not work as it should - path to device on host is needed to be pass and there is a minor/major number which makes no sense to pass too, since path on host is needed. What is more there is no possibility to change path on container [according to REST API](https://docs.podman.io/en/latest/_static/api.html#operation/ContainerCreateLibpod). On top of all these things, this REST API is non where near compliant with podman CLI (options are totally different).
Either way I've heard this is to be reworked in future. Here I just want to give possibility to expose devices in a simplest form.

```
>>> import podman
>>> c = client.containers.create(name='test', image='image:tag', devices=['/dev/hwrng', '/dev/tpm0'])
>>> c.start()
>>>
>>> c.attrs['HostConfig']['Devices']
[{'PathOnHost': '/dev/hwrng', 'PathInContainer': '/dev/hwrng', 'CgroupPermissions': ''}, {'PathOnHost': '/dev/tpm0', 'PathInContainer': '/dev/tpm0', 'CgroupPermissions': ''}]

...
[root@host ~]# podman exec test ls -l /dev/tpm0 /dev/hwrng
crw------- 1 root root 10, 183 Nov  9 13:40 /dev/hwrng
crw-rw---- 1 root 9999 10, 224 Nov  9 13:40 /dev/tpm0
```


jwhonce:
@mwhahaha Looking into changing the API to make this more useful.

jwhonce:
@msisj With podman4 we're making breaking changes to the API. I get changing creating devices on the list. /cc @mwhahaha 

Once that merges we'll get the python bindings tweaked. Thanks for bringing this to our attention. /cc @mheon @baude

jwhonce:
@msisj With podman4 we're making breaking changes to the API. I get changing creating devices on the list. /cc @mwhahaha 

Once that merges we'll get the python bindings tweaked. Thanks for bringing this to our attention. /cc @mheon @baude

rhatdan:
@msisj What is going on with this PR?

msisj:
@rhatdan 
When I've discussed this fix with @jwhonce he told me to just create a draft PR to show where the problem lies. As far as I understood it, it is not going to be merged (please @jwhonce correct me if I'm wrong).
Once I get that information I can either close it, or resolve conflicts and merge it.

jwhonce:
@msisj I have added functionality to API service in https://github.com/containers/podman/pull/13851 to read and parse the Path field to support this functionality.

msisj:
@jwhonce Thanks for the update. I'll adapt podman-py when your change is merged and is available for me.

msisj:
@jwhonce can you re-review the changes. I've noticed that your mentioned PR to podman repository was declined as the options were possible to achieve with previous implementation.
So I've changed code a bit and implemented a test for devices options.
I've also changed the manifest test because it was failing but I'm not 100% sure what was changed on podman side and if the change is proper, just wanted to tests to pass, so please check it out too.

rhatdan:
LGTM
@mwhahaha @jwhonce @cdoern PTAL

podman/tests/integration/test_container_create.py
```diff
@@ -135,6 +135,24 @@ def test_container_mounts(self):
                 )
             )
 
+    def test_container_devices(self):
```
cdoern, line 138: can you exec_run in this test and run `mount | grep "/dev/foo"` or something of the like to make sure the device is actually in the container not just in the config?

podman/domain/containers_create.py
```diff
@@ -422,6 +422,9 @@ def to_bytes(size: Union[int, str, None]) -> Union[int, None]:
             "work_dir": pop("working_dir"),
         }
 
+        for device in args.pop("devices", []):
+            params["devices"].append({"path": device})
```
cdoern, line 426: curious if we want to add other fields supported by the API... though this seems good enough for a first implementation.

cdoern (approved):
left one comment on the test. But other than that LGTM

podman/domain/containers_create.py
msisj: For now I just wanted to match to [the previously written help for devices](https://github.com/containers/podman-py/blob/main/podman/domain/containers_create.py#L56-L61). To be fair I'm not so sure whether the [rest of the parameters in API](https://github.com/containers/podman-py/blob/main/podman/domain/containers_create.py#L56-L61) works properly, especially taking into account [these comments](https://github.com/containers/podman-py/blob/main/podman/domain/containers_create.py#L293-L300).
So I've decided that unblocking passing devices is good for now, up until the `Major/Minor` is figured out

podman/tests/integration/test_container_create.py
msisj: Okay, I've extended test to check whether the device is visible in running container (it's not visible in mounts, so instead I've listed devices and compared with original major/minor numbers)

PR #148, opened 2021-11-12
Check if list is empty in the pythonic way

q0w:
1. The pythonic way to do it is from the [PEP 8 style guide](https://www.python.org/dev/peps/pep-0008/):
For sequences, (strings, lists, tuples), use the fact that empty sequences are false:
```python
# Correct:
if not seq:
if seq:
# Wrong:
if len(seq):
if not len(seq):
```

2. Use `[ ]` instead of `list()` (and its faster to initialize)

### DIFF
--- a/podman/api/tar_utils.py
+++ b/podman/api/tar_utils.py
@@ -23,8 +23,8 @@ def prepare_containerignore(anchor: str) -> List[str]:
         with ignore.open() as file:
             return list(
                 filter(
-                    lambda l: len(l) > 0 and not l.startswith("#"),
-                    list(line.strip() for line in file.readlines()),
+                    lambda l: l and not l.startswith("#"),
+                    (line.strip() for line in file.readlines()),
                 )
             )
     return []
@@ -102,7 +102,7 @@ def create_tar(
         name = pathlib.Path(name)
 
     if exclude is None:
-        exclude = list()
+        exclude = []
     else:
         exclude = exclude.copy()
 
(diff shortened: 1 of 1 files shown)


rhatdan:
@mwhahaha @jwhonce PTAL

rhatdan:
Thanks @q0w 
Looks like gating tests are failing.

podman/api/tar_utils.py
```diff
@@ -123,7 +123,7 @@ def _exclude_matcher(path: str, exclude: List[str]) -> bool:
     Note:
         FIXME Not compatible, support !, **, etc
     """
-    if len(exclude) == 0:
+    if not l:
```
jwhonce, line 126: "l" is undefined in this function.

jwhonce (changes requested):
I believe you wanted to use parameter exclude

podman/api/tar_utils.py
q0w: oh, yes, mb

mwhahaha (approved):
/lgtm

I ran pylint and there's a whole bunch of things to probably fix like this now. it's very unhappy. I might try and tackle that if i get some extra time

jwhonce:
> /lgtm
> 
> I ran pylint and there's a whole bunch of things to probably fix like this now. it's very unhappy. I might try and tackle that if i get some extra time

I use  "/usr/bin/pylint --msg-template='{abspath}:{line},{column}: {msg} ({symbol})'" --output-format=colorized --max-line-length=100 --disable=fixme --ignore tests podman. Combined with the .pylintrc in the root of the repository. I am curious what you are seeing re: pylint being unhappy. One outstanding issue is to dump the non-compatible API from the repository.

jwhonce (approved):

PR #149, opened 2021-11-12
Fixup lint items

mwhahaha:
This change switches to using fstrings for url paths and addresses
additional new linting rules.

Newer pylint versions have some new rules which we're triggering

* inconsistent-return-statements (2.7)
* arguments-renamed (2.9)
* unspecified-encoding (2.10)
* use-list-literal (2.10)
* use-dict-literal (2.10)
* consider-using-f-string (2.11)

Disabled few options until the CI version is upgraded past 2.9

* duplicate-code
* bad-option-value

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -142,7 +142,10 @@ disable=print-statement,
         comprehension-escape,
         no-self-use,
         no-member, # PyCQA/pylint#3157
-        too-many-public-methods # tests can have many members
+        too-many-public-methods, # tests can have many members
+        fixme, # don't warn on fixme stuff
+        duplicate-code, # current CI version doesn't honor the minimum for some reason
+        bad-option-value # needed until pylint version is increased to a version with it
 
 # Enable the message, report, category or checker with the given id(s). You can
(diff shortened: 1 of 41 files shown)


mwhahaha:
@jwhonce fyi this has the various things. If you use the packaged pylint it's older and missing things. If you run tox, you get the latest

jwhonce:
@mwhahaha Work looks good. Happy to merge once tests go green

Additionally,  I am thinking it is time to remove the old code/API and just support the compatible API. 

In my IDE, I am using the following options for pylint(2.9.6) "--msg-template='{abspath}:{line},{column}: {msg} ({symbol})'" --output-format=colorized --max-line-length=100 --disable=fixme --ignore tests podman

rhatdan:
@cevich I think we should update the images that these tests are based on to F35

cevich:
> I think we should update the images that these tests are based on to F35

Oh, woops, I missed this repo...

cevich:
...https://github.com/containers/podman-py/pull/158

PR #154, opened 2021-12-26
Support Python 3.10

jonathanunderwood:
This PR adds support for Python 3.10. This required:

- Catering for the movement of Iterator and Iterable in the collections package in Python 3.10 (backwards compatibility with earlier Python versions is maintained)
- Migrating from Nose to PyTest for the test runner. Nose doesn't run under Python 3.10 and is unmaintained. PyTesthas really taken the community mindshare in recent years, and so this seems like a reasonable move. Two tests required modification for this. Those modifications were bug fixes.

### DIFF
--- a/podman/tests/integration/test_containers.py
+++ b/podman/tests/integration/test_containers.py
@@ -2,7 +2,12 @@ import io
 import random
 import tarfile
 import unittest
-from collections import Iterator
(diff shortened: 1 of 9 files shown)


jwhonce:
LGTM

Issue #152, opened 2021-12-13
Context path error during image build


martin-endress:
According to [image.build](https://github.com/containers/podman-py/blob/b4bbc98b7dd983f3dd44145314ab98c3e96c4b6e/podman/domain/images_build.py#L28), path references the build context directory. This works for ".", but not for subdirectories.

Given the following directory tree:
```
src
├── context_dir
│   ├── Dockerfile
│   └── <further files in context>
└── example.py
```
and the following code:
```python
client.images.build(
  path="context_dir/",
  dockerfile="context_dir/Dockerfile",
)
```
results in:
`Build Error( "stat /var/tmp/libpod_builder2054394681/build/context_dir/Dockerfile: no such file or directory`

Versions:
```
"Version": "3.4.2",
"ApiVersion": "1.40",
"MinAPIVersion": "1.24",
```

My Workaround:
```python
os.chdir('context_dir')
clientimages.build(
  path=".",
  dockerfile="Dockerfile",
)
```

jonathanunderwood:
This is hopefully fixed by #155

PR #155, opened 2021-12-26
Fix image build method to handle containerfile location correctly

jonathanunderwood:
The docstring for BuildMixin.build() specifies that the dockerfile
argument is relative to the path/build context.

BuildMixin.build() calls api.tar_utils.prepare_containerfile() whose
docstring specifies that the return value is relative to the context
directory.

This commit fixes a bug where the absolute path to the containerfile
was being returned from prepare_containerfile() causing the subsequent
image build to fail as the containerfile was not found at the erroneous
path.

This commit also fixes an incorrect test for the prepare_containerfile()
function.


rhatdan:
@jwhonce @mwhahaha PTAL

jonathanunderwood:
note: fixes #152

jwhonce:
LGTM

PR #157, opened 2022-01-21
Fix unit tests expect for open

mwhahaha:
These tests are currently failing with:

Expected: open(PosixPath('/root/.config/containers/containers.conf'))
Actual: open(PosixPath('/root/.config/containers/containers.conf'),
encoding='utf-8')


mwhahaha:
rolled into #149

PR #160, opened 2022-01-26
Update CI to fail on Podman 4.0 errors

jwhonce:
CI has been testing against containers/podman/main to report issues.
With the release of Podman 4.0, any errors are now failures.

* Makefile target "test" and "lint" updated to use tox covering
  python 3.6, 3.8, 3.9 and 3.10.
* Added black formatting check to Makefile target "lint"
* Source code changes made to satisfy "tox -e black-format" and
  "tox -e pylint"

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -61,6 +61,7 @@ test_task:
 
     script:
         - ${SCRIPT_BASE}/enable_ssh.sh
+        - ${SCRIPT_BASE}/build_podman.sh
         - ${SCRIPT_BASE}/enable_podman.sh
         - ${SCRIPT_BASE}/test.sh
 
--- a/Makefile
+++ b/Makefile
@@ -8,25 +8,23 @@ DESTDIR ?=
 EPOCH_TEST_COMMIT ?= $(shell git merge-base $${DEST_BRANCH:-main} HEAD)
 HEAD ?= HEAD
 
-export PODMAN_VERSION ?= "3.2.0"
+export PODMAN_VERSION ?= "4.0.0"
 
 .PHONY: podman
 podman:
 	rm dist/* || :
-	python -m pip install --user -r requirements.txt
+	$(PYTHON) -m pip install --user -r requirements.txt
 	PODMAN_VERSION=$(PODMAN_VERSION) \
 	$(PYTHON) setup.py sdist bdist bdist_wheel
 
 .PHONY: lint
-lint:
-	$(PYTHON) -m pylint podman || exit $$(($$? % 4));
+lint: tox
+	$(PYTHON) -m tox -e black,pylint
 
 .PHONY: tests
-tests:
-	python -m pip install --user -r test-requirements.txt
-	DEBUG=1 coverage run -m unittest discover -s podman/tests
-	coverage report -m --skip-covered --fail-under=80 --omit=./podman/tests/* --omit=.tox/* \
-	--omit=/usr/lib/* --omit=*/lib/python*
+tests: tox
+	# see tox.ini for environment variable settings
+	$(PYTHON) -m tox -e pylint,coverage,py36,py38,py39,py310
 
 .PHONY: unittest
 unittest:
@@ -37,6 +35,12 @@ unittest:
 integration:
 	coverage run -m unittest discover -s podman/tests/integration
 	coverage report -m --skip-covered --fail-under=80 --omit=./podman/tests/* --omit=.tox/* --omit=/usr/lib/*
+
+.PHONY: tox
+tox:
+	-dnf install -y python3 python3.6 python3.8 python3.9
+	# ensure tox is available. It will take care of other testing requirements
+	$(PYTHON) -m pip install --user tox
 
 .PHONY: test-release
 test-release: SOURCE = $(shell find dist -regex '.*/podman-[0-9][0-9\.]*.tar.gz' -print)
--- a/contrib/cirrus/build_podman.sh
+++ b/contrib/cirrus/build_podman.sh
@@ -2,13 +2,9 @@
 
 set -xeo pipefail
 
-mkdir -p "$GOPATH/src/github.com/containers/"
-cd "$GOPATH/src/github.com/containers/"
+systemctl stop podman.socket || :
 
-systemctl stop podman.socket ||:
 dnf erase podman -y
-git clone https://github.com/containers/podman.git
+dnf copr enable rhcontainerbot/podman-next -y
+dnf install podman -y
 
-cd podman
-make binaries
-make install PREFIX=/usr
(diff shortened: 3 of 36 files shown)


Makefile
```diff
@@ -8,25 +8,22 @@ DESTDIR ?=
...
-	python -m pip install --user -r requirements.txt
+	$(PYTHON) -m pip install --user -r requirements.txt
 	PODMAN_VERSION=$(PODMAN_VERSION) \
 	$(PYTHON) setup.py sdist bdist bdist_wheel
 
 .PHONY: lint
-lint:
-	$(PYTHON) -m pylint podman || exit $$(($$? % 4));
+lint: tox
+	tox -e black,pylint
```
mwhahaha, line 22: Looks like it can't find tox when you pip install --user tox.  I think it's a PATH problem where $HOME/.local/bin is not in root's path by default.

  WARNING: The script virtualenv is installed in '/root/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.
  WARNING: The scripts tox and tox-quickstart are installed in '/root/.local/bin' which is not on PATH.
  Consider adding this directory to PATH or, if you prefer to suppress this warning, use --no-warn-script-location.


Seems to work fine locally as non-root.

mwhahaha: looks like you could update this to "$(PYTHON) -m tox -e black,pylint" instead

mwhahaha (changes requested):

Makefile
```diff
@@ -8,25 +8,22 @@ DESTDIR ?=
...
+	$(PYTHON) -m tox -e black,pylint
 
 .PHONY: tests
-tests:
-	python -m pip install --user -r test-requirements.txt
-	DEBUG=1 coverage run -m unittest discover -s podman/tests
-	coverage report -m --skip-covered --fail-under=80 --omit=./podman/tests/* --omit=.tox/* \
-	--omit=/usr/lib/* --omit=*/lib/python*
+tests: tox
+	DEBUG=1 $(PYTHON) -m tox -e pylint,coverage,py36,py38,py39,py310
```
mwhahaha, line 26: coverage should run the tests so you probably don't need to add py3* in this list

jwhonce: @mwhahaha When I remove the py*, tox only runs for the default interpreter rather than iterating over the versions.

mwhahaha: Right but it was failing in CI because some of them weren't installed.

jwhonce: @mwhahaha Yup, @cevich is building a new image with all the interpreters (as I type) and changes in the podman build tool chain is also tripping me up. Will get there. :+1:

cevich:
Okay, so it appears like the new image "works" (getting actual test errors now).  That means the dedicated image I'd like to build, will most likely work well here.  Thanks for trying that.

cevich:
@jwhonce I've got the new dedicated image coming in via https://github.com/containers/podman-py/pull/161  Assuming that passes (and no further package changes are needed), you'll want to merge/rebase on that to use the new image here.

contrib/cirrus/build_podman.sh
```diff
@@ -2,13 +2,43 @@
...
+FILE_DIR=$TARGET_DIR/podman-v4.0.0-dev
+
+curl -SOsl \
+    --output-dir $TARGET_DIR \
+    --url "https://api.cirrus-ci.com/v1/artifact/github/containers/podman/Build%20for%20fedora-35/gosrc/$PODMAN_TARBALL"
+
+tar zxf $TARGET_DIR/$PODMAN_TARBALL
+
+SRC_DIR=$FILE_DIR/usr/bin
+install -m 755 $SRC_DIR/* /usr/bin
```
cevich, line 31: `$FILE_DIR` is going to change anytime someone touches podman's `version.go`.  But there should only/ever be a single subdirectory.  So it's probably safer to install from `$TARGET_DIR/podman-v*/usr/bin/*` and let bash figure it out.  Similar for below.

cevich: Another maybe easier thought...Lokesh is building a test RPM in podman CI, but we're not currently creating an artifact for it.  I could make that (simple) change, and get you the URL of the test `rpm` instead.

cevich: https://github.com/containers/podman/pull/13344

cevich: Lokesh has an even easier solution:  Apparently the latest from main is available in COPR, so just:

`dnf copr enable rhcontainerbot/podman-next -y`

and `dnf upgrade -y podman` :smile:

jwhonce: @cevich I'll update the cirrus support scripts to pull from copr

cevich:
Oh you got it working!

.cirrus.yml
```diff
@@ -61,6 +61,7 @@ test_task:
 
     script:
         - ${SCRIPT_BASE}/enable_ssh.sh
+        - ${SCRIPT_BASE}/build_podman.sh
```
cevich, line 64: minor-nit: This should probably be renamed `upstream-podman.sh` or similar

pyproject.toml
```diff
@@ -20,11 +20,18 @@ profile = "black"
...
+    "pyxdg>=0.26",
     "requests>=2.24",
+    "setuptools>=46.4",
+    "sphinx",
     "toml>=0.10.2",
     "urllib3>=1.24.2",
-    "pyxdg>=0.26",
-    "setuptools>=46.4",
     "wheel",
 ]
```
cevich, line 30: Minor-nit: It appears (maybe I'm wrong) the `requires` block needs to match `requirements.txt`?  These are the kinds of things which will drive developers crazy if they ever get out of sync for some reason.   Short of some kind of script/automation, consider adding comments to both files indicating developers changing one, should always also change the other.  It's kind of a "cheat" but may save someone hours of debugging.

setup.cfg
```diff
@@ -32,10 +32,10 @@ include_package_data = True
 python_requires = >=3.6
 test_suite =
 install_requires =
+    pyxdg>=0.26
     requests>=2.24
     toml>=0.10.2
     urllib3>=1.24.2
-    pyxdg>=0.26
```
cevich, line 38: Ugh...really? Please tell me this doesn't need to also match `requirements.txt` and `pyproject.toml`?

If so...same 'add a comment' suggestion should apply here too.

-*sigh*-

cevich (approved):
LGTM short of all files ending in `.py` :grin:   Made some comments about dependency-list duplication :vomiting_face: not sure if anything can be done...I'm assuming you're more of an expert than I and have already considered any/all ramifications :grinning:

PR #172, opened 2022-04-01
Allow relabeling mounts

msisj:

Previously relabeling mounts was only possible via using `propagation` key, which sounds terrible.
Refactored options passed to mounts to be able to add more without adding if for each of them.

Also added tests for bind and tmpfs mount and added examples (as they are not obvious, since there is no working documentation right now).

### DIFF
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -101,6 +101,27 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
             mounts (List[Mount]): Specification for mounts to be added to the container. More
                 powerful alternative to volumes. Each item in the list is expected to be a
                 Mount object.
+                For example :
+                 [
+                    {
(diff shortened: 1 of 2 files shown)


msisj:
Ok I do not understand why this is failing on pylint.
I'd like to emphasise the fact that I did not touch the mentioned files.
How did it pass previous PR then?

jwhonce:
@msisj I understand your frustration, when it was suggested I thought tox was going to be very helpful but it appears to be pulling in newer/different versions of pylint. Which in turns implies what was acceptable code becomes not-acceptable. If you rebase, some of these flagged errors are now ignored on main. That has been my solution either ignore the error if it is too pedantic or fix the code.

msisj:
@jwhonce Perfect, it worked after backmerge, thanks!

rhatdan:
LGTM

PR #176, opened 2022-05-09
Allow pushing images with tag different than latest

msisj:
I've made this PR because it looks like that `tag` parameter was skipped when pushing image to registry.


giuseppe:
@jwhonce PTAL

PR #183, opened 2022-06-15
CoPR: fix CentOS 8 Stream build

lsm5:
Signed-off-by: Lokesh Mandvekar <lsm5@fedoraproject.org>

c8s fails without this change: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/4537996/

c8s successful with this change: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/4538017/ 

@rhatdan @jwhonce PTAL.


rhatdan:
LGTM

Issue #102, opened 2021-05-29
podman client hangs when starting container


jbpratt:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When trying to run a container (with possibly invalid configuration) via `podman-py`, causes the system service to hang. Killing the client call will still leave the service spinning.
```
❯ python -c 'import podman;podman.PodmanClient().containers.run(detach=True, image="docker.io/library/nginx", mounts=[{"target": "tests123", "source": "tests/", "mode": "bind"}], tty=True)'
^CTraceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/tmp/xyz/venv/lib/python3.9/site-packages/podman/domain/containers_run.py", line 66, in run
    container.start()
  File "/tmp/xyz/venv/lib/python3.9/site-packages/podman/domain/containers.py", line 350, in start
    response = self.client.post(
  File "/tmp/xyz/venv/lib/python3.9/site-packages/podman/api/client.py", line 304, in post
    return self._request(
  File "/tmp/xyz/venv/lib/python3.9/site-packages/podman/api/client.py", line 402, in _request
    self.request(
  File "/tmp/xyz/venv/lib/python3.9/site-packages/requests/sessions.py", line 542, in request
    resp = self.send(prep, **send_kwargs)
  File "/tmp/xyz/venv/lib/python3.9/site-packages/requests/sessions.py", line 655, in send
    r = adapter.send(request, **kwargs)
  File "/tmp/xyz/venv/lib/python3.9/site-packages/requests/adapters.py", line 439, in send
    resp = conn.urlopen(
  File "/tmp/xyz/venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/tmp/xyz/venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 445, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/tmp/xyz/venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 440, in _make_request
    httplib_response = conn.getresponse()
  File "/usr/lib/python3.9/http/client.py", line 1345, in getresponse
    response.begin()
  File "/usr/lib/python3.9/http/client.py", line 307, in begin
    version, status, reason = self._read_status()
  File "/usr/lib/python3.9/http/client.py", line 268, in _read_status
[...]

PR #185, opened 2022-06-30
Fix issue with container's run method

mrbazzan:
See #102


rhatdan:
Looks like this is failing a linting issue?

mrbazzan:
> Looks like this is failing a linting issue?

I have no idea why this is failing...I did not touch the mentioned files.

mrbazzan:
> I have no idea why this is failing...I did not touch the mentioned files.

I reformatted the `podman/api/typing_extensions.py` using `tox -e black-format`, and now the `Gating test` passes

rhatdan:
I think you want this to say exited or running.  Now it just says wait for the container to exit.  Which will cause it to hang for a long time.

mrbazzan:
> I think you want this to say exited or running. Now it just says wait for the container to exit. Which will cause it to hang for a long time.

Thanks. I've updated it.
Do you think I should update the `.wait()` doc in a different commit?

rhatdan:
Yes SGTM
Does this fix your issue?
/approve
LGTM
@jwhonce @mwhahaha PTAL

mrbazzan:
> Does this fix your issue?

Yes, the first part.

The other issue is this [line](https://github.com/containers/podman-py/blob/21a48cb81808b7e479630ecb612df32df4ad3550/podman/domain/containers_run.py#L80).
`.wait()` return an integer which is not subscriptable.

Issue #184, opened 2022-06-29
Question/Information required on some aspects of podman-py module


mrbazzan:
Hello,

I'm using `podman-py` for some container action, and I have a couple of questions;

1) Is it possible to get real-time logging of what is happening during image build(from `PodmanClient().images.build`) in podman-py. I noticed that the option is available for pull(from `PodmanClient().images.pull`) with `stream=True`

2) Does the PodmanClient support some kind of event loop that could be subscribed to, and receive the “stdout/stderr” as events regardless of podman command/method (build/pull) used?

3) There is an issue (I think) with this [line](https://github.com/containers/podman-py/blob/21a48cb81808b7e479630ecb612df32df4ad3550/podman/domain/containers_run.py#L80). The API doesn’t respect the REST endpoint, the REST returns a simple integer instead of JSON/dict. 
This prompts a `TypeError, int object is not subscriptable`

![libpod](https://user-images.githubusercontent.com/64538336/176464913-effcb5bb-35de-4fe7-8c9d-3bfedd0bfa91.png)

rhatdan:
SGTM

PR #187, opened 2022-07-12
Fix TypeError in container's wait method

mrbazzan:
For the doc change: list of string does not work with `environment` argument

Fixes No.3 in #184, `.wait()` returns an integer not a dictionary

### DIFF
--- a/podman/domain/containers_run.py
+++ b/podman/domain/containers_run.py
@@ -77,7 +77,7 @@ class RunMixin:  # pylint: disable=too-few-public-methods
         if log_type in ("json-file", "journald"):
             log_iter = container.logs(stdout=stdout, stderr=stderr, stream=True, follow=True)
 
-        exit_status = container.wait()["StatusCode"]
+        exit_status = container.wait()
         if exit_status != 0:
             log_iter = None
             if not kwargs.get("auto_remove", False):
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -65,9 +65,11 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
             dns_search (List[str]): DNS search domains.
             domainname (Union[str, List[str]]): Set custom DNS search domains.
             entrypoint (Union[str, List[str]]): The entrypoint for the container.
-            environment (Union[Dict[str, str], List[str]): Environment variables to set inside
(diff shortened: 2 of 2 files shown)


rhatdan:
LGTM
@mwhahaha @jwhonce @cdoern PTAL

podman/domain/containers_run.py
```diff
@@ -77,7 +77,7 @@ def run(
         if log_type in ("json-file", "journald"):
             log_iter = container.logs(stdout=stdout, stderr=stderr, stream=True, follow=True)
 
-        exit_status = container.wait()["StatusCode"]
+        exit_status = container.wait()
         if exit_status != 0:
```
cdoern, line 81: how does this check still work if you remove the ["StatusCode"] aspect? now you are checking `if DICT != 0`

rhatdan: def wait(self, **kwargs) -> Dict[Literal["StatusCode", "Error"], Any]:
Looks like to original code is correct

sylencecc: Wait wait wait, there is definitely an issue here.
By default, `wait()` in podman-py uses the libpod operation  [ContainerwaitLibPod](https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerWaitLibpod), which returns a JSON-encoded integer. For that case, the type annotation
```
def wait(self, **kwargs) -> Dict[Literal["StatusCode", "Error"], Any]:
```
isn't correct. That annotation matches the podman compat API, where [ContainerWait](https://docs.podman.io/en/latest/_static/api.html#tag/containers-(compat)/operation/ContainerWait) returns a JSON object with the expected Keys `Error` and `StatusCode`.
I just discovered podman-py and don't know the code base enough yet to fix it myself, but would recommend to look into that issue again. I got here having experienced the same issue as in #201

rhatdan: @cdoern PTAL

jonathanunderwood:
@rhatdan wondering why this was closed, as the issue is still present. As @sylencecc points out, the proposed fix is correct, but the type annotation for the return type of`wait()` is wrong and also needs fixing.

rhatdan:
I don't remember but might have been lack of activity.

Care to open a new PR based on this?

PR #196, opened 2022-07-29
podman-py login support

### LINKED ISSUE #189: login is not implemented in SystemManager
Using podman-py 4.0.0

https://github.com/containers/podman-py/blob/f46011ca174215eee26321ee0aee22ab510e8ec5/podman/domain/system.py#L37

I don't know if this is intentional or not, but if it is intentional a comment or a `raise NotImplementedError()` would be useful to have.

cdoern:
use the compat auth endpoint to allow users to access login from within podman-py

resolves #189

Signed-off-by: Charlie Doern <cdoern@redhat.com>

### DIFF
--- a/podman/domain/system.py
+++ b/podman/domain/system.py
@@ -3,6 +3,7 @@ import logging
 from typing import Any, Dict, Optional
 
 from podman.api.client import APIClient
+from podman import api
 
 logger = logging.getLogger("podman.system")
 
@@ -40,8 +41,8 @@ class SystemManager:
         password: Optional[str] = None,
         email: Optional[str] = None,
         registry: Optional[str] = None,
-        reauth: bool = False,
-        dockercfg_path: Optional[str] = None,
+        reauth: Optional[bool] = False,  # pylint: disable=unused-argument
+        dockercfg_path: Optional[str] = None,  # pylint: disable=unused-argument
     ) -> Dict[str, Any]:
         """Log into Podman service.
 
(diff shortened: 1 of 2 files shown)


cdoern:
not sure what to do with `reauth` or `dockercfg_path` @jwhonce

rhatdan:
dockercfg_path would be on the server side or require the client to handle it.  I do not know what reauth is?

podman/domain/system.py
```diff
@@ -51,11 +50,23 @@ def login(
...
-            dockercfg_path: Path to custom configuration file.
-                Default: $HOME/.config/containers/config.json
         """
 
+        payload = {
+            "username": username,
+            "password": password,
+            "email": email,
+            "serveraddress": registry,
+        }
```
jwhonce, line 60:
```suggestion
        }
        payload = api.prepare_body(payload)
```

will clean up any field that are None from caller.

podman/domain/system.py
```diff
@@ -51,11 +50,23 @@ def login(
             email: Registry account email address
             registry: URL for registry access. For example,
                 https://quay.io/v2
-            reauth: If True, refresh existing authentication. Default: False
```
jwhonce, line 54: docker-py caches previous authorization results and with reauth False the code just reads the cache rather than round trip to the service.

Rather than delete the keyword, change the text to inform developers it is ignored.

podman/domain/system.py
```diff
@@ -51,11 +50,23 @@ def login(
...
-            dockercfg_path: Path to custom configuration file.
```
jwhonce, line 55: This is the path `reauth` records the authentication information. it can also be ignored if not used.

Rather than delete the keyword, change the text to inform developers it is ignored.

jwhonce (changes requested):

cdoern:
@jwhonce it seems that the API does not like the `api.prepare_body` conversion:  `lookup fake_registry: no such host' not found in 'failed to parse request: json: cannot unmarshal string into Go value of type types.AuthConfig'`

cdoern:
@jwhonce @lsm5 @rhatdan PTAL, I'd like to get this in before I cut a release tmrw

rhatdan:
LGTM
/approve

rhatdan:
@baude PTAL

lsm5 (approved):
LGTM

cdoern:
I think @jwhonce 's change request is blocking

jwhonce:
> @jwhonce it seems that the API does not like the `api.prepare_body` conversion: `lookup fake_registry: no such host' not found in 'failed to parse request: json: cannot unmarshal string into Go value of type types.AuthConfig'`

Sounds like the handler is not unmarshaling the data correctly.

cdoern:
> > @jwhonce it seems that the API does not like the `api.prepare_body` conversion: `lookup fake_registry: no such host' not found in 'failed to parse request: json: cannot unmarshal string into Go value of type types.AuthConfig'`
> 
> Sounds like the handler is not unmarshaling the data correctly.

fixed now, I was necessarily using json.Dumps

PR #205, opened 2022-09-01
adding pass-through layers, output and outputformat to building images.

ericpershey:
While working to replicate local podman image build commands in python, the 3 options, layers, output and outputformat were removed from kwargs in images.build.  These options are now passthrough to the REST API and replicated the caching that is done with the command line.  I extracted these options from the REST call made below and once added as pass-through, allowed caching to work.  Note, I am new to this so feedback would be great.  I could hardly find the output and outputformat options in any documentation except for buildah build under podman.

```
podman build -f /.../test_pdw_dockerfile.ubuntu2004 -t "pershey/test_pdw_dockerfile.ubuntu2004:latest"
```

Signed-off-by: Eric Pershey <pershey@anl.gov>

### DIFF
--- a/podman/domain/images_build.py
+++ b/podman/domain/images_build.py
@@ -165,9 +165,9 @@ class BuildMixin:
             "squash": kwargs.get("squash"),
             "t": kwargs.get("tag"),
             "target": kwargs.get("target"),
-            'layers': kwargs.get("layers"),
-            'output': kwargs.get("output"),
-            'outputformat': kwargs.get("outputformat"),
+            "layers": kwargs.get("layers"),
+            "output": kwargs.get("output"),
+            "outputformat": kwargs.get("outputformat"),
         }
 
         if "buildargs" in kwargs:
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@cdoern @jwhonce @mwhahaha PTAL

rhatdan:
@baude PTAL

rhatdan:
@cdoern @jwhonce @mwhahaha PTAL

podman/domain/images_build.py
```diff
@@ -162,6 +165,9 @@ def _render_params(kwargs) -> Dict[str, List[Any]]:
             "squash": kwargs.get("squash"),
             "t": kwargs.get("tag"),
             "target": kwargs.get("target"),
+            'layers': kwargs.get("layers"),
```
cdoern, line 168: can you follow the double quoted format of the above items?

cdoern (commented):
one nit

podman/domain/images_build.py
ericpershey: Done

rhatdan:
Thanks @ericpershey 
LGTM

PR #222, opened 2022-11-02
Clean up pylint configuration

jwhonce:
* Removed configuration items from .pylintrc that are no longer supported and generating warnings.
* Locked "pylint" tox environment to python 3.10. Python 3.11 is not supported by Pylint 2.15.5. Dill dependency needs to be updated.


rhatdan:
LGTM
@cdoern @mwhahaha PTAL

PR #224, opened 2022-12-20
Added port binding range

msisj:

Added support for `range` in `ports`.
The range is available at REST API but was not supported in podman-py.
I've tried to keep the backward compatibility (hopefully it works, previously written test case works at the very least).
Unfortunately it was impossible to add `range` without adding new syntax for ports (using dictionary instead of None or int or tuple or list).
The newly added syntax should work in the same manner as the previously used one, but with addition of `range` keyword.

### DIFF
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -149,7 +149,8 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
             pids_limit (int): Tune a container's pids limit. Set -1 for unlimited.
             platform (str): Platform in the format os[/arch[/variant]]. Only used if the method
                 needs to pull the requested image.
-            ports (Dict[str, Union[int, Tuple[str, int], List[int]]]): Ports to bind inside
+            ports (Dict[str, Union[int, Tuple[str, int], List[int],
+                      Dict[str, Union[int, Tuple[str, int], List[int]]]]]): Ports to bind inside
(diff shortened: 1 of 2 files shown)


msisj:
@jwhonce @baude I keep getting error on Gating test 

> Failed to start: INVALID_ARGUMENT: Operation with name "operation-1671539560903-5f041a1c171ad-d9a05ac7-54e6698d" failed with status = HttpJsonStatusCode{statusCode=INVALID_ARGUMENT} and message = BAD REQUEST

Could you help me with understanding what is going wrong?

robbmanes:
I suspect something similar to what is happening [in this issue](https://github.com/containers/buildah/pull/4109/commits/36d95b0503960d2268f5aeacec9006de114b3bcf) is happening here, requiring us to change the gating image - not a problem with the PR, but with the image in the gating test.

I don't have write access to this repo to test different versions in cirrus so I'll ping others about it to take a look/fix it.

rhatdan:
LGTM

cevich:
CI failure fixed and I added a cirrus-cron job to run daily on `main`. That should prevent this kind of failure from happening again.

Note: The gating task is now failing with lint problems, I'm assuming those are legit issues.  Poke me again if otherwise.

msisj:
Thanks for checking it out, now everything works as expected

PR #225, opened 2023-01-05
Update cirrus image

rhatdan:
Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>


.cirrus.yml
```diff
@@ -14,12 +14,12 @@ env:
...
     ####
-    FEDORA_NAME: "fedora-36"
-    PRIOR_FEDORA_NAME: "fedora-35"
+    FEDORA_NAME: "fedora-37"
+    PRIOR_FEDORA_NAME: "fedora-36"
     UBUNTU_NAME: "ubuntu-2204"
 
     # Google-cloud VM Images
-    IMAGE_SUFFIX: "c6193881921355776"
+    IMAGE_SUFFIX: "c5069932136759296"
```
lsm5, line 22: @rhatdan just curious where you got the image id from

lsm5: AFAICT the latest image built was at https://github.com/containers/automation_images/pull/241 , which has `cache	fedora-podman-py	c4987131005763584` . So I think that would be the one to try unless I'm missing something.

rhatdan: from podman.

rhatdan: Trying yours

umohnani8:
Is this ready? @lsm5 @cevich PTAL

cevich:
These images have the "bad" kernel in them (it allows multiple processes to bind to the same port).  Not sure if we want to bump the images again or call the tests passing as a sign of "good enough".

Ref: https://github.com/containers/podman/pull/16525#issuecomment-1377651915
and  https://github.com/containers/automation_images/pull/245

cevich:
After thinking another minute...let's just merge this.  The tests are green, if another image update is found to be needed, it can be done.  No reason to hold back progress.

PR #229, opened 2023-01-26
Bump to v4.4.0

rhatdan:
Bump the podman-py version to 4.4.0 to be in lockstep with podman versions.

Signed-off-by: Daniel J Walsh <dwalsh@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ DESTDIR ?=
 EPOCH_TEST_COMMIT ?= $(shell git merge-base $${DEST_BRANCH:-main} HEAD)
 HEAD ?= HEAD
 
-export PODMAN_VERSION ?= "4.3.0"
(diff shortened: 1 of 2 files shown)


rhatdan:
@mwhahaha @jwhonce @cdoern Do I need to wait until podman 4.4 ships for this to pass tests?

rhatdan:
@cevich @lsm5 can we get an updated image with podman 4.4 installed?

cevich:
The CI VM Image used here is shared with multiple repos. Doing a one-off for a lesser-utilized repo. doesn't seem like a wise use of time.  Can this wait until 4.4 makes it into Fedora's updates-testing repo?  At that time it will be trivial to get it into an image.

rhatdan:
SGTM

PR #230, opened 2023-02-03
Setup secrets using IDs in container create.

timc:
Create the JSON objects for secrets in container create requests from either podman.domain.secrets.Secret objects or string IDs.


timc:
Ignore

PR #241, opened 2023-02-20
[v4.4-rhel] Revert "Use modern tomllib/tomli modules for reading TOML files"

umohnani8:
This reverts commit c5a356fb4ea8a6fb66a6d20bdc2c9cffe615028b.

Signed-off-by: Urvashi Mohnani <umohnani@redhat.com>

### DIFF
--- a/podman/domain/config.py
+++ b/podman/domain/config.py
@@ -1,20 +1,16 @@
 """Read containers.conf file."""
-import sys
 import urllib
 from pathlib import Path
 from typing import Dict, Optional
 
 import xdg.BaseDirectory
 
-from podman.api import cached_property
+try:
+    import toml
+except ImportError:
+    import pytoml as toml
 
-if sys.version_info >= (3, 11):
-    from tomllib import loads as toml_loads
-else:
-    try:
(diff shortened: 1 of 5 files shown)


umohnani8:
@rhatdan @lsm5 PTAL
Will cut a v4.4.1 after this

lsm5:
@umohnani8 do we need a -rhel branch given that we're reverting to toml on the main branch itself?

lsm5:
> @umohnani8 do we need a -rhel branch given that we're reverting to toml on the main branch itself.

If the branch was created for other reasons, then don't mind me :D

umohnani8:
The branch is created to track any backports that may need to go into 4.4 so we can cut new 4.4.x from there. So even though we are fixing this in main too, I would like to have a commit to the rhel-4.4 branch so we can cut a 4.4.1 from there instead of main to keep things aligned.

lsm5 (approved):
/lgtm

lsm5:
for whatever reason, mergebot didn't merge even an hour after PR was good to go.

PR #242, opened 2023-02-21
[CI:BUILD] Packit: initial enablement

lsm5:
This commit will run COPR builds on every PR against all active
releases of CentOS Stream and Fedora, thus allowing buildability checks before the
PR merges.
    
On a new upstream release, the `job: propose_downstream` will get
triggered by packit service and update Fedora's official packaging
sources.
    
The jobs `koji_build` and `bodhi_update` work on the official Fedora
sources downstream at https://src.fedoraproject.org/rpms/python-podman
   
The `trigger: pull_request` builds are done on a custom COPR project:
`rhcontainerbot/packit-builds`.
Ref: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/
   
The build targets are set in the copr itself, so we don't need to
explicitly mention them in `.packit.yaml`, making upstream configuration
a lot simpler.
    
The `trigger: commit` builds are done on a second custom COPR project:
`rhcontainerbot/podman-next`.
Ref: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/
The `podman-next` copr is meant for end user consumption.
    
With this PR, downstream Fedora sources will get synced with
`rpm/python-podman.spec` present upstream, so upstream will remain
independent of downstream and the packaging process will become a part of
upstream maintenance. The downstream tasks add zero extra cost to
upstream release process (except filing a github issue on downstream
task failure).


.packit.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Get Version from HEAD
+VERSION=$(grep '^__version__' podman/version.py | cut -d\" -f2 | sed -e 's/-/~/')
+
+# Generate source tarball from HEAD
+git archive --prefix=$PACKAGE-$VERSION/ -o $PACKAGE-$VERSION.tar.gz HEAD
+
+# RPM Spec modifications
+
+# Update Version in spec with Version from Cargo.toml
```
jwhonce, line 19:
```suggestion
# Update Version in spec with Version from version.py
```

.packit.yaml
```diff
@@ -0,0 +1,25 @@
...
+specfile_path: python-podman.spec
+
+jobs:
+  - job: copr_build
+    trigger: pull_request
+    owner: rhcontainerbot
+    project: packit-builds
+    enable_net: true
+    srpm_build_deps:
+      - cargo
```
jwhonce, line 17:
```suggestion
      - cargo
```
Is cargo now a dependency?

jwhonce (changes requested):

.packit.sh
lsm5: copy pasta from aardvark-dns :D

.packit.yaml
lsm5: copy pasta from aardvark-dns :D

lsm5:
@umohnani8 @rhatdan PTAL

/cc @jwhonce @cevich @edsantiago 

python rpms are `noarch` on fedora and rhel, so I could potentially reduce the number of tasks in the CI, but that would mean complicating the packit config by quite a bit. So, I'd rather keep aarch64 and x86_64 tasks as they are set by default in `rhcontainerbot/packit-builds` copr repo: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/build/6006581/

edsantiago:
Belated LGTM. Nice work.

PR #244, opened 2023-02-22
Fix RHEL and Fedora builds. Add pytoml fallback and fix setuptools incompatability.

gotmax23:


gotmax23:
Force pushed to add DCO signoffs

gotmax23:
Force pushed to add a necessary specfile change that I stashed and forgot to push :(

thesamesam:
cc @mgorny

gotmax23:
Force pushed to fix black linting failures

podman/domain/config.py
```diff
@@ -1,17 +1,20 @@
...
-
 from podman.api import cached_property
 
+if sys.version_info >= (3, 11):
+    from tomllib import loads as toml_loads
+else:
+    try:
+        from tomli import loads as toml_loads
+    except ImportError:
+        from toml import loads as toml_loads
```
mgorny: You may want to remove the sentence about "backwards compatibility" since it's preserved in the end.

mgorny (approved):
LGTM, thanks!

podman/domain/config.py
gotmax23: I could just squash the first two commits, but I'd prefer to keep it like. I want to keep the metadata from the original commit and properly attribute you.

lsm5:
@gotmax23 tried a copr build with this PR and el9 builds have failed saying `ModuleNotFoundError: No module named 'xdg'`. https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/build/5559996/ . It failed for f36 too, but f36 doesn't matter much anymore.

gotmax23:
🙁. I tested this on rawhide and el8 but not in between. Older versions of setuptools are apparently unable to parse `attr: podman.version.__version__` without actually importing `podman.version`. We could revert https://github.com/containers/podman-py/pull/244/commits/cbc3b46cb7aed954abbc098a542fb6df01ad70ac, but that would make the RHEL `sed`s more complicated and adds unnecessary duplication. Can we just set `version` in setup.cfg manually instead of using `attr`?

gotmax23:
``` diff
diff --git a/setup.cfg b/setup.cfg
index 1d33e3e..0c89b66 100644
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = podman
-version = attr: podman.version.__version__
+version = 4.4.0
 author = Brent Baude, Jhon Honce
 author_email = jhonce@redhat.com
 description = Bindings for Podman RESTful API
```

lsm5:
> slightly_frowning_face. I tested this on rawhide and el8 but not in between. Older versions of setuptools are apparently unable to parse `attr: podman.version.__version__` without actually importing `podman.version`. We could revert [cbc3b46](https://github.com/containers/podman-py/commit/cbc3b46cb7aed954abbc098a542fb6df01ad70ac), but that would make the RHEL `sed`s more complicated and adds unnecessary duplication. Can we just set `version` in setup.cfg manually instead of using `attr`?

I'm cool with anything as long as we have successful builds. Let me know whenever I can retest copr builds.

rhatdan:
LGTM

gotmax23:
Well, I wanted to address @lsm5's feedback in this PR, but I guess that can be a follow up.

gotmax23:
Submitted https://github.com/containers/podman-py/pull/247

PR #247, opened 2023-02-28
Specify version verbatim in setup.cfg

gotmax23:
This restores compatibility with older setuptools versions.

NB, the version needs to be kept in sync between setup.cfg and version.py.

Relates: https://github.com/containers/podman-py/pull/244

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = podman
-version = attr: podman.version.__version__
+version = 4.4.1
 author = Brent Baude, Jhon Honce
 author_email = jhonce@redhat.com
 description = Bindings for Podman RESTful API


gotmax23:
https://copr.fedorainfracloud.org/coprs/gotmax23/python-podman/build/5580484/

rhatdan:
Is there a way to test if they are out of sync?

umohnani8:
Changes LGTM
@lsm5 PTAL and merge

gotmax23:
> Is there a way to test if they are out of sync?

You can use a tool like [bump2version](https://pypi.org/project/bump2version/) to easily bump versions in both places if you'd like.

rhatdan:
I don't care how it is done. I just want to make sure we prevent human mistakes.

gotmax23:
It's hard for me to make suggestions, as I don't know what the release process for this project looks like.

lsm5:
> It's hard for me to make suggestions, as I don't know what the release process for this project looks like.

@umohnani8 we'll need to create a RELEASE_PROCESS.md as in podman and I guess we should be ok to include bump2version as part of the process.

lsm5:
for whatever reason, mergebot isn't working. Merging manually..

Issue #239, opened 2023-02-18
No support for fetching stats of all containers


RazCrimson:
**Current Scenario**
`podman-py` currently does not directly provide any means to fetch the stats of all containers. 

But this is allowed in as per the [Podman REST API](https://docs.podman.io/en/latest/_static/api.html?version=v4.4#tag/containers/operation/ContainersStatsAllLibpod).
Note that the `containers` parameter can take in multiple IDs or no ID to return data for all containers.
The current method that uses this API `Container.stats` does not allow such actions.

**Proposed Solution**
I propose to add an additional `stats` method under `ContainerManager` that would optionally take in cointainer_ids and stream the results.

Shall I go ahead and make a PR for this with the required changes or are there any other ways to implement this that would be considered better?

TomSweeneyRedHat:
@RazCrimson was this solved with your PR?  (#237), or are there more bits to chase here?

TomSweeneyRedHat:
Thanks for the clarification @RazCrimson .  FWIW, PR's are always gratefully accepted!

umohnani8:
@RazCrimson thank yo for your contributions! Did your PRs address this issue or is there still some work left?

inknos:
closing, seems fixed to me. up to investigate on a new issue if there are more updates or missing features

PR #237, opened 2023-02-14
Fix Container.stats behaviour

RazCrimson:
Fixes #218 

The current behavior of the `Container.stats` method does not take into account the streaming mode, leading to it freezing when we use it in stream mode. It also has other issues regarding decoding even when its being used in a non-stream mode. This PR fixes those issues.

I've also updated some type-hints, which I think are wrongly typed.
Feel free to request changes if you think they are inappropriate.

PR #250, opened 2023-03-12
Fix stream issues in container.top

### LINKED ISSUE #249: Streaming not supported in Container.top
Current implementation of `top` method on container doesn't support streaming. 
If we try to stream, it just hangs up infinitely.

```
Python 3.10.9 (main, Dec 19 2022, 17:35:49) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import podman
>>> client = podman.PodmanClient(base_uri="http+unix:///run/user/1000/podman/podman.sock")
>>> client.containers.list()[0].top(stream=False)
{'Processes': [['mongodb', '1', '0', '0.446', '41m4.487375206s', '?', '11s', 'mongod --bind_ip_all']], 'Titles': ['USER', 'PID', 'PPID', '%CPU', 'ELAPSED', 'TTY', 'TIME', 'COMMAND']}
>>> client.containers.list()[0].top(stream=True)

```
This cause for this issue is similar to #218 

Other problems in `top` is that, it decodes non-stream response to json, but sends the streamed responses as plain binary strings.

RazCrimson:
Changes:
- Add a common `stream_helper` to stream and decode json responses
- Switches existing methods to use the newer helper.
- Fixes Container.top issues mentioned in #249 

The `stream_helper` would be useful when other APIs require streaming and JSON decoding like for the API mentioned  in #239

Please do let me know in case something needs to be changed or amended.

Closes #249

### DIFF
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -6,7 +6,6 @@ from contextlib import suppress
 from typing import Any, Dict, Iterable, Iterator, List, Mapping, Optional, Tuple, Union
 
 import requests
-from requests import Response
 
 from podman import api
 from podman.domain.images import Image
@@ -457,22 +456,19 @@ class Container(PodmanResource):
             NotFound: when the container no longer exists
             APIError: when the service reports an error
         """
+        stream = kwargs.get("stream", False)
+
         params = {
+            "stream": stream,
             "ps_args": kwargs.get("ps_args"),
-            "stream": kwargs.get("stream", False),
         }
-        response = self.client.get(f"/containers/{self.id}/top", params=params)
+        response = self.client.get(f"/containers/{self.id}/top", params=params, stream=stream)
         response.raise_for_status()
 
-        if params["stream"]:
-            self._top_helper(response)
+        if stream:
+            return api.stream_helper(response, decode_to_json=True)
 
         return response.json()
-
-    @staticmethod
-    def _top_helper(response: Response) -> Iterator[Dict[str, Any]]:
-        for line in response.iter_lines():
-            yield line
 
     def unpause(self) -> None:
         """Unpause processes in container."""
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -389,7 +389,9 @@ class Container(PodmanResource):
         )
         response.raise_for_status()
 
-    def stats(self, **kwargs) -> Iterator[Union[bytes, Dict[str, Any]]]:
+    def stats(
+        self, **kwargs
+    ) -> Union[bytes, Dict[str, Any], Iterator[bytes], Iterator[Dict[str, Any]]]:
         """Return statistics for container.
 
         Keyword Args:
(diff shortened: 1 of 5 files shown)


umohnani8:
Changes LGTM
@rhatdan @lsm5 @jwhonce PTAL

podman/domain/containers.py
```diff
@@ -413,20 +414,9 @@ def stats(self, **kwargs) -> Iterator[Union[bytes, Dict[str, Any]]]:
...
-    def _stats_helper(
-        decode: bool, body: Iterator[bytes]
-    ) -> Iterator[Union[bytes, Dict[str, Any]]]:
-        """Helper needed to allow stats() to return either a generator or a bytes."""
-        for entry in body:
-            if decode:
-                yield json.loads(entry)
-            else:
-                yield entry
+        return json.loads(response.content) if decode else response.content
```
jwhonce, line 419: This changes the behaviour of stats() to read all the contents before returning the JSON payload. Given there is a streaming flag I don't see this change as wrong, just potentially breaking for developers who may be using the previous implementation.

jwhonce (approved):

podman/domain/containers.py
RazCrimson: The change in loading the response was made in 3ad81a73
It has already been merged in #237 and released. 😅 

Also the existing code had issues for non-stream mode, where it was encoding each in the response with an `\n` appended. I ended up replacing it with a simple `json.loads`. 

Snippet for Ref:
```py
        with io.StringIO() as buffer:
            for entry in response.text:
                buffer.write(json.dumps(entry) + "\n")
            return buffer.getvalue()
```

Please feel free to suggest any better alternatives. 
I'll make the required changes.

lsm5:
LGTM if @rhatdan and @jwhonce are happy.

@RazCrimson might wanna squash commits?

lsm5:
> @lsm5 Is this current level of squashing fine or would you like me to squash everything into one commit?

i was just checking if you needed / wanted to squash at all :) . Either way is fine by me.

Also, there's a ModuleNotFoundError now :(

RazCrimson:
Weird, there are no actual changes in the files before and after squashing.

Triggering another build did not fix it too.

Any ideas? 🤔

lsm5:
@cevich any idea whatsup with the fedora-37 failure?

cevich:
> @cevich any idea whatsup with the fedora-37 failure?

Unfortunately no, I encountered the same/similar errors while trying to update the CI VM images over in https://github.com/containers/podman-py/pull/251

I think we need a python-guru to take a look.

mwhahaha:
Probably need to add extras into the requirements. It was removed from `fixtures` via https://github.com/testing-cabal/fixtures/commit/f8df2192bf1190ff21a3067bb9e132b6db85191e however we're using `fixtures~=3.0.0`. Could try bumping fixtures to `4.0.0` or including extras in the requirements.

mwhahaha:
Fixing CI via #252

lsm5:
@mwhahaha thanks! @RazCrimson rebasing should make CI green.

RazCrimson:
@lsm5 Done!

rhatdan:
LGTM

PR #252, opened 2023-03-15
Bump fixtures requirement

mwhahaha:
CI is breaking due to a missing requirement on extras. This was dropped in newer versions.


cevich:
Thanks Alex.

PR #285, opened 2023-06-13
Refactor Makefile to support tox / MacOS Ventura

jwhonce:
* Assumes homebrew is installed and configured
* Uses which in place of command -v which is not implimented on MacOS
* Assumes if dnf is not found use brew. aptget support could be added in future

Note: Testing machine had run xcode-select --install


umohnani8:
LGTM

PR #288, opened 2023-06-15
Packit: remove .packit.sh

lsm5:
Packit's default actions should suffice for handling copr builds, making the `.packit.sh` script unnecessary.

Epoch for copr rpms has been bumped to 102 to preserve any issues with upgrade path for users.


lsm5:
@umohnani8 @rhatdan PTAL

cc: @jwhonce @cevich @edsantiago

PR #289, opened 2023-06-15
Keep main branch version ahead of release branches

lsm5:
Use `-dev` to denote development branch.

 `LIBPOD_URL` in `poodman/tests/__init__.py` doesn't work with `-dev`, so
we only use `vX.Y.Z`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ DESTDIR ?=
 EPOCH_TEST_COMMIT ?= $(shell git merge-base $${DEST_BRANCH:-main} HEAD)
 HEAD ?= HEAD
 
-export PODMAN_VERSION ?= "4.5.0"
+export PODMAN_VERSION ?= "4.6.0-dev"
 
 .PHONY: podman
 podman:
--- a/podman/tests/__init__.py
+++ b/podman/tests/__init__.py
@@ -3,5 +3,5 @@
 # Do not auto-update these from version.py,
 #   as test code should be changed to reflect changes in Podman API versions
 BASE_SOCK = "unix:///run/api.sock"
-LIBPOD_URL = "http://%2Frun%2Fapi.sock/v4.5.0/libpod"
+LIBPOD_URL = "http://%2Frun%2Fapi.sock/v4.6.0/libpod"
 COMPATIBLE_URL = "http://%2Frun%2Fapi.sock/v1.40"
--- a/podman/version.py
+++ b/podman/version.py
@@ -1,4 +1,4 @@
 """Version of PodmanPy."""
 
-__version__ = "4.5.0"
+__version__ = "4.6.0-dev"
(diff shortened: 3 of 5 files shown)


lsm5:
I'm cleaning up packit stuff in https://github.com/containers/podman-py/pull/288 and noticed main version is older than what's released.

lsm5:
> I'm cleaning up packit stuff in #288 and noticed main version is older than what's released.

hmm, maybe it's better to get #288 in first.

rhatdan:
@lsm5 does this need a rebase?

lsm5:
> @lsm5 does this need a rebase?

yes it does. I'll do it now.

rhatdan:
@lsm5 Still a draft?

lsm5:
> @lsm5 Still a draft?

I'll remove draft status once builds pass.

lsm5:
@umohnani8 @rhatdan Ready to go

rhatdan:
LGTM

umohnani8:
I think let's add `-dev` to the version here so that when we actually want to cut 4.6 we can remove the `-dev` and cut from that commit. Keeping it at 4.6 right will be a bit confusing in the commit history when we finally cut and release 4.6.
Once we cut 4.6, we will create a `release-4.6` branch for any backport and future 4.6.y releases and the main branch can then be bumped to `4.7-dev`. 
@lsm5 @rhatdan wdyt?

lsm5:
> I think let's add `-dev` to the version here so that when we actually want to cut 4.6 we can remove the `-dev` and cut from that commit. Keeping it at 4.6 right will be a bit confusing in the commit history when we finally cut and release 4.6. Once we cut 4.6, we will create a `release-4.6` branch for any backport and future 4.6.y releases and the main branch can then be bumped to `4.7-dev`. @lsm5 @rhatdan wdyt?

works for me. I can add that to the PR and we'll see what blows up :)

lsm5:
argh, got distracted. Let me fix this one.

lsm5:
@umohnani8 @rhatdan ping any idea whatsup with the test:
```
E       requests_mock.exceptions.NoMockAddress: No mock address: GET http://%2frun%2fapi.sock/v4.6.0/libpod/containers/json
```

Is it that `-` character doesn't work here, so it's not taking `-dev` ?

lsm5:
@umohnani8 @rhatdan Looks like `LIBPOD_URL` in `podman/tests/__init__.py` needs to be `v4.6.0` and nothing else. v4.6.0-dev doesn't work.  If that's ok, I'll squash commits, then we can merge this.

lsm5:
Also, I think sometime ago there was some blocker that prevented us from specifying the version only once. Has that issue been revisited?

lsm5:
Squashed commits. Also updated commit message. PTAL

PR #290, opened 2023-06-16
Fix catching errors in the response body of containers/prune

francisbergin:
Hi!

When `containers.prune()` is called, it currently fails silently since it looks for errors under the wrong key. The API returns it under the `Err` key and not the `error` key.

Tested using podman `4.5.1`.

Thank you!


rhatdan:
LGTM
@jwhonce @umohnani8 @mwhahaha PTAL

PR #294, opened 2023-06-25
Add type validation

francisbergin:


podman/api/client.py
```diff
@@ -178,7 +178,7 @@ def _normalize_url(base_url: str) -> urllib.parse.ParseResult:
     def delete(
         self,
         path: Union[str, bytes],
-        params: Union[None, bytes, Mapping[str, str]] = None,
+        params: Union[None, bytes, Mapping[str, Any]] = None,
```
jwhonce, line 181: This change seems spurious as none of the other like methods were flagged, ie get/put/head/... These methods all call _request() which has `params: Union[None, bytes, Mapping[str, str]] = None,`

jwhonce (commented):
@francisbergin I like what I see on the whole. I am interested in what you find concerning the convenience  methods. Changing delete() seems like an error in the type checker.

podman/api/client.py
francisbergin: This change was to fix this error:
```
podman/domain/networks_manager.py:202: error: Dict entry 0 has incompatible type "str": "bool | None"; expected "str": "str"  [dict-item]
```
This is the call to delete in that file:
```
response = self.client.delete(f"/networks/{name}", params={"force": force})
```
Where force is typed `Optional[bool]`.

francisbergin: So the type checker didn't flag the delete method directly; it flagged the call to it which did not follow the type hints.

jwhonce: Ah, should the caller be fixed instead?

francisbergin: I think it is reasonable to call `client.delete("...", params={"force": True}` so supporting a bool value should be good.

However, I'm not sure the use of `Mapping[str, Any]` is the best approach. Since the value for params is sent directly to `requests.Session.request`, the most appropriate supported type would be that one. [Typeshed defines it as](https://github.com/python/typeshed/blob/7646d38302ab7c2d56c2620ac43928331cc65c98/stubs/requests/requests/sessions.pyi#L95-L103):
```python
_ParamsMappingKeyType: TypeAlias = str | bytes | int | float
_ParamsMappingValueType: TypeAlias = str | bytes | int | float | Iterable[str | bytes | int | float] | None
_Params: TypeAlias = (
    SupportsItems[_ParamsMappingKeyType, _ParamsMappingValueType]
    | tuple[_ParamsMappingKeyType, _ParamsMappingValueType]
    | Iterable[tuple[_ParamsMappingKeyType, _ParamsMappingValueType]]
    | str
    | bytes
)
```

But this seems like a lot. Also, if I understand correctly, using [Any](https://docs.python.org/3/library/typing.html#the-any-type) disables type verification for the specific variable.

Maybe we can just add support for bool to allow for the specific usage and adapt later as required?

```python
params: Union[None, bytes, Mapping[str, Union[str, bool]]] = None,
```

jwhonce: I am comfortable with that change, as long as it is propagated to all the http verb helper functions, ie get/post/etc

francisbergin:
@jwhonce The gating tests are now passing with type hint validation! However, I think this PR still needs a bit of work:

- Careful review of a few code changes (I can create independent PRs for those)
- `#type: ignore[...]` instances that could be removed with a bit of extra work
- `Any` instances that could be replaced with more specific types
- Using convention of `list/dict/tuple` or `List/Dict/Tuple` across package

There are a lot of changes required to get mypy to pass but I believe it brings a lot of value to the project.

podman/api/client.py
```diff
@@ -378,14 +377,17 @@ def _request(
...
+        timeout_kw: dict[str, Any] = {}
         timeout = timeout or self.timeout
         if timeout_kw is not None:
             timeout_kw["timeout"] = timeout
 
         compatible = kwargs.get("compatible", False)
         path_prefix = self.compatible_prefix if compatible else self.path_prefix
 
+        if isinstance(path, bytes):
+            path = path.decode()
```
francisbergin, line 389: @jwhonce Since `path` can be of type `bytes`, converting it to a string is required to get `.lstrip(str)` to be valid. Without conversion, a `TypeError` would be thrown on a bytes path. It seems all calls to `APIClient` are done within the package and all use strings for the `path` parameter. Is so, specifying `path` as `str` only would remove the need for a conversion here.

podman/domain/config.py
```diff
@@ -44,12 +44,12 @@ def id(self):  # pylint: disable=invalid-name
         """str: Returns identifier for service connection."""
         return self.name
 
-    @cached_property
+    @cached_property  # type: ignore[operator]
```
francisbergin, line 47: For some reason, mypy thinks `cached_property` is a module and fails with a "module not callable" error.

podman/domain/images.py
```diff
@@ -1,17 +1,22 @@
...
 import logging
-from typing import Any, Dict, Iterator, List, Optional, Union
+from typing import Any, Dict, Iterator, List, Optional, TYPE_CHECKING, Union
 
 from podman import api
 from podman.domain.manager import PodmanResource
 from podman.errors import ImageNotFound
 
+if TYPE_CHECKING:
+    from podman.domain.images_manager import ImagesManager
```
francisbergin, line 10: To avoid cyclic import.

podman/domain/manager.py
```diff
@@ -1,12 +1,13 @@
...
 from collections import abc
-from typing import Any, List, Mapping, Optional, TypeVar, Union
+from typing import Any, Iterable, Mapping, Optional, TypeVar, Union
 
 from podman.api.client import APIClient
 
 # Methods use this Type when a subclass of PodmanResource is expected.
-PodmanResourceType: TypeVar = TypeVar("PodmanResourceType", bound="PodmanResource")
+# pylint: disable=invalid-name
+PodmanResourceType = TypeVar("PodmanResourceType", bound="PodmanResource")
```
francisbergin, line 10: I'm not 100% confident in the `PodmanResource/PodmanResourceType` changes in this module.

podman/domain/registry_data.py
```diff
@@ -32,14 +37,14 @@ def __init__(self, image_name: str, *args, **kwargs) -> None:
...
-    def pull(self, platform: Optional[str] = None) -> Image:
+    def pull(self, platform: Optional[str] = None) -> Union[Image, List[Image], Iterator[str]]:
         """Returns Image pulled by identifier.
 
         Args:
             platform: Platform for which to pull Image. Default: None (all platforms.)
         """
         repository = api.parse_repository(self.image_name)
-        return self.manager.pull(repository, tag=self.id, platform=platform)
+        return self.manager.pull(repository[0], tag=self.id, platform=platform)
```
francisbergin, line 47: `api.parse_repository` returns a tuple but `ImagesManager.pull` takes a string as first argument.

podman/domain/registry_data.py
```diff
@@ -63,7 +68,7 @@ def has_platform(self, platform: Union[str, Mapping[str, Any]]) -> bool:
 
         if isinstance(platform, dict):
             if not {"os", "architecture"} <= platform.keys():
-                version = self.client.version()
+                version = self.client.version()  # type: ignore[operator]
```
francisbergin, line 71: Here `self.client` is a `APIClient` where `.version` is a string. It would be in `PodmanClient` that `.version` would be callable. This specific line should throw an error as it is.

umohnani8:
@francisbergin any updates here?

francisbergin:
Hi! No, I didn't have time to continue working on this. I still think it would be good to have but the PR still needs some work and I don't have much time now. Maybe another time!

Issue #264, opened 2023-04-29
List tags parameter on images search method


madhuravius:
I believe `listTags` (`--list-tags` equivalent on cli / [api](https://github.com/containers/podman/blob/19152fa349199e88212aa9607c27139730d1710b/pkg/api/handlers/compat/images_search.go#L26)) is a flag that can be enabled on podman's api for the search behavior: https://docs.podman.io/en/latest/markdown/podman-search.1.html#list-tags

It would be nice to have this flag enabled so we can search for these tags in our results. I have a [simple proof here of function below](https://github.com/madhuravius/podman-py/pull/1). 

```py
>>> from podman import PodmanClient
>>> import pprint as pp
>>> uri = "unix:///run/user/1000/podman/podman.sock"
>>> client = PodmanClient(base_url=uri)
>>> pp.pprint(client.images.search(term="docker.io/nginx", list_tags=True)[0:5])
[{'Automated': '',
  'Description': '',
  'Index': 'docker.io',
  'Name': 'docker.io/library/nginx',
  'Official': '',
  'Stars': 0,
  'Tag': '1'},
 {'Automated': '',
  'Description': '',
  'Index': 'docker.io',
  'Name': 'docker.io/library/nginx',
  'Official': '',
  'Stars': 0,
  'Tag': '1-alpine'},
...
```

I did not see a comparable issue opened for this.

inknos:
Closing, looks like it's fixed on the linked pr. Don't hesitate to reopen if more attention is needed :)

PR #295, opened 2023-06-27
Includes the possibility to use the 'listTags' parameter in image search endpoint

apozsuse:
From Podman 4.0, the endpoint for image search ([doc](https://docs.podman.io/en/latest/_static/api.html?version=v4.0#tag/images/operation/ImageSearchLibpod) ) includes the possibility to include a `listTags` (bool) parameter that allows the user get all the available tags for a specific image in the registry.

Similar suggestion (and solution) was raised in [Issue 264](https://github.com/containers/podman-py/issues/264).

The parameter is only included in the API request if the user includes it in the kwargs parameters for the function, so this change should not break anything if the user is interacting with some podman instance < 4.0.

Unit and integration tests have been included in the PR.

Signed-off-by: apozsuse <andres.munoz@suse.com>

### DIFF
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -44,7 +44,6 @@ class ImagesManager(BuildMixin, Manager):
 
                 - dangling (bool)
                 - label (Union[str, List[str]]): format either "key" or "key=value"
-
         Raises:
             APIError: when service returns an error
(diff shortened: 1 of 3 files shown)


jwhonce (approved):
LGTM

PR #298, opened 2023-07-02
Fix passing filters to networks prune

francisbergin:


francisbergin:
The build seems to be failing because of the [bump to fixtures 4.1.0](https://github.com/containers/podman-py/commit/3e3246896e50d8fe7b7659d4f353f6bec718dfa9) which [dropped support for Python 3.6](https://github.com/testing-cabal/fixtures/blob/master/NEWS).

rhatdan:
Can you rebase off of current main branch, should fix the fixtures issue.

francisbergin:
Done!

PR #306, opened 2023-07-24
Packit: mention `downstream_package_name: python-podman` in config

lsm5:
The Fedora package name doesn't match upstream name. As a result, packit cannot automatically detect the right package repo for downstream tasks unless downstream_package_name is specified.

Success can only be verified in the next upstream release.

Ref: https://packit.dev/docs/configuration#downstream_package_name


lsm5:
@umohnani8 the lack of this change kept the Fedora tasks from being triggered. Safe to merge.

lsm5:
I'll bump v4.6.0 manually on Fedora.

PR #310, opened 2023-08-08
packit: Build PRs into default packit COPRs

martinpitt:
Building all PRs of all container projects into the same COPR does not properly isolate PRs from each other.

To avoid that, change the copr_build configuration to use the packit default COPRs, which are specific to the particular PR, and disappear after a few weeks. Depending projects should only run against what landed in podman-py/main i.e. the podman-next COPR.

----

This is exactly the same as https://github.com/containers/crun/pull/1260 ; please see the discussion there, this change needs to be applied to all container projects, and then all land together. Let's keep all the relevant discussions and tracking in  the crun PR please.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,27 +2,26 @@
 # See the documentation for more information:
 # https://packit.dev/docs/configuration/
 
-# Build targets can be found at:
-# https://copr.fedorainfracloud.org/coprs/rhcontainerbot/packit-builds/
-
 downstream_package_name: python-podman
 specfile_path: rpm/python-podman.spec
 upstream_tag_template: v{version}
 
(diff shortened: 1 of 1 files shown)


.packit.yaml
```diff
@@ -2,27 +2,24 @@
...
-    job: copr_build
+  - job: copr_build
     trigger: pull_request
-    owner: rhcontainerbot
-    project: packit-builds
-    enable_net: true
-    srpm_build_deps:
-      - make
+    targets:
+      - fedora-all
```
lsm5, line 16: Need centos-stream targets. Apart from that, LGTM.

lsm5 (approved):
/lgtm

PR #319, opened 2023-08-31
Fixes the secrets parameter, adds the secret_env parameter

andryyy:
1. Minor fix

Minor typo fix out of context: "example :" to "example:"

2. Secrets

The previous implementation of the secrets parameter allowed a string or Secret object to be defined.
The parameter passed to the API resulted in {"ID":str}.

While testing the secrets API I found that having only a single secret defined worked fine while more than one secret resulted in an exception:

```
500 Server Error: Internal Server Error (more than one result secret with prefix : secret is ambiguous)
```

The parameter ID was never used and a prefix of "" (empty) assumed.
Having a single entry in the Secrets Manager will always result in a single result, no matter the filter.

After losing myself in the API documentation, the correct data format I found is the following:

```
{
    "source": "my_secret", # name or id
    "target": "/my_secret", # optional
    "uid": 1000, # optional
    "gid": 1000, # optional
    "mode": 0o400, # optional
}
```

No changes for existing setups are necessary. Data (str, object, dict) is still expected within a list.

3. Secrets as environment variables

I added the secret_env parameter to set a secret's content as environment variable:

```
{
  "var_name": "secret_name",
}
```

---

Inline documentation was updated. :-)

### DIFF
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -212,7 +212,7 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
                     - As list of Secret objects the corresponding IDs are read from:
                         [Secret, Secret]
 
-                    - As list of dictionaries
+                    - As list of dictionaries:
                         [
                             {
                                 "source": "my_secret",  # A string representing the ID or name of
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
Thanks @andryyy

rhatdan:
@umohnani8 @jwhonce @baude PTAL

umohnani8:
Changes LGTM, but tests are unhappy

andryyy:
Okay 👍

Thank you for looking into it.

rhatdan:
https://github.com/containers/podman-py/pull/320 is merged

andryyy:
Done @rhatdan

PR #321, opened 2023-09-02
Fix image SCP when not providing quiet

hpvb:
Due to a logic error when trying to use the image scp functionality the destination wasn't passed unless "quiet" was also set.


rhatdan:
/approve
LGTM

PR #324, opened 2023-09-05
Change default value for tty in exec_run

### LINKED ISSUE #323: Incorrect documentation and incompatibility with docker-py in exec_run
exec_run uses `tty = True` by default, although the documentation comment below states that the default is `False`. Furthermore, `docker-py` uses false as the default value, so I think we should use `False` here as well, to minimize suprises when switching over.

maxi0604:
Closes #323

According to the documentation comment below, this value should be set to False. Furthermore, this is the behavior of docker-py, so it will be expected by many users. Therefore I think it should be `False` by default.

### DIFF
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -131,7 +131,7 @@ class Container(PodmanResource):
         stdout: bool = True,
         stderr: bool = True,
         stdin: bool = False,
-        tty: bool = True,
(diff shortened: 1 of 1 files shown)


rhatdan:
LGTM
@umohnani8 @jwhonce PTAL

jwhonce:
LGTM

PR #338, opened 2023-11-07
Fix up docs and format issues

umohnani8:
Clean up some format issues and also fix the command to build the docs so that is built accurately and in the right location for it to show up in readthedocs.

### DIFF
--- a/podman/api/client.py
+++ b/podman/api/client.py
@@ -7,7 +7,7 @@ from typing import Any, ClassVar, IO, Iterable, List, Mapping, Optional, Tuple, 
 import requests
 from requests.adapters import HTTPAdapter
 
-from podman import api
+from podman import api  # pylint: disable=cyclic-import
 from podman.api.ssh import SSHAdapter
 from podman.api.uds import UDSAdapter
 from podman.errors import APIError, NotFound
--- a/podman/api/client.py
+++ b/podman/api/client.py
@@ -1,4 +1,5 @@
 """APIClient for connecting to Podman service."""
+
 import json
 import urllib.parse
 from typing import Any, ClassVar, IO, Iterable, List, Mapping, Optional, Tuple, Type, Union
--- a/podman/api/ssh.py
+++ b/podman/api/ssh.py
@@ -203,7 +203,7 @@ class SSHConnection(urllib3.connection.HTTPConnection):
 class SSHConnectionPool(urllib3.HTTPConnectionPool):
     """Specialized HTTPConnectionPool for holding SSH connections."""
 
-    ConnectionCls = SSHConnection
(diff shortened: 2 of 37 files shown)


rhatdan:
LGTM

podman/api/ssh.py
```diff
@@ -202,6 +203,7 @@ def connect(self) -> None:
 class SSHConnectionPool(urllib3.HTTPConnectionPool):
     """Specialized HTTPConnectionPool for holding SSH connections."""
 
+    # pylint: disable=invalid-name
     ConnectionCls = SSHConnection
```
jwhonce, line 207:
```suggestion
    ConnectionCls = SSHConnection     # pylint: disable=invalid-name
```
Should disable the error for just this line, not the whole file.

podman/api/uds.py
```diff
@@ -89,6 +90,7 @@ def connect(self) -> None:
 class UDSConnectionPool(urllib3.HTTPConnectionPool):
     """Specialization of HTTPConnectionPool for holding UNIX domain sockets."""
 
+    # pylint: disable=invalid-name
     ConnectionCls = UDSConnection
```
jwhonce, line 94:
```suggestion
    ConnectionCls = UDSConnection     # pylint: disable=invalid-name

```

jwhonce (commented):
Nits.

LGTM

podman/api/client.py
```diff
@@ -7,7 +7,7 @@
 import requests
 from requests.adapters import HTTPAdapter
 
-from podman import api
+from podman import api  # pylint: disable=cyclic-import
```
umohnani8, line 10: @jwhonce this fixes the cyclic import error we were getting. I think this should be fine as we are only using `api` to get the version number so I don't think having a cyclic import here should break anything. But not entirely sure.

jwhonce: Ack, shouldn't be a problem...

podman/api/ssh.py
umohnani8: Fixed

podman/api/uds.py
umohnani8: Fixed

umohnani8:
@cevich this fixes the cyclic import error and should help get the tests in your PR passing once in.
Can I please get a lgtm here @lsm5 @rhatdan @cevich

cevich (approved):
LGTM - Goodness this is big.  Thanks Urvashi!

lsm5:
@umohnani8 LGTM. Do you wanna squash commits or did you intend for them to be 3 separate commits?

umohnani8:
I was planning on leaving them as 3 separate commits @lsm5

PR #341, opened 2023-11-14
Update pull progress bar colour

umohnani8:
The Downloading bar is now in the default terminal text color and is bolded for better visibility.
Bars will switch to green when completed.

### DIFF
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -6,7 +6,7 @@ import logging
 import urllib.parse
 from typing import Any, Dict, Generator, Iterator, List, Mapping, Optional, Union
 import requests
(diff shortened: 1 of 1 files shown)


umohnani8:
This is what the new output looks like
image

@baude @jwhonce @TomSweeneyRedHat PTAL

baude:
LGTM thanks for your work

edsantiago:
Switching to bold is a good move: I tested with a colorblindness simulator, and it stands out under all models. It also looks much better to my eye. Plus, @baude's blessing carries a lot of weight. LGTM2.

TomSweeneyRedHat:
Late, but looks better to me than the red/green prior.  Nice work @umohnani8

PR #353, opened 2023-12-11
Add rich dep to setup.cfg

umohnani8:
Fixes https://github.com/containers/podman-py/issues/352


umohnani8:
@rhatdan are you able to request the cherry-pick? I am shown as an owner in the containers org but seem to can't trigger the cherry-pick

Issue #360, opened 2023-12-21
Centos Stream 9 builds failing


umohnani8:
The centos stream 9 builds are continuously failing on PRs with error that it couldn't find the `rich` package. 

I checked if the version was the problem by removing any version restrictions, but it wasn't the case. Seems like the `rich` package probably doesn't exist on centos stream 9 or it has a very different name.

umohnani8:
@lsm5 can you please look into this when you are back.

yselkowitz:
This affects ELN/c10s as well, due to https://github.com/containers/podman-py/pull/340 having added a hard dependency on `rich`, which has a few new runtime dependencies and collectively have a huge build dependency tree.  At a minimum, please make the dependency (and hence the progress bar functionality) optional (e.g. with a try/except of import rich etc.), or find an alternative implementation which does not introduce so many new dependencies.

yselkowitz:
ping @lsm5

PR #361, opened 2024-01-03
Make progress_bar an extra feature

yselkowitz:
This allows the 'rich' dependency, which has additional dependencies and is not available in RHEL, to be optional.

Fixes: #360

### DIFF
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -6,7 +6,6 @@ import logging
 import urllib.parse
 from typing import Any, Dict, Generator, Iterator, List, Mapping, Optional, Union
 import requests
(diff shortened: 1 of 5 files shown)


lsm5:
LGTM.

@umohnani8 @TomSweeneyRedHat it's either this or we end up owning python-rich package in RHEL.

cc @jnovy @jwhonce

lsm5:
@umohnani8 I think we should get a new release out with this patch.

jnovy:
Agree for the new release ASAP @lsm5 

@yselkowitz this is brilliant, thank you!

TomSweeneyRedHat:
Neat change @yselkowitz !  Thanks for running this down and putting it together.

PR #364, opened 2024-01-15
Enable user defined netns

dcasier:
Allow user to specify netns => Example {"network_mode": "ns:/var/run/netns/<NETNS>"}


rhatdan:
Thanks @dcasier 
/approve
LGTM
@umohnani8 @jwhonce PTAL

PR #367, opened 2024-01-22
Bump version to v4.9

umohnani8:


umohnani8:
@rhatdan @lsm5 PTAL

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = podman
-version = 4.9.0-dev
+version = 4.9.0
 author = Brent Baude, Jhon Honce
 author_email = jhonce@redhat.com
```
lsm5, line 5: @umohnani8 maybe you should add / update with your name here. wdyt @jwhonce ?

lsm5 (approved):
/lgtm
/hold

setup.cfg
umohnani8: I will do it in the next update :)

Issue #330, opened 2023-09-27
Unify usage of `workdir`, `work_dir` and `working_dir`


aparcar:
Looking through the code I can find all three of them, would it be possible to unify them?

rhatdan:
Interested in opening a PR to do that?

PR #332, opened 2023-09-29
rename `working_dir` to `workdir` for consistency

aparcar:
Currently `podman-py` uses all three versions of `work_dir`, `working_dir` and `workdir` (not to mention `WorkingDir`).

This commit tries to unify the parameter usage by allowing `workdir` for container `create` or `run`. For backwards compatibility `working_dir` still works but a deprecation warning is added.

Since upstream Podman uses a variety of *workdir* versions the `podman-py` codebase can't be simplified further.

Fix: #330

PR #368, opened 2024-01-22
Add `workdir` as alias for `working_dir`

umohnani8:
Currently `podman-py` uses all three versions of `work_dir`, `working_dir` and `workdir` (not to mention `WorkingDir`).

This commit tries to unify the parameter usage by allowing `workdir` for container `create` or `run`. For backwards compatibility `working_dir` still works.

Since upstream Podman uses a variety of *workdir* versions the `podman-py` codebase can't be simplified further.

Fix: #330

### DIFF
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -331,6 +331,7 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
 
             volumes_from (List[str]): List of container names or IDs to get volumes from.
             working_dir (str): Path to the working directory.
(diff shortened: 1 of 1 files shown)


umohnani8:
Replacing https://github.com/containers/podman-py/pull/332

rhatdan:
What about the tests?

umohnani8:
@rhatdan no tests are needed, we are just adding an alias

lsm5 (approved):
/lgtm
/hold

PR #372, opened 2024-02-12
from_env: Use default base_url if no environment is given

kajinamit:
The APIClient class has its own mechanism to detect the default value for base_url if no specific value is provided. However this mechanism can't be used when creating an instance by from_env, because from_env requires (DOCKER|CONTAINER)_HOST environment.

This drops the validation from from_env, so that users can use from_env without explicitly setting the HOST environment.


rhatdan:
Should be a new release with podman 5.0 next week.

rhatdan:
@umohnani8 PTAL

PR #374, opened 2024-02-19
Use new json connections file

umohnani8:
Podman has updated where it will store its
system connection information to a new json
format file.
Add support to podman-py to read from both the
new json file and old toml file giving preference
to the new json file.


jwhonce (approved):
LGTM once CI is happy. 

One nit, a second test which verifies precedence on a duplicated connection name would be nice to have.

podman/domain/config.py
```diff
@@ -64,15 +69,22 @@ def __init__(self, path: Optional[str] = None):
...
+            with open(self.path, encoding='utf-8') as file:
+                self.attrs = json.load(file)
+
+        # Read the old toml file configuration
+        if old_toml_file.exists():
+            with old_toml_file.open(encoding='utf-8') as file:
                 buffer = file.read()
-            self.attrs = toml_loads(buffer)
+            loaded_toml = toml_loads(buffer)
+            self.attrs.update(loaded_toml)
```
jwhonce, line 87: I think you should load the toml file, then update the dict with results of the json load.

jwhonce (changes requested):

podman/domain/config.py
```diff
@@ -64,15 +69,22 @@ def __init__(self, path: Optional[str] = None):
 
         if path is None:
             home = Path(xdg.BaseDirectory.xdg_config_home)
-            self.path = home / "containers" / "containers.conf"
+            self.path = home / "containers" / "podman-connections.json"
+            old_toml_file = home / "containers" / "containers.conf"
```
Luap99, line 73: Does this work on windows? The windows containers.conf path differ but I am not sure if this repo cares about windows support at all?

podman/domain/config.py
umohnani8: https://github.com/containers/podman-py/pull/374/files#diff-891006306a18924b730354f5400759161db17c6e727057a5fa71ef90d525a3b9R130 is where we create the services we return to the user. The `__init__` function just reads the two files and loads the keys and values into the `self.attrs` dict.

podman/domain/config.py
```diff
@@ -62,17 +67,43 @@ class PodmanConfig:
...
 
+        self.is_default = False
         if path is None:
             home = Path(xdg.BaseDirectory.xdg_config_home)
-            self.path = home / "containers" / "containers.conf"
+            self.path = home / "containers" / "podman-connections.json"
+            old_toml_file = home / "containers" / "containers.conf"
+            self.is_default = True
+        # this elif is only for testing purposes
+        elif "@@is_test@@" in path :
```
umohnani8, line 77: So I tried using mock env to imitate the path we default to but it just wouldn't work. This is the best work around I could come up with. If anyone is able to figure out the mock env thing I am happy to remove this and update the test accordingly.

Looking at the code, seems like we grab the path to look at based on `XDG_CONFIG_HOME` and that is what I kept trying to override in the tests, but it would always end up using the path on my local machine instead.

umohnani8:
Tests are green!

@jwhonce PTAL

podman/domain/config.py
umohnani8: I am not sure about our windows story, @jwhonce do you have any idea?

jwhonce: It used to work but did not use the windows registry. I no longer have a windows license to spin on a VM for testing.

jwhonce (approved):
LGTM

podman/domain/config.py
Luap99: well works yeah sure it can read these paths but these are not the paths podman actually uses on windows

podman/domain/config.py
```diff
@@ -62,17 +67,45 @@ class PodmanConfig:
...
             home = Path(xdg.BaseDirectory.xdg_config_home)
-            self.path = home / "containers" / "containers.conf"
+            self.path = home / "containers" / "podman-connections.json"
+            old_toml_file = home / "containers" / "containers.conf"
+            self.is_default = True
+        # this elif is only for testing purposes
+        elif "@@is_test@@" in path:
+            test_path = path.replace("@@is_test@@", '')
+            self.path = Path(test_path) / "podman-connections.json"
+            old_toml_file = Path(test_path) / "containers.conf"
```
Luap99, line 80: this sounds like an ugly hack but I am not a python programmer so I don't really care if this is the right thing to solve it.

Luap99 (commented):
A couple of notes:
This is not honouring CONTAINERS_CONF or CONTAINERS_CONF or PODMAN_CONNECTIONS_CONF env overwrites.
It does not support reading containers.conf.d

I am not saying these have to be fixed here but I wanted to list it. I am fine to leave them unfixed until a user complains.

podman/domain/config.py
umohnani8: I agree, I really dislike it too but I tried everything with the mock Env options but it just wouldn't pick up the env I was setting for `XDG_CONFIG_HOME`.
I can look into it further and open a PR in the future if I can figure it out.

umohnani8:
> A couple of notes: This is not honouring CONTAINERS_CONF or CONTAINERS_CONF or PODMAN_CONNECTIONS_CONF env overwrites. It does not support reading containers.conf.d
> 
> I am not saying these have to be fixed here but I wanted to list it. I am fine to leave them unfixed until a user complains.

Agreed, I think it might be best to call `podman system connection ls`. Will have a discussion with @jwhonce on that and see what we can do with that. In the meantime let's get this in so we are somewhat up to date with the new connections file.

umohnani8:
@rhatdan @lsm5 can I please get a lgtm here

Luap99:
Let's see if I have some power here
/lgtm

Issue #293, opened 2023-06-18
Enable types in podman package


ghost:
Adding a `py.typed` file in the root of the `podman` package would allow type checking for it in projects that use the package.

For example, the following script:

```python3
import podman
client = podman.PodmanClient()
ok = client.containers.exists(1234)
print("Exists: " + ok)
```

Would not cause any mypy errors (issues would be caught at runtime):

```console
$ mypy --ignore-missing-imports script.py 
Success: no issues found in 1 source file
```

But with the `py.typed` file, we can catch the issues earlier:

```console
$ touch venv/lib/python3.11/site-packages/podman/py.typed
$ mypy --ignore-missing-imports script.py 
script.py:3: error: Argument 1 to "exists" of "ContainersManager" has incompatible type "int"; expected "str"  [arg-type]
script.py:4: error: Unsupported operand types for + ("str" and "bool")  [operator]
Found 2 errors in 1 file (checked 1 source file)
```

Maybe package type issues should be fixed in #292 before enabling type information in the package?

rhatdan:
SGTM
@jwhonce @umohnani8 WDYT?

francisbergin:
Hi!

The chosen tool would be a compliment to Pylint which would focus on type checking. The main tools available that I am aware of are [Mypy], [Pyright], [Pyre] and [Pytype]. I believe the basic functionality should be quite similar between these 4 projects.

I started #294 as a POC of changes required to get the type checker to pass. The [first run](https://cirrus-ci.com/task/6041181319069696) had 331 type errors and the [current run](https://cirrus-ci.com/task/5748288507346944) with fixes has 179 errors. In this PR, I adapted the type hints to the code but perhaps in some cases the code needs to be adapted to the type hints.

Please let me know what you think of the changes so far and if it is worth continuing the effort.

Note that for this specific issue, the change would just be to add a py.typed file to the published package which would allow users to type check their usage of the podman package.

Thank you!

[Mypy]: https://github.com/python/mypy
[Pyright]: https://github.com/microsoft/pyright
[Pyre]: https://github.com/facebook/pyre-check
[Pytype]: https://github.com/google/pytype

jnsnow:
[...]

PR #381, opened 2024-03-13
Include py.typed marker file.

jonded94:
I'm using this package in codebases checked with common static type checking tools, such as `mypy`. Unfortunately, it will error:
```
error: Skipping analyzing "podman": module is installed, but missing library stubs or py.typed marker  [import]
```

This was already apparent in https://github.com/containers/podman-py/issues/293. Including this file would at least enable checking against the many type hints that are already a part of this library and would help massively.


jwhonce:
@jnsnow PTAL, TIA

umohnani8:
Thank you @jonded94! Can you please rebase with upstream and sign your commit and repush, I think that should help fix the DCO and gating failure.

Changes LGTM

jonded94:
Did a rebase and a force repush, with a signed commit (which also shows to be verified in Github), but unfortunately that DCO job still fails?

umohnani8:
I believe DCO is failing because your commit doesn't have the `Signed off by...` line it. Can you please ensure `git commit -s --amend` adds that for you and then repush.

jonded94:
Tests seem to run through now, thanks for your help!

rhatdan:
/approve
/lgtm
Thanks @jonded94

PR #388, opened 2024-05-08
[skip-ci] Packit: enable c10s downstream sync

lsm5:
This commit will enable downstream syncing to CentOS Stream 10. The centos maintainer will need to manually run `packit propose-downstream` and `centpkg build` until better centos integration is in place.

This commit also builds both rhel and centos stream copr rpms so we can check for things like differences in python toolchain.

EL8 jobs have also been deleted. CentOS Stream 8 will go EOL soon. We won't be shipping anything from main into EL8.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -3,32 +3,67 @@
 # https://packit.dev/docs/configuration/
 
 downstream_package_name: python-podman
-specfile_path: rpm/python-podman.spec
+update_release: false
 upstream_tag_template: v{version}
+
+packages:
+  python-podman-fedora:
+    pkg_tool: fedpkg
+    specfile_path: rpm/python-podman.spec
+  python-podman-centos:
+    pkg_tool: centpkg
+    specfile_path: rpm/python-podman.spec
+  python-podman-rhel:
+    specfile_path: rpm/python-podman.spec
 
 srpm_build_deps:
   - make
 
 jobs:
+  # Copr builds for Fedora
   - job: copr_build
     trigger: pull_request
+    packages: [python-podman-fedora]
     targets:
       - fedora-all
-      - centos-stream-8
+
+  # Copr builds for CentOS Stream
+  - job: copr_build
+    trigger: pull_request
+    packages: [python-podman-centos]
+    targets:
+      - centos-stream-10
       - centos-stream-9
+
+  # Copr builds for RHEL
+  - job: copr_build
+    trigger: pull_request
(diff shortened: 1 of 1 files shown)


lsm5:
@inknos PTAL

cc @TomSweeneyRedHat @jwhonce @umohnani8

jwhonce:
LGTM, I don't see any problems but also I do not know packit

.packit.yaml
```diff
@@ -3,33 +3,67 @@
...
     trigger: commit
     branch: main
     owner: rhcontainerbot
     project: podman-next
 
+  # Downstream sync for Fedora
   - job: propose_downstream
     trigger: release
+    packages: [python-podman-fedora]
     update_release: false
```
inknos, line 55: What's the reason for not setting this parameter globally like [here](https://packit.dev/docs/configuration/upstream/propose_downstream#syncing-the-release-to-centos-stream)?

.packit.yaml
```diff
@@ -3,33 +3,67 @@
...
     dist_git_branches:
       - fedora-all
 
+  # Downstream sync for CentOS Stream
+  - job: propose_downstream
+    trigger: release
+    packages: [python-podman-centos]
+    update_release: false
+    dist_git_branches:
+      - c10s
```
inknos, line 65: This might be a trivial question, that it's simply a missing context from my side. Why not c9s too?

.packit.yaml
lsm5: done. Thanks

.packit.yaml
lsm5: Yes, I'd like to have c9s too, but the trouble in c9s is that the buildroot doesn't have rpmautospec yet so we'll need additional gymnastics to handle changelog which I'd rather avoid. I've filed a jira to get rpmautospec included and will enable c9s soon as it's resolved.

inknos: I see, thanks. Would it make sense to include the JIRA issue in a TODO comment in the code so we can keep track of it in the packit file?

Other than that the PR looks good.

TomSweeneyRedHat:
LGTM

.packit.yaml
lsm5: done. PTAL

inknos: thanks, LGTM

lsm5:
@inknos PTAL. Previous comment addressed.

PR #393, opened 2024-05-23
[skip-ci] Packit: use default `update_release` behavior

lsm5:
The release needs to be updated if you're building for a persistent copr otherwise you'll end up with multiple builds with no change in the release tag.
See recent builds at: https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/package/python-podman/ Ref Build IDs: 7478065, 7455330, 7446571

Packit by default updates the Release in spec file for copr builds which is what we want.

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -3,7 +3,6 @@
 # https://packit.dev/docs/configuration/
 
 downstream_package_name: python-podman
-update_release: false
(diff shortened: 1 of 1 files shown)


lsm5:
@inknos PTAL 

cc @jwhonce @rhatdan @umohnani8

PR #395, opened 2024-05-29
Fix README TypeError when one container is running

### LINKED ISSUE #379: Example from Readme fails with TypeError
For me running the example from the Readme file results in:

```
Traceback (most recent call last):
[...]
    first_name = container['Names'][0]
TypeError: 'Container' object is not subscriptable
```

It seems that `Container` does not implement `__getitem__(self, key)`. However, I think probably `container.name` was meant to be used.

I think it would also be good to clarify why

```
container = client.containers.get(first_name)
```

is used in the example, as it seems to return the same container as already retrieved from the list. Perhaps, the line could be omitted or a clarifying comment or an `assert` showing that both objects are equal could be added.

Edit: for the problem to show a container must be running, since otherwise the loop will not be entered.

Python: 3.10.12
podman==4.9.0 (podman-py)
OS: Ubuntu 22.04

inknos:
closes #379

PTAL @jwhonce

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,12 @@ with PodmanClient(base_url=uri) as client:
 
     # find all containers
     for container in client.containers.list():
-        first_name = container['Names'][0]
-        container = client.containers.get(first_name)
+        # After a list call you would probably want to reload the container
+        # to get the information about the variables such as status.
+        # Note that list() ignores the sparse option and assumes True by default.
+        container.reload()
         print(container, container.id, "\n")
+        print(container, container.status, "\n")
 
         # available fields
         print(sorted(container.attrs.keys()))


README.md
```diff
@@ -35,9 +35,11 @@ with PodmanClient(base_url=uri) as client:
 
     # find all containers
     for container in client.containers.list():
-        first_name = container['Names'][0]
-        container = client.containers.get(first_name)
+        # it is important to get the container object again with the get()
+        # method to get all the fields like container.status populated
+        container = client.containers.get(container.name)
```
jwhonce, line 40:
```suggestion
        container.reload()
```
would also populate all the fields.

PS. The comment could also include a notice that sparse=True is our default. Future work could allow sparse=False on the list() call.

inknos: thanks @jwhonce, good idea. I committed new changes

jwhonce:
LGTM

PR #400, opened 2024-06-19
Add ssh env variables to tox testenv

inknos:
For users who have a ssh key protected by passphrase tests can hang and fail on 'Enter passphrase for key' if the correct ssh agent env variable is not set in tox.


jwhonce:
@inknos It looks like your change impacted how ssh connectiions are being parsed. Or did the podman API server change?

inknos:
no, there must be something wrong. I'll close the PR for now and try to look more in my local env. This looks like being a fix for me only :/

PR #405, opened 2024-06-30
Update index.rst Client -> PodmanClient

jwoehr:
I believe the example to be incorrect as it stands. The class invoked should be `PodmanClient`


inknos:
thanks

/lgtm

Issue #378, opened 2024-03-04
Running a locally non-existent image fails with AttributeError


rsommer:
The code for running a new container with an image that is not present locally tries to access the `ImagesManager` via `self.client.images` as seen here: 
https://github.com/containers/podman-py/blob/53b238b75f7419f55560d6b5ef7d24729cd186bd/podman/domain/containers_run.py#L63
As `self.client` is an `APIClient` instance and not an instance of `PodmanClient`, an `AttributeError` occurs:

```python
File /tmp/venv/lib/python3.11/site-packages/podman/domain/containers_run.py:63, in RunMixin.run(self, image, command, stdout, stderr, remove, **kwargs)
     61     container = self.create(image=image, command=command, **kwargs)
     62 except ImageNotFound:
---> 63     self.client.images.pull(image, platform=kwargs.get("platform"))
     64     container = self.create(image=image, command=command, **kwargs)
     66 container.start()

AttributeError: 'APIClient' object has no attribute 'images'
```
This is with `podman-py==4.9.0` running against podman 4.9.2. podman versions prior to 4.8.0 contain another error preventing running a locally non-existing image (See: https://github.com/containers/podman-py/issues/231).

This can be reproduced with the following snippet (execute `podman image rm docker.io/library/busybox:latest` if you have already pulled `busybox` prior to testing):
```sh
#!/bin/bash
python3 -m venv venv
venv/bin/pip install podman==4.9.0
venv/bin/python -c 'import podman; client = podman.PodmanClient(base_url="unix:///run/podman/podman.sock"); client.containers.run("docker.io/library/busybox:latest")'
```

PR #408, opened 2024-07-17
Fix the locally non-existent image fails with AttributeError

milanbalazs:
The code for running a new container with an image that is not present locally tries to access the `ImagesManager` via `self.client.images`.
As `self.client` is an `APIClient` instance and not an instance of `PodmanClient`, an `AttributeError` occurs:

```python
AttributeError: 'APIClient' object has no attribute 'images'
```

With this fix the `PodmanClient` object is used instead of `APIClinet` correctly.

**A very minimal test code:**

```python
import podman

with podman.PodmanClient() as client:
    client.containers.run("docker.io/library/busybox:latest")
```

The above code failed with `AttributeError` before my fix. Currently it pulls the Image if it's not available in local and run the container.

**Note:**
-  I guess the other `@cached_property` methods also can be extended with my solution if it's needed. I have extended only the `ContainersManager` with `podman_client` argument.

It's fix for:
https://github.com/containers/podman-py/issues/378

### DIFF
--- a/podman/client.py
+++ b/podman/client.py
@@ -138,7 +138,7 @@ class PodmanClient(AbstractContextManager):
     @cached_property
     def containers(self) -> ContainersManager:
         """Returns Manager for operations on containers stored by a Podman service."""
-        return ContainersManager(client=self.api)
+        return ContainersManager(client=self.api, podman_client=self)
 
     @cached_property
     def images(self) -> ImagesManager:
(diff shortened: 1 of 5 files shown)


rhatdan:
LGTM
@jwhonce @umohnani8 PTAL

inknos (approved):

Issue #322, opened 2023-09-05
demux doesn't work in exec_run => no way to separate stdout and stderr


maxi0604:
Hello,

It seems like `demux` in `Container.exec_run` doesn't work, and that there is currently no way to separate stdout and stderr when using `exec` through this library. Quickly looking at the code, it seems to me like the `demux` parameter is only referenced in the doc comment, but not used in the code: https://github.com/containers/podman-py/blob/004ebda439975fdf81d87e84ae1387cf1aa0d3fe/podman/domain/containers.py#L142C1-L142C14

Thanks for considering this issue

inknos:
ack. not just that, `stream` parameter is also unused there. I am taking a look

omkarkhatavkar:
@inknos Do you have any further update on fixing this issue ?

JacobCallahan:
@maxi0604 If you want to take the implementation I did for demuxing, you can see it here: https://github.com/SatelliteQE/broker/blob/master/broker/binds/containers.py#L8-L48

inknos:
> Do you have any further update on fixing this issue ?

I am sorry I got distracted by other duties and forgot after PTO. will move it on the higher priority now.

PR #410, opened 2024-07-26
Enable demux option in `exec_run`

inknos:
`exec_run` now returns a tuple of bytes if demux is True the first element being the stdout and the second the stderr of the exec_run call.

Implementation is courtesy of:
https://github.com/SatelliteQE/broker/blob/60a52941f2eb297ccbdf7a0fa0b932eb23ad926b/broker/binds/containers.py#L8-L48

Resolves: https://github.com/containers/podman-py/issues/322

Thanks @JacobCallahan

### DIFF
--- a/podman/api/output_utils.py
+++ b/podman/api/output_utils.py
@@ -1,0 +1,49 @@
+"""Utility functions for dealing with stdout and stderr."""
+
+HEADER_SIZE = 8
+STDOUT = 1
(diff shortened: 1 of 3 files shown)


jwhonce:
LGTM

rhatdan:
Thanks @inknos 
/lgtm
/approve

Issue #275, opened 2023-05-31
client.images.prune() raises


danmilon:
Hey! Thanks for the great software!

When doing an images `prune()`, it gives me:

```
python -c 'import podman; client = podman.PodmanClient(base_url="unix:///run/podman/podman.sock"); client.images.prune();'

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/<redacted>/lib/python3.7/site-packages/podman/domain/images_manager.py", line 151, in prune
    for element in response.json():
TypeError: 'NoneType' object is not iterable
```

Upon investigation, I see that `response.text` is `null\n` and status code is 200.

I'm using podman-py version 4.5.0 and `podman version` is:

```
$ podman version
Version:      3.4.2
API Version:  3.4.2
Go Version:   go1.15.9
Built:        Thu Jan  1 00:00:00 1970
OS/Arch:      linux/amd64
```

When doing `podman image prune` it works fine.

inknos:
Resolved by https://github.com/containers/podman-py/pull/412

PR #412, opened 2024-08-01
Fix the TypeError exception in the images.prune method

milanbalazs:
If the prune doesn't remove images, the API returns `null` (with `200` status code) and it's interpreted as `None` (`NoneType`) so the for loop throws the following exception:

```
"TypeError: 'NoneType' object is not iterable".
```

My fix handles the above described case and the `client.images.prune()` returns a valid Dict with zero values, which is correct because the Podman didn't remove anything:

```python
{
    "ImagesDeleted": [],
    "SpaceReclaimed": 0,
}
```

**It's fix for:**
 - https://github.com/containers/podman-py/issues/275

### DIFF
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -158,17 +158,22 @@ class ImagesManager(BuildMixin, Manager):
         deleted: List[Dict[str, str]] = []
         error: List[str] = []
         reclaimed: int = 0
-        for element in response.json():
-            if "Err" in element and element["Err"] is not None:
-                error.append(element["Err"])
-            else:
-                reclaimed += element["Size"]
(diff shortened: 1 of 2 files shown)


inknos:
Thanks. I got some errors after pulling and local testing. Will get back as soon as I verify what's happening :)

milanbalazs:
> Thanks. I got some errors after pulling and local testing. Will get back as soon as I verify what's happening :)

Hi @inknos,

Thanks for your feedback! I have just fixed the error (Squashed the commits)!

rhatdan:
LGTM
@inknos @jwhonce @umohnani8 PTAL

inknos (approved):

Issue #312, opened 2023-08-09
ImagesManager.prune() - no support for query-params/kwargs?


Dsafe1:
Hello.

I am building a small application to manage my images and thus tried to prune ALL images not in use by using [ImagesManager.prune()](https://github.com/containers/podman-py/blob/1b6b6ea4e08df8f3a701519f7e97c68c26001709/podman/domain/images_manager.py#L126C8-L126C8).
[Podman specifies this as possible via the "all" query parameter](https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImageLoadLibpod). 

At first I struggled because I didn't realize that the "filters" argument is something else entirely, and that there is (afaik) no way to pass the "all" parameter to perform a full image prune.
Is my assessment correct or am I missing something?
If yes, then why is it this way? Wouldn't it be much more flexible to enable arbitrary query params for all routes so the user can talk to the entire Podman API using this library?

inknos:
Hey,

> If yes, then why is it this way? Wouldn't it be much more flexible to enable arbitrary query params for all routes so the user can talk to the entire Podman API using this library?

Implementing `external`, `all`, and `filters` params makes sense to me. @jwhonce wdyt?

jwhonce:
@inknos Adding `external` and `all` as kwargs would require extra validation if the client has been instantiated in `compatible_version` mode. That is, the client is running without the `libpod` extensions. Otherwise, seems like a good idea to me.

`filters` has already been implemented for the `compatible` entries. Adding the `libpod` `label` filter would complete the extensions.

inknos:
> would require extra validation if the client has been instantiated in compatible_version mode

I would pursue this path.

> Adding the libpod label filter would complete the extensions.

This also makes sense to me

PR #413, opened 2024-08-01
Add all, external, and label to Image.prune()

inknos:
Param all is now supported by prune.
Param external is now supported by prune.
Filter by label, which was already supported, is now documented and tested.

Resolves: https://github.com/containers/podman-py/issues/312

### DIFF
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -135,24 +135,39 @@ class ImagesManager(BuildMixin, Manager):
             yield self.get(item)
 
     def prune(
-        self, filters: Optional[Mapping[str, Any]] = None
+        self,
+        all: Optional[bool] = False,  # pylint: disable=redefined-builtin
+        external: Optional[bool] = False,
+        filters: Optional[Mapping[str, Any]] = None,
     ) -> Dict[Literal["ImagesDeleted", "SpaceReclaimed"], Any]:
         """Delete unused images.
 
         The Untagged keys will always be "".
 
         Args:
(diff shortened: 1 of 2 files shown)


rhatdan:
LGTM
@umohnani8 @jwhonce @mwhahaha PTAL

inknos:
Quoting @jwhonce from the issue

> Adding external and all as kwargs would require extra validation if the client has been instantiated in compatible_version mode

I scratched my head on this comment, but I think it's just a simple check that I am missing. Maybe you could help me out with what do you mean with this :)

inknos:
Also, apologies for several force pushes, looks like I can't get pylint locally and in GH work nicely together, so it keeps complaining 🤦🏻‍♂️

jwhonce:
> Quoting @jwhonce from the issue
> 
> > Adding external and all as kwargs would require extra validation if the client has been instantiated in compatible_version mode
> 
> I scratched my head on this comment, but I think it's just a simple check that I am missing. Maybe you could help me out with what do you mean with this :)

@inknos It is possible to configure the PodmanClient to only use the compatible API for a Podman service or talk to a Docker service. I was questioning how would we handle providing these additional parameters to the call since they are not supported in that case. Should podman-py report an error? Or, should it just make the call and allow the API service to either ignore them or report an error?

inknos:
> Should podman-py report an error? Or, should it just make the call and allow the API service to either ignore them or report an error?

As long as podman-py does not crash or give a traceback I like to see the api errors, more than translating what's happening. Maybe I would wrap the calls with a warning in the logs.

PR #416, opened 2024-08-02
[skip-ci] Packit: downstream_package_name for each package key

lsm5:
When the fedora package name doesn't match with the upstream repo, the `downstream_package_name` key needs to be mentioned for each package config (fedora, centos).

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -2,15 +2,16 @@
 # See the documentation for more information:
 # https://packit.dev/docs/configuration/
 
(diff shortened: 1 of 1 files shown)


lsm5:
This has been done prior for c/common upstream v/s containers-common package https://github.com/containers/common/commit/3c9661748b1423a20d9e8f3e6a2fb9bb32996634 .

lsm5:
@inknos @jwhonce @rhatdan PTAL.

PR #418, opened 2024-08-12
Fix podman search flake + update CI VM images

cevich:
Podman search is an infrequently used feature, that tends to flake a lot. Also stop linting twice, and update CI VM images.

Example failure log: https://cirrus-ci.com/task/4801063868432384

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -15,7 +15,7 @@ env:
     #### Cache-image names to test with (double-quotes around names are critical)
     ####
     # Google-cloud VM Images
-    IMAGE_SUFFIX: "c20240320t153921z-f39f38d13"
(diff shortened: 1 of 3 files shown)


inknos:
Ugh pylint complains... Should we actually add that step as a flake? It is kinda annoying and not friendly to test it before pushing, and this leads to many force pushes. We could remove the pylint step from the chain and add it as a precommit to simplify all the checks

inknos:
Besides pylint, I favor this change.

jwhonce:
LGTM

cevich:
It looks like the `Gating test` task also runs pylint.  So I think it can just be dropped.  Regardless, something else is broken with podman-search...

cevich:
This should be ready to go now (assuming the changes are acceptable).

PR #420, opened 2024-08-15
Update CI VM images

inknos:
New image here: https://github.com/containers/automation_images/pull/376

cc: @edsantiago


inknos:
testing with [this PR](https://github.com/containers/podman-py/pull/422) I was able to isolate the environments better. I found out that for py3.9/py3.11 the tests hang, for 3.12 the tests pass the failing one but have issues later with requests. some more work is needed and I guess these two prs will need to be considered together

inknos:
@edsantiago how can I re-trigger the tests? should I make a new PR?

Maybe it's worth mentioning that this PR was made initially for an image that was created a while ago, and then it was rebased on top of another image change.

edsantiago:
Usually there's a `Rerun test` button, but I don't see it here. Maybe you and I don't have the right privs on this repo. In that case, the only possible action you can take is to repush.

inknos:
force pushed already, I'll make a new one and close this with the new link

inknos:
New PR. https://github.com/containers/podman-py/pull/430

Issue #419, opened 2024-08-13
Arguments missing for podman load and podman login?


CVC0316:
Hi all!

I'm currently using podman-py to automatically load, tag and push images to a local registry. 

Now I've run into two problems.

The first one is that the images.load method takes no arguments like a path - at least as far as I can see. If I need to load an image from a tgz archive I'd expect the load method to accept a path instead of binary only. I'm not sure how to load any images with this implementation as I'd basically have to pipe the tgz into the method somehow? With podman I have the option "-i" but I don't see anything like that in the methods currently implemented. 

Furthermore I wanted to ask if there's an option to add a flag like --tls-verify=false for the login method. That's a smaller thing that's rather easy to work around of course. Still it might be nice to have if you're working with a local registry. 

Best regards and thanks in advance!

inknos:
@CVC0316 , thanks for the report. Currently, the podman-py's api is slightly off sync with podman, and some opts/pars might be missing. I am planning to bring it back to sync, but it takes time.

Would you like to open a PR to fix it or to take care of part of it? I can help if needed.

inknos:
Fixed https://github.com/containers/podman-py/pull/434

Please reopen if some features are missing

Issue #425, opened 2024-09-12
Tests are failing with timeout error


inknos:
Reproduces in GH and locally

1. pull upstream
2. `make tests`

tests get stuck at `podman/tests/integration/test_containers.py::ContainersIntegrationTest::test_container_crud`

See https://github.com/containers/podman-py/pull/420/checks?check_run_id=28860969899

inknos:
After bisecting podman, this is the first commit where podman-py tests hang https://github.com/containers/podman/commit/8a943311db5b686af231ddc026b7a0a072eb25d2

rhatdan:
@Luap99 PTAL

inknos:
I have a reproducer which is making me scratch my head

```python
from podman import PodmanClient
import subprocess

uri = "unix:///run/user/1000/podman/podman.sock"

client = PodmanClient(base_uri=uri)

c = client.containers.create('nginx:latest', name="test", detach=True)

print(client.containers.get(c.id)) # this line hangs
```

now, if you do

```python
for i in range(5):
    print(i)
    print(client.containers.get(c.id))
```

only the last get call in the loop hangs

Luap99:
I am not familiar with the python code here but if you do not call the wait endpoint then the linked commit should not change anything as the code is only used in there unless I am overlooking something.

Looking at your reproducer it doesn't even start the container so there is no reason to call wait either, and I cannot reproduce with your small python example either unless I did something wrong? I just copied your reproducer to a file in this repo and then just run it.

inknos:
apologies I deleted the comment with the reproducer already because it was incorrect. 

> if you do not call the wait endpoint then the linked commit should not change anything as the code is only used in there unless I am overlooking something

your comment shed some light, now I actually have something that can reproduce it

```python
from podman import PodmanClient
import subprocess

uri = "unix:///run/user/100/podman/podman.sock"
client = PodmanClient(base_uri=uri)

c = client.containers.create('nginx:latest', name="test", detach=True)
c.start()
```
works

```python
c = client.containers.run('nginx:latest', name="test", detach=True)
```
hangs

I think I know where to look now

Luap99:
Yes that makes more sense, I was able to reproduce with the test case once I figure out the dependencies.

[...]

PR #426, opened 2024-09-12
Extend the parameters of 'images.load' and 'login' methods

milanbalazs:
The new parameters of `login` method based on [SystemAuth docs](https://docs.podman.io/en/latest/_static/api.html#tag/system-(compat)/operation/SystemAuth):

- **auth**
- **identitytoken**: IdentityToken is used to authenticate the user and get an access token for the registry.
- **registrytoken**: RegistryToken is a bearer token to be sent to a registry
- **tls_verify**: Whether to verify TLS certificates.

The new parameter of `images.load` method:

- **file_path**: Path of the Tarball. If it's set then the `load` method opens the file and reads it as bytes. It means the method is able to handle file path as well, not only bytes. The scenarios when none of parameters or both of them are set are handled in the method.

**Fix for:**

- https://github.com/containers/podman-py/issues/419

### DIFF
--- a/podman/api/client.py
+++ b/podman/api/client.py
@@ -320,6 +320,7 @@ class APIClient(requests.Session):
 
         Keyword Args:
             compatible: Will override the default path prefix with compatible prefix
+            verify: Whether to verify TLS certificates.
 
         Raises:
             APIError: when service returns an error
@@ -394,6 +395,7 @@ class APIClient(requests.Session):
 
         Keyword Args:
             compatible: Will override the default path prefix with compatible prefix
+            verify: Whether to verify TLS certificates.
 
         Raises:
             APIError: when service returns an error
@@ -412,6 +414,7 @@ class APIClient(requests.Session):
         # TODO should we have an option for HTTPS support?
         # Build URL for operation from base_url
         uri = urllib.parse.ParseResult(
+            # TODO: Does it make sense: "https" if kwargs.get("verify", None) else "http" ?
             "http",
             self.base_url.netloc,
             urllib.parse.urljoin(path_prefix, path),
@@ -429,6 +432,7 @@ class APIClient(requests.Session):
                     data=data,
                     headers=(headers or {}),
                     stream=stream,
+                    verify=kwargs.get("verify", None),
                     **timeout_kw,
                 )
             )
--- a/podman/api/client.py
+++ b/podman/api/client.py
@@ -411,11 +411,10 @@ class APIClient(requests.Session):
 
         path = path.lstrip("/")  # leading / makes urljoin crazy...
 
-        # TODO should we have an option for HTTPS support?
+        scheme = "https" if kwargs.get("verify", None) else "http"
         # Build URL for operation from base_url
         uri = urllib.parse.ParseResult(
-            # TODO: Does it make sense: "https" if kwargs.get("verify", None) else "http" ?
-            "http",
+            scheme,
             self.base_url.netloc,
             urllib.parse.urljoin(path_prefix, path),
             self.base_url.params,
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -3,9 +3,11 @@
 import io
 import json
 import logging
+import os
 import urllib.parse
-from typing import Any, Dict, Generator, Iterator, List, Mapping, Optional, Union
+from typing import Any, Dict, Iterator, List, Mapping, Optional, Union
 import requests
+from pathlib import Path
 
 from podman import api
 from podman.api import Literal
@@ -114,16 +116,18 @@ class ImagesManager(BuildMixin, Manager):
         )
 
     def load(
-        self, data: Optional[bytes] = None, file_path: Optional[str] = None
-    ) -> Generator[Image, None, None]:
+        self, data: Optional[bytes] = None, file_path: Optional[os.PathLike] = None
+    ) -> List[Image, None, None]:
         """Restore an image previously saved.
 
         Args:
             data: Image to be loaded in tarball format.
             file_path: Path of the Tarball.
+                       It works with both str and Path-like objects
 
         Raises:
-            APIError: when service returns an error
+            APIError: When service returns an error.
+            PodmanError: When the arguments are not set correctly.
         """
         # TODO fix podman swagger cannot use this header!
         # headers = {"Content-type": "application/x-www-form-urlencoded"}
@@ -136,10 +140,11 @@ class ImagesManager(BuildMixin, Manager):
                 "Only one parameter should be set from 'data' and 'file_path' parameters."
             )
 
+        # Load a tarball containing the image
         if file_path:
-            # Load a tarball containing the image
-            with open(file_path, "rb") as tarball_file:
-                data = tarball_file.read()  # Read the tarball file as bytes
+            # Convert to Path if file_path is a string (This works with both str and Path-like objects)
+            file_path = Path(file_path)
+            data = file_path.read_bytes()  # Read the tarball file as bytes
 
         response = self.client.post(
             "/images/load", data=data, headers={"Content-type": "application/x-tar"}
@@ -147,8 +152,7 @@ class ImagesManager(BuildMixin, Manager):
         response.raise_for_status()
 
         body = response.json()
-        for item in body["Names"]:
-            yield self.get(item)
+        return [self.get(item) for item in body["Names"]]
 
     def prune(
         self, filters: Optional[Mapping[str, Any]] = None
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -6,8 +6,8 @@ import logging
 import os
 import urllib.parse
 from typing import Any, Dict, Iterator, List, Mapping, Optional, Union
-import requests
 from pathlib import Path
+import requests
 
 from podman import api
 from podman.api import Literal
@@ -142,7 +142,8 @@ class ImagesManager(BuildMixin, Manager):
 
         # Load a tarball containing the image
         if file_path:
-            # Convert to Path if file_path is a string (This works with both str and Path-like objects)
+            # Convert to Path if file_path is a string
+            # (This works with both str and Path-like objects)
             file_path = Path(file_path)
             data = file_path.read_bytes()  # Read the tarball file as bytes
 
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -132,26 +132,7 @@ class ImagesManager(BuildMixin, Manager):
         # TODO fix podman swagger cannot use this header!
         # headers = {"Content-type": "application/x-www-form-urlencoded"}
 
-        def _generator(
-            bytearg: Optional[bytes] = None, file_path_arg: Optional[os.PathLike] = None
-        ) -> Generator[bytes, None, None]:
-            post_data = bytearg
-            # Load a tarball containing the image
-            if file_path_arg:
-                # Convert to Path if file_path is a string
-                # (This works with both str and Path-like objects)
-                file_path_object = Path(file_path_arg)
-                post_data = file_path_object.read_bytes()  # Read the tarball file as bytes
-
-            response = self.client.post(
-                "/images/load", data=post_data, headers={"Content-type": "application/x-tar"}
-            )
-            response.raise_for_status()
-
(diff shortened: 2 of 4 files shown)


milanbalazs:
I think the test failure (time-out) is connected to this issue:

- https://github.com/containers/podman-py/issues/425

inknos (commented):
@milanbalazs, thanks for the PR. tests should now be working if you want to rebase it.

Overall looks good, but I'd like to wait for the comments I asked before going on

podman/domain/system.py
```diff
@@ -55,20 +59,29 @@ def login(
             reauth: Ignored: If True, refresh existing authentication. Default: False
             dockercfg_path: Ignored: Path to custom configuration file.
                 https://quay.io/v2
+            auth = None,
```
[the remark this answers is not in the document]
milanbalazs: The `auth` parameter's documentation is empty in the API documentation as:

- https://docs.podman.io/en/latest/_static/api.html#tag/system-(compat)/operation/SystemAuth

podman/api/client.py
```diff
@@ -412,6 +414,7 @@ def _request(
         # TODO should we have an option for HTTPS support?
         # Build URL for operation from base_url
         uri = urllib.parse.ParseResult(
+            # TODO: Does it make sense: "https" if kwargs.get("verify", None) else "http" ?
             "http",
```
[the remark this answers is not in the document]
milanbalazs: Thanks. I have just done it in the new commit.

podman/domain/system.py
```diff
@@ -44,6 +44,10 @@ def login(
         registry: Optional[str] = None,
         reauth: Optional[bool] = False,  # pylint: disable=unused-argument
         dockercfg_path: Optional[str] = None,  # pylint: disable=unused-argument
+        auth: Optional[str] = None,
+        identitytoken: Optional[str] = None,
+        registrytoken: Optional[str] = None,
+        tls_verify: Optional[Union[bool, str]] = None,
```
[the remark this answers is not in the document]
milanbalazs: If you agreed the usage of `tls_verify` and let me know how it should be handled exactly, I will change it if it's needed.

milanbalazs:
I don't really understand why the UnitTest says:

```text
    @requests_mock.Mocker()
    def test_load(self, mock):
>       with self.assertRaises(PodmanError):
E       AssertionError: PodmanError not raised
```

Based on the change, it's clearly visible that the `PodmanError` should be raise in case of missing or too many arguments:

- https://github.com/containers/podman-py/pull/426/files#diff-85825cbd525470577e722fecdb9188caca30be38075e9f689358c63fb8b39595R131

@inknos Do you have any idea?

podman/domain/system.py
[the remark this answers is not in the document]
inknos: yeah I know, that's why I wanted to investigate it at the source. Anyway, add a comment in the code, please. I don't want this to block your pr

podman/domain/system.py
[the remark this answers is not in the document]
inknos: I will track it and discuss it async, it's a good thing to have and I would make it more general on a future PR. but for now let's move on with this change :)

podman/domain/images_manager.py
```diff
@@ -113,18 +113,34 @@ def get_registry_data(
             collection=self,
         )
 
-    def load(self, data: bytes) -> Generator[Image, None, None]:
+    def load(
```
inknos, line 116:
```
>>> from podman import PodmanClient
>>> client = PodmanClient()
>>> client.images.load()
<generator object ImagesManager.load at 0xffff6dde4370>
```

Headaches indeed... I am trying to figure out what is wrong here and why the tests are failing.

milanbalazs: Thanks! Perhaps just I don't see something, but I don't really understand why the exception isn't thrown.

inknos: got it.

So, this function does not raise

```python
from typing import Optional, Generator

def function(bytearg: Optional[bytes] = None, string: Optional[str] = None) -> Generator[bytes, None, None] | None:
    # no arguments or two are not allowed. Only one can be passed
    if bytearg is None and string is None:
        print("None provided: only one argument is allowed")
        raise ValueError("At least one argument must be provided.")

    if bytearg is not None and string is not None:
        print("Two provided: only one argument is allowed")
        raise ValueError("Only one argument is allowed.")

    if bytearg is not None:
        print("bytearg: {}".format(bytearg))
        yield

    if string is not None:
        print("string: {}".format(string))
        yield

    print("Debug: bytearg: {}, string: {}".format(bytearg, string))
    yield

if __name__ == "__main__":
    function()
```

while this does! why?

```python
from typing import Optional, Generator

def function(bytearg: Optional[bytes] = None, string: Optional[str] = None) -> Generator[bytes, None, None] | None:
    # no arguments or two are not allowed. Only one can be passed
    if bytearg is None and string is None:
        print("None provided: only one argument is allowed")
        raise ValueError("At least one argument must be provided.")

    if bytearg is not None and string is not None:
        print("Two provided: only one argument is allowed")
        raise ValueError("Only one argument is allowed.")

    if bytearg is not None:
        print("bytearg: {}".format(bytearg))
        return

    if string is not None:
        print("string: {}".format(string))
        return

    print("Debug: bytearg: {}, string: {}".format(bytearg, string))
    return

if __name__ == "__main__":
    function()
```
because in the first example the function is a python generator so it `yields` the values. now, with the first snippet try again with

```python

if __name__ == "__main__":
    next(function())
```

the function raises now.

---

Quick solution, make the function return and don't make it raise.

Complicated solution. I don't know right now, I like using exceptions but I am not sure if it's really possible with generators. I'll keep scratching my head on this for a bit.

What do you think @milanbalazs ?

milanbalazs: Ahh... Good to know! Thanks for your investigation. I have updated the UnitTest implementation. If I put a `return` instead of `PodmanError` exception then the result is the same, actually (Nothing happened). Because of the GeneratorType the `return` or `raise` is "activated" only then the iterator starts, so in my opinion a specific exception is better then a `StopIteration` exception which is thrown by `return` statement.

@inknos What do you think about it?

BTW, currently I am fighting with the PyLint in your gate... It's also a tricky checker :)

podman/domain/system.py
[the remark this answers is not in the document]
martinhoyer: Not necessarily related to this discussion, but since this would require adding `Union` to be imported, how about importing `from __future__ import annotations` (required for Python<3.10 in this context) and doing  `Optional[bool | str]` ?  
 i.e. not changing those `Optional[str]` to `str | None`, but since there is extra import...

..oh wait, this project still supports Python <3.8, so it's a no-go. Anyway, leaving the comment.

podman/domain/images_manager.py
martinhoyer: @inknos 

> while this does! why?
> because in the first example the function is a python generator so it yields the values.

If I understand things correctly, it's because you are creating a generator object, but not "executing it". You'd have to start iterating. By replacing `yield` with `return` it stops being a generator and is actually a regular function which *is* being executed.

podman/domain/images_manager.py
```diff
@@ -113,18 +113,34 @@ def get_registry_data(
...
+
+        if data and file_path:
+            raise PodmanError(
+                "Only one parameter should be set from 'data' and 'file_path' parameters."
+            )
+
+        if file_path:
+            # Load a tarball containing the image
+            with open(file_path, "rb") as tarball_file:
+                data = tarball_file.read()  # Read the tarball file as bytes
```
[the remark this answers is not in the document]
milanbalazs: Thanks! I will replace it!

podman/domain/images_manager.py
milanbalazs: Shell I change the `Generator` (`yield`) to a simple `return` (and return the list of results. Like: `return [self.get(item) for item in body["Names"]]`)? I didn't want to refactor the "base" mechanism of the method but if you say I should change it, I am willing to do it.

milanbalazs: I have changed the Generator to a list with return statement. It seems the tests are OK, just the Pylint complains, as I mentioned before.

inknos: > I have changed the Generator to a list with return statement. It seems the tests are OK, just the Pylint complains, as I mentioned before.

Pylint seems broken on other prs too :/ Ignore it for now.

About the iterator. I probably explained myself wrongly, apologies. We cannot change the return type of the function, or we will break the API. 

If this were the first implementation, then this would be the right path, but we need to keep the consistency here. Please, let's keep the tests with `next()` and the generator return type.

Good news though. I did some research and learned new things. Consider this code:

```python
 def function(bytearg: Optional[bytes] = None, string: Optional[str] = None) -> Generator[bytes, None, None] | None:
     def _generator(bytearg: Optional[bytes] = None, string: Optional[str] = None) -> Generator[bytes, None, None]:
         if bytearg is not None:
             print("bytearg: {}".format(bytearg))
             yield

         if string is not None:
             print("string: {}".format(string))
             yield

         print("Debug: bytearg: {}, string: {}".format(bytearg, string))
         yield

     if bytearg is None and string is None:
         print("None provided: only one argument is allowed")
         raise ValueError("At least one argument must be provided.")

     if bytearg is not None and string is not None:
         print("Two provided: only one argument is allowed")
         raise ValueError("Only one argument is allowed.")

     return _generator(bytearg, string)

if __name__ == "__main__":
    # function() # raises ValueError
    print(function(bytearg='a')) # returns a generator
```

consider this example. now we have a function which returns a generator but it's not a generator itself. so, raising exceptions work as expected and the return statement gets the execution interrupted on the right steps.

wdyt? :) @milanbalazs

milanbalazs: Nice! Thanks, this is quite useful! I have changed the code and tests based on your suggestion. I have tested and it handles the Generator as well as the exceptions as expected. Let me know if I have to change anything. :)

milanbalazs: No problem, I like to deliver quality work even if it is a lot of work. :)

podman/domain/images_manager.py
```diff
@@ -113,26 +115,52 @@ def get_registry_data(
...
+                "/images/load", data=post_data, headers={"Content-type": "application/x-tar"}
+            )
+            response.raise_for_status()
 
-        body = response.json()
-        for item in body["Names"]:
-            yield self.get(item)
+            body = response.json()
+            for item in body["Names"]:
+                yield self.get(item)
```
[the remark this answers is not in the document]
milanbalazs: I agree! I have just changed the code based on your suggestion and resolve the merge conflict.

inknos:
PR looks good now. I'd like to clean up the commits a bit and rebase. I made a test branch with just a logical squash of the commits in 2. Then I rebased it with the `upstrea/main`.

If you are ok with it, @milanbalazs , I would like to push it to this branch. It's just polishing, there is absolutely no changes in your code.

https://github.com/inknos/podman-py/tree/squash-pr-upstream-426

_(for a clean look you could also try [previewing a PR](https://github.com/containers/podman-py/compare/main...inknos:podman-py:squash-pr-upstream-426?expand=1) to see how it would look.)_

Once again, it's just a squash of the many commits, if you don't mind.

inknos:
Thanks @milanbalazs. I noticed that you created this PR from your main branch. Unfortunately I cannot edit it (maybe I can but I don't want to mess up with your main branch) and it is **good practive** to make a PR from a branch.

You could open a new PR with the fixes, link it here and about this one, "close with comment". Then I'll pass the new one.

Here are the steps to achieve it quickly if you need them.
```bash
# add the remote with the squashed commits
git remote add inknos https://github.com/inknos/podman-py.git 
# clone my branch
git fetch inknos refs/heads/squash-pr-upstream-426 
# optional: checkout to it
git checkout squash-pr-upstream-426
# optional: check what's there
git log -p
# push it to your GH and open a new PR
git push origin squash-pr-upstream-426
# optional: remove the remote and the branch once it's done
git switch main ; git branch -D squash-pr-upstream-426 ; git remote remove inknos
```

Thanks :)

milanbalazs:
@inknos Thanks for your support! Here is the new PR: https://github.com/containers/podman-py/pull/434

Issue #402, opened 2024-06-24
Container is not deleted if detach=True and remove=True are specified


landgraf:
```
remove: Delete container when the container's processes exit. Default: False.

>>> import podman
>>> podman.__version__
'5.0.0'
>>> c = client.containers.run("myimage", detach=True, remove=True, command="/bin/false")
>>> c.status
'exited'

26034e386c92  myimage                                            /bin/false            14 seconds ago     Exited (1) 14 seconds ago              interesting_williamson

```

inknos:
A note:

The following example won't reproduce the issue

```
from podman import PodmanClient

uri = "unix:///run/user/1000/podman/podman.sock"

with PodmanClient(base_url=uri) as client:
    c = client.containers.run("nginx", remove=True, command="/bin/false")

    print(c.status)
```

while 

```
from podman import PodmanClient

uri = "unix:///run/user/1000/podman/podman.sock"

with PodmanClient(base_url=uri) as client:
    c = client.containers.run("nginx", deatch=True, remove=True, command="/bin/false")

    print(c.status)
```

will reproduce

inknos:
Ok, the problem is just that [detach](https://github.com/containers/podman-py/blob/main/podman/domain/containers_run.py#L70-L71) is called before [remove](https://github.com/containers/podman-py/blob/main/podman/domain/containers_run.py#L86-L87). I can see how the options can live together, but since I would avoid delegating the removal to the user or threading/scheduling the removal.

Maybe @jwhonce might have something to add about the reasoning behind the code, or if there were some thoughts already on this detach/remove process.

PR #427, opened 2024-09-14
Remove the container in case of detach mode

milanbalazs:
Currently if the `detach=True` and `remove=True` parameters are specified the container is not deleted.

**Reproduce the issue:**

```python

uri = "unix:///run/user/1000/podman/podman.sock"

    c = client.containers.run("nginx", deatch=True, remove=True, command="/bin/false")

    print(c.status)
```

With this change if the above parameters are provided then the container object is returned and the container removing (wait method is included) is started on a separated thread.

**Fix for:**

- https://github.com/containers/podman-py/issues/402

### DIFF
--- a/podman/api/client.py
+++ b/podman/api/client.py
@@ -324,6 +324,7 @@ class APIClient(requests.Session):
 
         Keyword Args:
             compatible: Will override the default path prefix with compatible prefix
+            verify: Whether to verify TLS certificates.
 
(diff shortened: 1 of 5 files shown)


milanbalazs:
This PR includes this https://github.com/containers/podman-py/pull/426 PR. Probably it would be nice to rebase this PR after merge the previous one.

PR #431, opened 2024-09-23
Don't use `root` as default user for exec_run

aparcar:
Instead use whatever user the container uses per default, which might be `root` but may also be something else. This avoids manually figuring out the default user in case some files inside the container are specifically owned by a non-root user.

### DIFF
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -153,7 +153,7 @@ class Container(PodmanResource):
             stdin: Attach to stdin. Default: False
             tty: Allocate a pseudo-TTY. Default: False
             privileged: Run as privileged.
-            user: User to execute command as. Default: root
+            user: User to execute command as.
             detach: If true, detach from the exec command.
                 Default: False
             stream: Stream response data. Default: False
@@ -176,7 +176,6 @@ class Container(PodmanResource):
             APIError: when service reports error
         """
         # pylint: disable-msg=too-many-locals
-        user = user or "root"
         if isinstance(environment, dict):
             environment = [f"{k}={v}" for k, v in environment.items()]
         data = {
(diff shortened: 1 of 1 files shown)


jwhonce:
@aparcar Interesting change. The docker-py documentation states the default for user is `root`, but the "as built" code uses an empty string.

/cc @inknos

rhatdan:
Hardcoding to the "root" user will ignore the user within the IMage, which I believe is wrong, so this seems like the proper change.

rhatdan:
Tests are blowing up, though.

aparcar:
Looks unrelated to my changes?

inknos:
> Looks unrelated to my changes?

Thanks @aparcar for the pr

We have the same pylint issue on another PR. I am taking a look but can't see what's the cause of the issue at a first glance.

I am ok with the change but let's see if the test's flake disappears. Will come back later to this.

inknos (approved):

rhatdan:
/approved
LGTM

I have no idea why gating tests are failing?  @jwhonce @lsm5 @umohnani8 Do  you know what is going on?

inknos:
@rhatdan I am on it, PR with a fix in few minutes

Quick explaination of whwat's happening.

Currently, pylint is flaky. With this assumption, we know that sometimes pylint can be right even though it did not complain in the past for no apparent reason.

While this can seem unrelated to this PR it could still be a perfectly valid complaint such for `too-many-positional-arguments`. It can be disabled for `__init__` functions and split with `*` for public functions.

inknos:
Linking for ref. Hopefully will see green lights

https://github.com/containers/podman-py/pull/432

inknos:
@aparcar tests should be good now if you want to rebase and test it

aparcar:
Rebased

umohnani8:
LGTM

lsm5 (approved):
/approve

PR #432, opened 2024-09-24
Fix/Disable Pylint R0917

inknos:
Two ways of fixing this:
- Ignore R0917 for internal functions such as __int__
- Fix the positional arguments with *

Testing against Pylint stable.
Source of Pylint 3.3.1 docs
https://pylint.readthedocs.io/en/stable/user_guide/messages/refactor/too-many-positional-arguments.html


inknos:
@rhatdan if you want to take a look here are the fixes for the tests.

cc: @jwhonce

rhatdan:
LGTM

PR #435, opened 2024-09-26
Remove the container in case of detach mode

milanbalazs:
Currently if the `detach=True` and `remove=True` parameters are specified in  the same command, the container is not deleted.

**Reproduce the issue:**

```python

uri = "unix:///run/user/1000/podman/podman.sock"

    c = client.containers.run("nginx", deatch=True, remove=True, command="/bin/false")

    print(c.status)
```

With this change if the above parameters are provided then the container object is returned and the container removing (wait method is included) is started on a separated thread.

**Fix for:**

- https://github.com/containers/podman-py/issues/402

### DIFF
--- a/podman/domain/containers_run.py
+++ b/podman/domain/containers_run.py
@@ -1,6 +1,7 @@
 """Mixin to provide Container run() method."""
 
 import logging
+import threading
 from contextlib import suppress
 from typing import Generator, Iterator, List, Union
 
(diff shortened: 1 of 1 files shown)


rhatdan:
These options should be sent to the server, to tell it to remove the container when it is finished, I don't think it is needs to be done on the client side?

milanbalazs:
> These options should be sent to the server, to tell it to remove the container when it is finished, I don't think it is needs to be done on the client side?

I agree that would be an elegant solution, but based on the API documentation the described options are not implemented on server side (API):

- https://docs.podman.io/en/latest/_static/api.html#tag/containers/operation/ContainerStartLibpod

Currently the `containers.start` method handles only the `detachKeys` parameter and the API also waits only this query parameter:

https://github.com/containers/podman-py/blob/da8fca9b9359e568b96d7cc0550bb4465cc537fe/podman/domain/containers.py#L388-L398

In my opinion, if you would like to pass these parameters to server side (via API), then some changes are needed on server side as well (at least in the API component).

rhatdan:
OK
LGTM

PR #437, opened 2024-09-27
Renovate: Update default assignment

cevich:
@inknos PTAL.  Happy to change this to somebody(s) else if desired or remove the default assignment entirely.


inknos:
sure! thanks @cevich 
/lgtm

Issue #438, opened 2024-09-27
Add tcp-ns option for pasta network driver


aparcar:
Hi, I tried all combinations I could think of but could not figure out how to create a TCP namespace forwarding of a host port to the container. Specifically I want to add 3128 from the host to the container and with the following lines in `containers.conf` I got it working just fine:

```toml
[network]
pasta_options = ["-T", "3128:3128"]
```

However, I don't want to add this to all containers, so I tried things with podman-py to no success! The closest suggestion is available at https://github.com/containers/podman-py/issues/248#issuecomment-1463728689 however my attempts resulting in some permission errors which I did not debug further (since using `containers.conf` just worked).

```python
instance = podman_client.containers.create(
    ...
    networks={"pasta": {}},
    network_options={"pasta": ["tcp-ns=3128:3128"]},
    ...
)
```

More Ideas were things like `network_options={"pasta": ["T=3128:3128"]}` or even `network_options={"pasta": ["T", "3128:3128"]}`.

What the right way?

sbrivio-rh:
As far as I understand #439 shows how to solve this now, correct...?

Luap99:
I have never used podman-py but the json format should be `{"pasta": ["--tcp-ns","3128:3128"]}` or `{"pasta": ["-T","3128:3128"]}` The pasta code literally just uses the pasta(1) argument and passes them them 1 to 1 to pasta (exception being --map-gw)

PR #439, opened 2024-10-03
Update config.py

kajal072001:
Addressing #438 
Changes

1. Integrate Configuration Management:
Use the enhanced `PodmanConfig` class to dynamically retrieve and apply network options from `containers.conf`.

2. Utilize Detailed Logging:
Configure logging to capture detailed information for debugging purposes.

3. Validate Network Driver Support:
Confirm that the Pasta network driver supports the `-T `option and handles port mappings as expected.

### DIFF
--- a/podman/domain/config.py
+++ b/podman/domain/config.py
@@ -1,13 +1,16 @@
-"""Read containers.conf file."""
+"""Read containers.conf file and create Podman containers with network options."""
 
 import sys
 import urllib
 from pathlib import Path
 from typing import Dict, Optional
 import json
+import logging
 
 from podman.api import cached_property
 from podman.api.path_utils import get_xdg_config_home
+from podman import PodmanClient
+from podman.errors import PodmanError
 
 if sys.version_info >= (3, 11):
     from tomllib import loads as toml_loads
@@ -65,7 +68,6 @@ class PodmanConfig:
 
     def __init__(self, path: Optional[str] = None):
         """Read Podman configuration from users XDG_CONFIG_HOME."""
-
         self.is_default = False
         if path is None:
             home = Path(get_xdg_config_home())
@@ -101,7 +103,7 @@ class PodmanConfig:
                     ) from e
 
         # Read the old toml file configuration
-        if self.is_default and old_toml_file.exists():
+        if self.is_default and old_toml_file and old_toml_file.exists():
             with old_toml_file.open(encoding='utf-8') as file:
                 buffer = file.read()
             loaded_toml = toml_loads(buffer)
@@ -124,17 +126,17 @@ class PodmanConfig:
     def services(self):
         """Dict[str, ServiceConnection]: Returns list of service connections.
 
-        Examples:
-            podman_config = PodmanConfig()
-            address = podman_config.services["testing"]
-            print(f"Testing service address {address}")
-        """
+            Examples:
+                podman_config = PodmanConfig()
+                address = podman_config.services["testing"]
+                print(f"Testing service address {address}")
+            """
         services: Dict[str, ServiceConnection] = {}
 
         # read the keys of the toml file first
         engine = self.attrs.get("engine")
         if engine:
-            destinations = engine.get("service_destinations")
+            destinations = engine.get("service_destinations", {})
             for key in destinations:
                 connection = ServiceConnection(key, attrs=destinations[key])
                 services[key] = connection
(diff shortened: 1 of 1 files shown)


podman/domain/config.py
```diff
@@ -1,13 +1,16 @@
...
 import sys
 import urllib
 from pathlib import Path
 from typing import Dict, Optional
 import json
+import logging
 
 from podman.api import cached_property
 from podman.api.path_utils import get_xdg_config_home
+from podman import PodmanClient
```
inknos, line 12: This line is giving a circular import. Please, remove it. There are ways to call the client such as `podman_client = self.client`

podman/domain/config.py
```diff
@@ -155,19 +157,128 @@ def services(self):
...
+        image (str): The container image to use.
+        container_name (str): The name for the container.
+        **kwargs: Additional keyword arguments for container creation.
+
+    Returns:
+        podman.containers.Container: The created and started container instance.
+    """
+    try:
+        # Initialize Podman client
+        podman_client = PodmanClient(base_url="unix://run/podman/io.podman")
```
inknos, line 212: This line relates to the fix above

podman/domain/config.py
```diff
@@ -155,19 +157,128 @@ def services(self):
...
+            port_mapping=port_mapping,
+            image=image,
+            container_name=container_name,
+            **additional_kwargs
+        )
+
+        print(f"Container '{container.name}' is running with TCP namespace forwarding from host port {port_mapping.split(':')[0]} to container port {port_mapping.split(':')[1]}.")
+
+    except Exception as e:
+        print(f"Failed to create and start container: {e}")
```
inknos, line 284: I am perplexed about the `if __name__ == "__main__":` block. What would be it's use case?

There are circular imports failures related to it, so I would remove it. The example looks nice though so actually it could be documented.

inknos (changes requested):
Thanks for your PR, I'll help you getting it through. Let's start with the first comments and then I'll get back with an in depth analysis tomorrow or in the next days

Could you also sign your commit?

```
git commit -s --amend
```

inknos:
@kajal072001 any update on my comments? :)

podman/domain/config.py
```diff
@@ -155,19 +157,128 @@ def services(self):
...
+        Returns:
+            Dict[str, list]: A dictionary where keys are network names and values are lists of options.
+        """
+        network_opts = {}
+        network_config = self.attrs.get("network", {})
+        for network_name, config in network_config.items():
+            # Assuming network options are stored as 'pasta_options', 'bridge_options', etc.
+            options_key = f"{network_name}_options"
+            if options_key in network_config:
+                network_opts[network_name] = network_config[options_key]
```
Luap99, line 192: besides pasta_options none of the other fields exists in containers.conf so it should not invent anything.

podman/domain/config.py
```diff
@@ -155,19 +157,128 @@ def services(self):
...
+        podman_config = PodmanConfig()
+        logging.debug("Podman configuration loaded.")
+
+        # Extract network options for the specified network
+        network_opts = podman_config.network_options.get(network_name, [])
+        logging.debug(f"Original network options for '{network_name}': {network_opts}")
+
+        # Append the port mapping using '-T' flag
+        network_opts += ["-T", port_mapping]
+        logging.debug(f"Updated network options for '{network_name}': {network_opts}")
```
Luap99, line 225: This looks wrong, -T maps container ports to the host and you allow only one string. Overall I don't follow this at all.

podman/domain/config.py
```diff
@@ -155,19 +157,128 @@ def services(self):
...
+        network_config = self.attrs.get("network", {})
+        for network_name, config in network_config.items():
+            # Assuming network options are stored as 'pasta_options', 'bridge_options', etc.
+            options_key = f"{network_name}_options"
+            if options_key in network_config:
+                network_opts[network_name] = network_config[options_key]
+        return network_opts
+
+
+def create_container_with_pasta(network_name: str, port_mapping: str, image: str, container_name: str, **kwargs):
```
Luap99, line 196: There is nothing special about pasta API wise, all the networks settings should be able to be set already via the normal API.

Luap99 (changes requested):
This doesn't look right or useful to me. Containers.conf should be parsed on the server never the client.
There a few exceptions (mainly connections) but everything else should already be read by the podman service. Beside also the config file parsing here be wrong anyway.

I don't think this PR is addressing the raised issue so I don't think it makes sense

PR #440, opened 2024-10-04
Packit: enable c9s downstream update

lsm5:
rpmautospec is now enabled on c9s envs so we should be ok to use the spec file maintained upstream for c9s as well.

Fixes: RUN-2123

@inknos PTAL


inknos:
thanks @lsm5 
/lgtm

PR #443, opened 2024-10-09
Make black environment dependent on 3.12.6

inknos:
Black complains with

Python 3.12.5 has a memory safety issue that can cause Black's AST safety checks to fail. Please upgrade to Python 3.12.6 or downgrade to Python 3.12.4

Let's fix the version for black until it's fixed


inknos:
VM needs to be updated. Closing

PR #453, opened 2024-10-22
[skip-ci] RPM: remove conditionals from changelog

lsm5:
All active Fedora and CentOS Stream environments support rpmautospec. So changelog conditionals can be removed.


lsm5:
@inknos PTAL

PR #454, opened 2024-10-22
Added stream support to Container.exec_run().

MattBelle:
Added support for streaming output in Container.exec_run(). Streaming functionality should support demuxed output if specified. Return format should match docker's python library.

### DIFF
--- a/podman/api/parse_utils.py
+++ b/podman/api/parse_utils.py
@@ -8,6 +8,7 @@ from datetime import datetime
 from typing import Any, Dict, Iterator, Optional, Tuple, Union
 
 from requests import Response
+from .output_utils import demux_output
 
 
 def parse_repository(name: str) -> Tuple[str, Optional[str]]:
@@ -79,11 +80,9 @@ def frames(response: Response) -> Iterator[bytes]:
         yield response.content[frame_begin:frame_end]
 
 
-def stream_frames(response: Response) -> Iterator[bytes]:
-    """Returns each frame from multiplexed streamed payload.
-
-    Notes:
-        The stdout and stderr frames are undifferentiated as they are returned.
+def stream_frames(response: Response, demux: bool = False) -> Iterator[Union[bytes, Tuple[bytes, bytes]]]:
+    """Returns each frame from multiplexed streamed payload. If ``demux`` then output will be
+    tuples where the first position is ``STDOUT`` and the second is ``STDERR``.
     """
     while True:
         header = response.raw.read(8)
@@ -95,6 +94,10 @@ def stream_frames(response: Response) -> Iterator[bytes]:
             continue
 
         data = response.raw.read(frame_length)
+
+        if demux:
+            data = demux_output(header + data)
+
         if not data:
             return
         yield data
--- a/podman/api/parse_utils.py
+++ b/podman/api/parse_utils.py
@@ -80,9 +80,13 @@ def frames(response: Response) -> Iterator[bytes]:
         yield response.content[frame_begin:frame_end]
 
 
-def stream_frames(response: Response, demux: bool = False) -> Iterator[Union[bytes, Tuple[bytes, bytes]]]:
-    """Returns each frame from multiplexed streamed payload. If ``demux`` then output will be
-    tuples where the first position is ``STDOUT`` and the second is ``STDERR``.
+def stream_frames(
+    response: Response, demux: bool = False
+) -> Iterator[Union[bytes, Tuple[bytes, bytes]]]:
+    """Returns each frame from multiplexed streamed payload.
+
+    If ``demux`` then output will be tuples where the first position is ``STDOUT`` and the second
+    is ``STDERR``.
     """
     while True:
         header = response.raw.read(8)
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -138,12 +138,12 @@ class Container(PodmanResource):
         privileged: bool = False,
         user=None,
         detach: bool = False,
-        stream: bool = False,  # pylint: disable=unused-argument
+        stream: bool = False,
         socket: bool = False,  # pylint: disable=unused-argument
         environment: Union[Mapping[str, str], List[str]] = None,
         workdir: str = None,
         demux: bool = False,
-    ) -> Tuple[Optional[int], Union[Iterator[bytes], Any, Tuple[bytes, bytes]]]:
+    ) -> Tuple[Optional[int], Union[Iterator[Union[bytes, Tuple[bytes, bytes]]], Any, Tuple[bytes, bytes]]]:
         """Run given command inside container and return results.
 
         Args:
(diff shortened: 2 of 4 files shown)


podman/domain/containers.py
```diff
@@ -166,10 +168,13 @@ def exec_run(
             demux: Return stdout and stderr separately
```
inknos, line 168: please, add a docstring for stream
```
demux (bool): Return stdout and stderr separately
stream (bool): Stream response data. Ignored if ``detach`` is true.
    Default: False
```

podman/domain/containers.py
```diff
@@ -198,9 +203,13 @@ def exec_run(
         exec_id = response.json()['Id']
         # start the exec instance, this will store command output
         start_resp = self.client.post(
-            f"/exec/{exec_id}/start", data=json.dumps({"Detach": detach, "Tty": tty})
+            f"/exec/{exec_id}/start", data=json.dumps({"Detach": detach, "Tty": tty}), stream=stream
         )
         start_resp.raise_for_status()
+
+        if stream:
```
inknos, line 210: stream should return only if demux is not set, right? I would set

`if stream and not demux`

podman/domain/containers.py
```diff
@@ -198,9 +203,13 @@ def exec_run(
...
+            return None, api.stream_frames(start_resp, demux=demux)
```
inknos, line 211: so, if demux is conditioned above, here I would remove it from the function signature

podman/api/parse_utils.py
```diff
@@ -95,6 +98,10 @@ def stream_frames(response: Response) -> Iterator[bytes]:
             continue
 
         data = response.raw.read(frame_length)
+
+        if demux:
+            data = demux_output(header + data)
+
```
inknos, line 104: following from the comments below, is demux required here? I believe it can be checked in the exec_run function. or am I missing something

inknos (changes requested):
Thanks for the PR.

Would you add some tests like these, for example, to test that stream returns the iterator correctly and that when demux is enabled stream is ignored.

What do you think?

https://github.com/containers/podman-py/commit/0c442dd8a264c62d1690e13c49a4d5744bb66f20

podman/domain/containers.py
MattBelle: Thats not how the docker sdk works. Both can be enabled and if so it returns a stream of tuples, 1 for each line where only the relevant filepointer spot is defined and the other is `None`. For example running the equivalent to the following script as your command:

```bash
echo STDOUT
>&2 echo STDERR
```

should produce a stream of tuples like:

```python
[
    (b'STDOUT\n', None),
    (None, b'STDERR\n')
]
```

podman/api/parse_utils.py
MattBelle: If we want to imitate the docker sdk we need it here because the header contains whether the data is `STDOUT` or `STDERR` and we lose the header in the output.

podman/domain/containers.py
inknos: all right, I see. we need proper tests then for the demux, demux-stream and stream options :)

podman/api/parse_utils.py
inknos: ok so, as before, now I understand the code and I am ok with the solution. but let's test it

podman/domain/containers.py
MattBelle: Added tests for stream and stream with demux. Don't know what you had in mind for just demux, but I ended up needing to modify that test because our return was slightly misaligned with docker (empty bytestring vs None). Let me know if you think I am missing some case you'd like to see.

inknos: fair enough, I was only providing some proof of concept for testing both options. I'll try your code. thanks :)

podman/tests/integration/test_container_exec.py
```diff
@@ -49,5 +49,42 @@ def test_container_exec_run_demux(self):
...
+    def test_container_exec_run_stream(self):
+        """Test streaming the output from a long running command."""
+        container = self.client.containers.create(self.alpine_image, command=["top"], detach=True)
+        container.start()
+
+        command = []
+        for i in range(3):
+            # We want to sleep so that the lines get  processed seperately
+            command.extend(['echo', str(i), ';', 'sleep', '.1', ';'])
+        command = ['/bin/sh', '-c', ' '.join(command)]
```
inknos, line 64: is there a reason not to have the command written explicitly but with a for loop? I would prefer to have the tests more plain, and I don't think there is a special need for a `for`

podman/tests/integration/test_container_exec.py
```diff
@@ -49,5 +49,42 @@ def test_container_exec_run_demux(self):
...
+            )
+        command = ['/bin/sh', '-c', ' '.join(command)]
+        error_code, output = container.exec_run(command, stream=True, demux=True)
+
+        self.assertEqual(error_code, None)
+        for index, data in enumerate(output):
+            if index % 2 == 0:
+                self.assertEqual(data, (f'{index}\n'.encode(), None))
+            else:
+                self.assertEqual(data, (None, f'{index}\n'.encode()))
```
inknos, line 90: Same here, I would rather have more lines of codes in tests unless there is a strong reason

inknos (changes requested):

podman/domain/containers.py
inknos: @MattBelle don't forget the docstring, please

inknos:
I would say that once the comments are addressed we are good to merge. The PR is in a very nice shape now 😃

inknos:
@MattBelle any update on the open conversations? :)

podman/domain/containers.py
MattBelle: Just a docstring update isn't enough. `deatch` + `stream` as is causes a buffer error because the `header` doesn't have enough bytes to unpack. Will add logic to to skip the frame streaming if `detach` is set.

podman/tests/integration/test_container_exec.py
MattBelle: I personally find the loop easier to read as I think it better highlights the iterative nature of the expected behavior. I will change it though if this goes against the established norm for the package.

podman/domain/containers.py
MattBelle: Detach should be working as described now. Added a test for it.

podman/tests/integration/test_container_exec.py
MattBelle: As alluded to earlier, I don't have a strong enough opinion in this instance to hold up the merge so I can keep the loop. I implemented it the way you requested under the assumption that that is what the average person would prefer.

podman/tests/integration/test_container_exec.py
MattBelle: Done

inknos:
thank you @MattBelle

/lgtm
/approve

Issue #151, opened 2021-12-11
can't specify a volume via a host path


leifliddy:
According to the documentation listed here https://github.com/containers/podman-py/blob/main/podman/domain/containers_create.py

```
            volumes (Dict[str, Dict[str, str]]): A dictionary to configure volumes mounted inside
                the container. The key is either the host path or a volume name, and the value is
                a dictionary with the keys:
                - bind: The path to mount the volume inside the container
                - mode: Either rw to mount the volume read/write, or ro to mount it read-only.
                For example:
                    {'/home/user1/': {'bind': '/mnt/vol2', 'mode': 'rw'},
                     '/var/www': {'bind': '/mnt/vol1', 'mode': 'ro'}}
```
...a dictionary is needed to configure volumes inside the container.

Ok, so I'm literally using the same example to create a container. 
```
client.containers.create(image=podman_image_name, name=podman_container_name, volumes={'/home/user1/': {'bind': '/mnt/vol2', 'mode': 'rw'}}
```

However, I'm met with the following error:
```Traceback (most recent call last):
  File "/home/user1/Desktop/podman-clamav-el6/./script-podman.py", line 167, in <module>
    ensure_container_exists_and_running()
  File "/home/user1/Desktop/podman-clamav-el6/./script-podman.py", line 119, in ensure_container_exists_and_running
    run_container()
  File "/home/user1/Desktop/podman-clamav-el6/./script-podman.py", line 139, in run_container
    client.containers.create(image=podman_image_name, name=podman_container_name, volumes={'/home/user1/': {'bind': '/mnt/vol2', 'mode': 'rw'}})    
  File "/usr/lib/python3.10/site-packages/podman/domain/containers_create.py", line 224, in create
    response.raise_for_status(not_found=ImageNotFound)
  File "/usr/lib/python3.10/site-packages/podman/api/client.py", line 65, in raise_for_status
    raise APIError(cause, response=self._response, explanation=message)
podman.errors.exceptions.APIError: 500 Server Error: Internal Server Error (error creating named volume "/home/user1/": error running volume create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument)
```

It's trying to create a named volume vs using the host path --even though the documentation clearly states **The key is either the host path or a volume name**

This was MUCH easier with ```docker-py``` where you could just provide a list of mounts.

I'll dig through the code later, but in the meantime any help would be appreciated.

leifliddy:
[...]

PR #460, opened 2024-10-30
Added support for mounting directories through the volume keyword.

MattBelle:
Fixes https://github.com/containers/podman-py/issues/151

The `volume` parameter when creating containers seems to be common cause of confusion and frustration when migrating from dockerpy to podmanpy. The docker API supports mounting directories through said parameter whereas podman wants users to use the `mounts` parameter despite what the documentation claims:

https://github.com/containers/podman-py/blob/7a674914edb15b2c3e657e74f0eca778a66eb9bc/podman/domain/containers_create.py#L304-L331

Currently, if a user tries to use the `volumes` keyword for directories they will get the following error:

`podman.errors.exceptions.APIError: 500 Server Error: Internal Server Error (error creating named volume "/home/user1/": error running volume create option: names must match [a-zA-Z0-9][a-zA-Z0-9_.-]*: invalid argument)`

I added support for directory mounting via `volumes` by checking if the key matches the provided pattern. If it does then the behavior doesn't change, if it doesn't then we translate the request into the equivalent mount specification.

This change differs slightly from the official API, but should minimize friction of people migrating and hopefully increase overall adoption.

### DIFF
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -16,6 +16,8 @@ from podman.errors import ImageNotFound
 
 logger = logging.getLogger("podman.containers")
 
+NAMED_VOLUME_PATTERN = re.compile(r'[a-zA-Z0-9][a-zA-Z0-9_.-]*')
+
 
 class CreateMixin:  # pylint: disable=too-few-public-methods
     """Class providing create method for ContainersManager."""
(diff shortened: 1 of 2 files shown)


rhatdan:
Thanks @MattBelle 
/approve
LGTM
@umohnani8 @jwhonce PTAL

rhatdan:
@inknos PTAL

umohnani8:
LGTM

PR #473, opened 2024-11-28
Add pre-commit workflow

inknos:


inknos:
@jwhonce would you like to take a look at this?

cc: @lsm5 as it introduces some cirrus, make and automation changes

lsm5:
@inknos not opposed to cirrus, but just wondering if we can do without cirrus? Can we achieve the same via TMT by any chance?

lsm5:
> @inknos not opposed to cirrus, but just wondering if we can do without cirrus? Can we achieve the same via TMT by any chance?

I'm cool with dropping cirrus in a followup PR.

jwhonce:
@inknos Please update https://github.com/containers/podman-py/blob/main/CONTRIBUTING.md to reflect the change of tools. Otherwise, looks good.

PR #475, opened 2024-11-28
/run/user/$UID as fallback if XDG_RUNTIME_DIR is not set

eighthave:
XDG mentions `/run/user/$UID` as the value for `XDG_RUNTIME_DIR`:
* https://www.freedesktop.org/software/systemd/man/latest/pam_systemd.html
* https://serverfault.com/questions/388840/good-default-for-xdg-runtime-dir/727994#727994

Archlinux, Debian, RedHat, Ubuntu, etc all use `/run/user/$UID` because they follow XDG:
* https://wiki.archlinux.org/title/XDG_Base_Directory

I found this while using podman-py on a minimal headless Debian server setup, where  `/run/user/$UID` was in use  but `XDG_RUNTIME_DIR` was not set.  The existing fallback in `/tmp` did not work.

### DIFF
--- a/podman/api/path_utils.py
+++ b/podman/api/path_utils.py
@@ -7,11 +7,19 @@ import stat
 
 
 def get_runtime_dir() -> str:
-    """Returns the runtime directory for the current user"""
+    """Returns the runtime directory for the current user
+
+    The value in XDG_RUNTIME_DIR is preferred, but that is not always set, for
+    example, on headless servers. /run/user/$UID is defined in the XDG documentation.
+
+    """
     try:
         return os.environ['XDG_RUNTIME_DIR']
     except KeyError:
         user = getpass.getuser()
+        run_user = f'/run/user/{user}'
+        if os.path.isdir(run_user):
+            return run_user
         fallback = f'/tmp/podmanpy-runtime-dir-fallback-{user}'
 
         try:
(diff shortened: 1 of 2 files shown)


podman/api/path_utils.py
```diff
@@ -7,11 +7,19 @@
...
+    The value in XDG_RUNTIME_DIR is preferred, but that is not always set, for
+    example, on headless servers. /run/user/$UID is defined in the XDG documentation.
+
+    """
     try:
         return os.environ['XDG_RUNTIME_DIR']
     except KeyError:
         user = getpass.getuser()
+        run_user = f'/run/user/{user}'
+        if os.path.iddir(run_user):
```
inknos, line 21: typo? did you mean `isdir`

inknos (changes requested):
Thanks for your changes. All makes sense to me. I would like to get all green checks in our automation before approving.

please address my comment and also check your commit summary. it contains long links that you can split so the checker won't complain that are longer than 72 characters. for me it's fine to split the urls even if they won't be clickable.

once the tests pass we can give this PR a go! :)

feel free to ask if you have more questions or if you want to design a test for your use case that would be awesome

podman/api/path_utils.py
eighthave: oops sorry yes.

eighthave:
I added two unit tests and addressed all the requested changes.  And the tests all pass now.

inknos (approved):

inknos:
Thanks @eighthave 

/approve
/lgtm

PR #476, opened 2024-11-29
Enable many ruff checks

inknos:
Depends on https://github.com/containers/podman-py/pull/473


inknos:
```
podman/domain/images_manager.py:291: in ImagesManager
    body: list[dict[str, Any]],  # noqa: A003
E   TypeError: 'function' object is not subscriptable
```

The errors come from removing the deprecated typing imports for Dict, List, Tuple and Type. Specifically, list is used in place of List and it's shadowed by the `list` method. Need to think of how to go around the issue without breaking the api in the first place

podman/domain/images_manager.py
```diff
@@ -31,6 +32,8 @@
 
 logger = logging.getLogger("podman.images")
 
+builtin_list = list
+
```
jwhonce, line 36: From my reading:

```
import builtins
>>> builtins.list
```

is more explicit about what is happening in this code. We are stuck with the naming collision. I think the safety of the typing is more important than the code bloat this will introduce. Or we can stay with List[].

inknos: `typing.List` is deprecated, that's why we should move away from it if we can.

https://docs.python.org/3/library/typing.html#typing.List

I like your solution better though, it's more clear. I believe the `list` shadowing will not be problematic for users though, but the ambiguity in returns would eventually flag unexpected errors in IDEs, testing, checkers and such.

inknos: @jwhonce I updated the PR with your suggestion

jwhonce: @inknos The joys of trying to graph type checking on a language that is not statically typed.

podman/domain/images_manager.py
```diff
@@ -200,8 +202,8 @@ def prune(
         response = self.client.post("/images/prune", params=params)
         response.raise_for_status()
 
-        deleted: List[Dict[str, str]] = []
-        error: List[str] = []
+        deleted: list[dict[str, str]] = []
+        error: list[str] = []
```
jwhonce, line 206:
```suggestion
        deleted: builtins.list[dict[str, str]] = []
        error: builtins.list[str] = []
```
May be more correct. I did not research where in the scope this list() is anchored.

podman/domain/images_manager.py
```diff
@@ -409,7 +415,7 @@ def pull(
         for item in response.iter_lines():
             obj = json.loads(item)
             if all_tags and "images" in obj:
-                images: List[Image] = []
+                images: list[Image] = []
```
jwhonce, line 418:
```suggestion
                images: builtins.list[Image] = []
```
See above

podman/domain/images_manager.py
```diff
@@ -473,22 +479,22 @@ def remove(
         response.raise_for_status(not_found=ImageNotFound)
 
         body = response.json()
-        results: List[Dict[str, Union[int, str]]] = []
+        results: list[dict[str, Union[int, str]]] = []
```
jwhonce, line 482:
```suggestion
        results: builtins.list[dict[str, Union[int, str]]] = []
```

podman/domain/manager.py
```diff
@@ -2,15 +2,16 @@
...
+from collections.abc import Mapping
 
 from podman.api.client import APIClient
 
 # Methods use this Type when a subclass of PodmanResource is expected.
 PodmanResourceType: TypeVar = TypeVar("PodmanResourceType", bound="PodmanResource")
 
 
-class PodmanResource(ABC):
+class PodmanResource:
```
jwhonce, line 14: Why remove ABC base clase?

podman/domain/networks_manager.py
```diff
@@ -161,8 +161,8 @@ def list(self, **kwargs) -> List[Network]:
         return [self.prepare_model(i) for i in response.json()]
 
     def prune(
-        self, filters: Optional[Dict[str, Any]] = None
-    ) -> Dict[api.Literal["NetworksDeleted", "SpaceReclaimed"], Any]:
+        self, filters: Optional[dict[str, Any]] = None
+    ) -> dict[api.Literal["NetworksDeleted", "SpaceReclaimed"], Any]:
```
jwhonce, line 165:
```suggestion
    ) -> dict[Literal["NetworksDeleted", "SpaceReclaimed"], Any]:
```
Do we need to use api.Literal any more?

jwhonce (changes requested):
Some nits and some questions. Looking good!

podman/domain/images_manager.py
inknos: I know...

podman/domain/images_manager.py
inknos: I changed all the ambiguous for the linter, probably, for the human eye is better to be more explicit with the checks as you suggest. I'll change it this way

podman/domain/manager.py
inknos: I think I might have been too strict with this one, but my reasoning was the following. The class has no @abstract method so it's not technically an interface. Now that I think about it, we should keep it abstract to prevent the instantiation of the class. I will bring back the ABC inheritance and silence the error on that line

podman/domain/networks_manager.py
inknos: probably not, I'll handle it in a separate commit

inknos:
@jwhonce I addressed the `builtins.list` comments. now the calls are explicit within the classes with the shadowed `list` methods. Then, I removed `api.Literal` and brought back `ABC` abstract subclassing.

Also, I removed the last commit as it was more like a placeholder for future ruff checks and I don't like it upstream but rather track it in a different way.

inknos:
@jwhonce all should be good now

jwhonce (approved):
Looks good.

LGTM

PR #478, opened 2024-12-02
fix: /run/user/ is based on UID not username

eighthave:
I got one key detail wrong in #475: `/run/user` is based on UID not username. This fixes that and is better tested.


inknos:
Nice!
/approve
/lgtm

Issue #492, opened 2025-01-06
[BUG] pull() progress_bar=True does not work show JSON output as intended for compact mode


D3vil0p3r:
I wrote this small Python code to test Podman-py pull progress bar:
```python
import podman
from podman import PodmanClient
import json

client = podman.from_env()
client.ping()

repository = "docker.io/ohmyzsh/zsh"

resp = client.images.pull(repository, progress_bar=True, stream=True)
for line in resp:
    print(line)
```
`progress_bar` description says:
```
            progress_bar (bool) - Display a progress bar with the image pull progress (uses
                the compat endpoint). Default: False
```
so `Uses the compat endpoint`, that should refer the podman API parameter `compatMode=true`. Generally, if I invoke directly the Podman API pull by setting `compatMode=true` by:
```
curl -XPOST --unix-socket /run/user/1000/podman/podman.sock -H content-type:application/json "http://d/v4.0.0/libpod/images/pull?reference=docker.io/athenaos/base&compatMode=true"
```
we get the JSON progress bar, that is right:
```
{"status":"Pulling fs layer","progressDetail":{},"id":"30babffc8f09"}
{"status":"Downloading","progressDetail":{"current":3373319,"total":152137719},"progress":"[=\u003e                                                 ]  3.373MB/152.1MB","id":"30babffc8f09"}
{"status":"Downloading","progressDetail":{"current":6650119,"total":152137719},"progress":"[==\u003e                                                ]   6.65MB/152.1MB","id":"30babffc8f09"}
{"status":"Downloading","progressDetail":{"current":9386247,"total":152137719},"progress":"[===\u003e
...
```
If I use Podman Py by setting `progress_bar=True` (so `compactMode=true`), sadly I don't get the JSON output but we directly get the "animated" progress bar.
![Image](https://github.com/user-attachments/assets/cb16280b-5d52-407a-8299-3300ac0c6e4a)

The problem of it, is that we cannot manipulate this response (i.e., if I want to manipulate the output). Can you please implement a way that Podman Py pull function can show the JSON output in compact mode as occurs for Docker Py?

PR #495, opened 2025-01-07
Implement compatMode raw JSON output and fix tls_verify init on pull()

D3vil0p3r:
Fix https://github.com/containers/podman-py/issues/492

Added `compatMode` param to `pull()` function. Default as `True` in order to give a detailed progress JSON output by default like:
```
...
{'status': 'Pulling fs layer', 'progressDetail': {}, 'id': 'ef3afbc03436'}
{'status': 'Downloading', 'progressDetail': {'current': 1862903, 'total': 37308955}, 'progress': '[==>                                                ]  1.863MB/37.31MB', 'id': '1546de61b6c3'}
{'status': 'Download complete', 'progressDetail': {}, 'id': 'fa424c11eb04'}
{'status': 'Pulling fs layer', 'progressDetail': {}, 'id': '3d3d38ab8766'}
{'status': 'Downloading', 'progressDetail': {'current': 1441042, 'total': 3248294}, 'progress': '[======================>                            ]  1.441MB/3.248MB', 'id': 'ef3afbc03436'}
...
```
while `compatMode=False` still keeps the one-field `stream` JSON output:
```
{'stream': 'Copying blob sha256:30babffc8f090f7c78c263b9a1b683b34ca5f73da74911ffe942d4d8225dca57\n'}
```
This approach differs from the already present `progress_bar=True` because does not show the "fancy" progress bar but a raw detailed progressing JSON output, useful for devs want to manipulate pull output.

Furthermore, the PR fixes `tls_verify` initialization in `pull()` because, according to its description, its Default should be `True`, but it was actually as `None`.

### DIFF
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -323,6 +323,8 @@ class ImagesManager(BuildMixin, Manager):
             auth_config (Mapping[str, str]) – Override the credentials that are found in the
                 config for this request. auth_config should contain the username and password
                 keys to be valid.
+            compatMode (bool) – Return the same JSON payload as the Docker-compat endpoint.
+                Default: True.
             platform (str) – Platform in the format os[/arch[/variant]]
             progress_bar (bool) - Display a progress bar with the image pull progress (uses
                 the compat endpoint). Default: False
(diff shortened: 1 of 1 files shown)


D3vil0p3r:
@rhatdan I hope you are well. When you have free time, could you please give a quick check? "Test on Fedora" is failing because it cannot find the Alpine image, so, looking on its logs, I think it is an error not related to the PR.

rhatdan:
Thanks @D3vil0p3r 
LGTM
/approve
@jwhonce @umohnani8 PTAL

umohnani8:
Changes LGTM
Just wondering if changing the default value of tlsVerify would cause current workloads to break. I agree that the default should be fixed, but need to ensure we don't break existing environments. @jwhonce wdyt?

D3vil0p3r:
Any news about this PR?

PR #497, opened 2025-01-10
fix: accept a string for the `command` argument of Container.start

vmsh0:
This makes its interface consistent with its own documentation, and with Container.run, by allowing both a list of arguments and a single string for the `command` parameter.


rhatdan:
Thanks @vmsh0 
/approve
LGTM
PTAL @jwhonce @umohnani8 @Luap99

inknos:
/approve
/lgtm
thanks @vmsh0

inknos (approved):

Issue #493, opened 2025-01-06
userns_mode option is not honoured


vmsh0:
Hi,

It seems like userns mode is not honoured by either podman-py or the podman API.

Running podman 4.9.3 on an up-to-date Ubuntu 24.04.1 LTS virtual machine.

```
$ cat /etc/sub{uid,gid}
[some random username]:100000:65536
containers:2147483647:2147483648
[some random username]:100000:65536
containers:2147483647:2147483648
```

When creating and running the same exact container with `podman-py` and with the Podman CLI, it is apparent that in the former case the `userns_mode='auto'` option is not honoured.

I attach this test case which shows the issue clearly: https://gist.github.com/vmsh0/f32b137b6d4bf1d3dbf5b6745e92911c

One unit test creates a container using `podman-py`: the test fails as after inspecting the container you can see that `UsernsMode` is empty. The other unit test creates the same container by calling the podman CLI: in this case, the test succeeds and you can see that `UsernsMode` is set to `'private'` as expected.

As far as I can see, this fails silently. I noticed the issue by pure chance while investigating a different problem. This might potentially have security implications, as users might be expecting their containers to run in a separate user namespace while they are really running in the root namespace.

rhatdan:
Make sure the service side is restarted?

vmsh0:
> Make sure the service side is restarted?

I apologize, but I'm not sure I understand what you mean. If you are asking me whether I tried to restart the service to check if it would fix it, I did.

In the meanwhile, I have also noticed that the `remove=True` option is ignored as well. This one is simpler: the option is [explicitly removed](https://github.com/containers/podman-py/blob/e79f4af138003aa13bec8b51338b547ebc27739d/podman/domain/containers_create.py#L379) before sending off the request. The REST API does have the respective flag, and hacking the value back into the request payload with the debugger results in the container being created (and then removed) with the correct option and the test passing. Is there a special reason why the `remove` option is filtered out? If not, would you accept a pull request to stop filtering it out?

Finally, while trying to mitigate the `remove=True` option being ignored and resulting in me needing to manually clean up containers between each run of the unit tests, I have attempted setting the `replace` option. While the API exposes it, `podman-py` does not handle that option. Would you accept a pull request to handle the `replace` option?

I apologize for reporting everything in this issue; please do let me know if I should split the three things (userns, remove, replace) into three different issues.

vmsh0:
[...]

PR #499, opened 2025-01-12
Proof-of-concept: set up necessary idmappings for userns_mode=auto in Container.create()

vmsh0:
This is a proof-of-concept commit to gather initial feedback.

This addresses the issue in #493, i.e., that passing `userns='auto'` to Container.create() results in the option being silently ignored.

Before this patch, `podman-py` used to set up the `userns` API parameter when the `userns` parameter was given to Container.create(). However, upon investigation, it seems like the Podman service silently ignores the passed `userns` if parameter `idmappings` is missing.

This patch addresses this behaviour by setting up `idmappings` with neutral values (i.e., the values resulting in the same behaviour as the Podman client when called with --userns=auto without more specific options), while specifically retaining any explicit values passed by the user using the undocumented argument `idmappings`.

I am looking for some feedback about this PR:
- Is the general structure of the solution acceptable?
- I do not see any other instances where a dict merge approach is used to set up default values. Is there perhaps a better place to put the `merge_dicts()` function? Or would it be better altogether to merge the relatively small structure manually and avoid adding that function altogether?
- I see that, in general, integration testing is pretty minimal and doesn't cover very many of the possible use cases of `podman-py`. In your view, what would be a good collection of integration tests for this PR/feature? At a minimum, I will be contributing one integration test which checks that passing the `userns='auto'` results in a container with a private user namespace, with IDs not overlapping with the initial namespace, as that is my use case

### DIFF
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -739,6 +739,33 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
         if "userns_mode" in args:
             params["userns"] = {"nsmode": args.pop("userns_mode")}
 
+            # proof-of-concept
+            def merge_dicts(d1, d2):
+                for dk in d2:
+                    if dk in d1:
+                        if not isinstance(d1[dk], dict) or not isinstance(d2[dk], dict):
+                            d1[dk] = d2[dk]
+                        else:
+                            d1[dk] = merge_dicts(d1[dk], d2[dk])
+                    else:
+                        d1[dk] = d2[dk]
+                return d1
+
+            default_idmappings = {
+                "AutoUserNs": False if params["userns"]["nsmode"] != "auto" else True,
+                "AutoUserNsOpts": None if params["userns"]["nsmode"] != "auto" else {
+                    "AdditionalUIDMappings": None,
+                    "AdditionalGIDMappings": None,
+                    "PasswdFile": "",
+                    "GroupFile": "",
+                    "InitialSize": 0,
+                    "Size": 0,
+                },
+            }
+
+            old_idmappings = params.get("idmappings") or {}
+            params.update(idmappings=merge_dicts(default_idmappings, old_idmappings))
+
         if "uts_mode" in args:
             params["utsns"] = {"nsmode": args.pop("uts_mode")}
 


rhatdan:
@giuseppe PTAL

giuseppe:
would you be interested to try to fix it in Podman?  If you are not interested/familiar with the project or language, then it is fine to fix it here

vmsh0:
> would you be interested to try to fix it in Podman? If you are not interested/familiar with the project or language, then it is fine to fix it here

I would be interested in attempting that :) I am not a Go programmer by trade, but I am probably familiar enough with it for this particular fix.

I have successfully built podman yesterday, now I'm trying to get the tests to run. Can I bother you (or someone else) if I can't get the tests to run? What is the preferred channel, GitHub or something more ephemeral (e.g., IRC)?

inknos (changes requested):
maybe late to the party, but here are the comms channels :) https://github.com/containers/podman/blob/main/README.md#communications

vmsh0:
Hi, sorry for reviving this after a year. I have twice attempted to allocate some time last year to get podman to compile and try to get this fixed, but this is just not fitting in in my life at the moment.

Would the project still be interested in the Python workaround?

podman/domain/containers_create.py
```diff
@@ -833,6 +833,33 @@ def parse_host_port(_container_port, _protocol, _host):
         if "userns_mode" in args:
             params["userns"] = normalize_nsmode(args.pop("userns_mode"))
 
+            # proof-of-concept
```
giuseppe, line 836: is this comment a left over?

vmsh0: Yep. So far this patch is just a workaround I have deployed in my software to make things work, but if it is something that you would include in the library I'll clean it up and perhaps see about adding tests

giuseppe: @jwhonce are you fine with it?

jwhonce: @vmsh0 I can live with the work-around if you add more detail on why we are doing it and under what conditions it should be removed. As for testing, more the better. :-) 

@inknos The errors reported during the build don't seem to match the code in the patch. ideas?

inknos: they are only linting errors. @vmsh0

```diff
diff --git a/podman/domain/containers_create.py b/podman/domain/containers_create.py
index 066902c..308549d 100644
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -847,7 +847,9 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
 
             default_idmappings = {
                 "AutoUserNs": False if params["userns"]["nsmode"] != "auto" else True,
-                "AutoUserNsOpts": None if params["userns"]["nsmode"] != "auto" else {
+                "AutoUserNsOpts": None
+                if params["userns"]["nsmode"] != "auto"
+                else {
                     "AdditionalUIDMappings": None,
                     "AdditionalGIDMappings": None,
                     "PasswdFile": "",
```

please reformat and change the subject line to < 72 characters so all linters are happy :)

inknos: also, my two cents: I am ok with the workaround **with an open issue in Podman** that cross-references the change in the code. @vmsh0 I am active on podman-py, so if you need help working on both repos feel free to tag me around or in our matrix channel and I'll be happy to help

vmsh0: @jwhonce 
> @vmsh0 I can live with the work-around if you add more detail on why we are doing it and under what conditions it should be removed.

Here, in the code, or somewhere else?

By the way, today I have validated that the workaround is still needed in recent Podman, as last year when I discovered this I was running Podman 4.

> As for testing, more the better. :-)

:+1: 

@inknos 
> also, my two cents: I am ok with the workaround **with an open issue in Podman** that cross-references the change in the code. @vmsh0 I am active on podman-py, so if you need help working on both repos feel free to tag me around or in our matrix channel and I'll be happy to help

The best case scenario for me would be if you could provide some support in getting podman to compile so I can still try to fix it there. But if that's too much, I would still appreciate your guidance in getting this patch ready for podman-py instead

vmsh0: Uh. I tried to build latest Podman, and this time it looks like I have succeeded without any hiccups. Could also reproduce the issue there. Let me look into this a bit more, and I'll find you on Matrix if I need help.

Thank you!

PR #503, opened 2025-01-22
Onboard TMT

inknos:
cc: @lsm5


.packit.yaml
```diff
@@ -77,3 +77,20 @@ jobs:
     packages: [python-podman-fedora]
     dist_git_branches:
       - fedora-branched # rawhide updates are created automatically
+
+  # Test linting on the codebase
+  # This test might break based on the OS and lint used, so we follow fedora-latest as a reference
+  - job: tests
+    trigger: pull_request
```
lsm5, line 84: include `packages: [python-podman-fedora]` here and likewise `-centos` for centos jobs. That will avoid the test jobs that are on fedora environment but say `python-podman-centos` in the end and vice versa.

.packit.yaml
```diff
@@ -77,3 +77,20 @@ jobs:
...
+    skip_build: true
+
+  # Test integration on all the targets
+  - job: tests
+    trigger: pull_request
+    targets:
+      - fedora-all
+      - centos-stream-9
+      - centos-stream-10
+      - epel-9
```
lsm5, line 96: It might be good to have separate jobs for fedora and centos even if the tests are the same. And each separate job can use the `packages: ..` key value as mentioned in the previous comment.

FWIW, we're tending towards avoiding epel (rhel) targets on other projects as things often break / fail to build because of the slow-moving nature. But if it works here and you want to keep it, fine by me. Might wanna have epel targets tested separately using the `python-podman-rhel` key and keep it separate from centos.

.packit.yaml
lsm5: hmm, this comment would probably be relevant to the test job below and not this one.

.packit.yaml
inknos: ok I am fixing the .packit file, thanks

plans/main.fmf
```diff
@@ -0,0 +1,59 @@
...
+        summary: Run Python Podman Tests on upstream PRs
+        discover+:
+            test: /tests/all_integration
+
+    /unittest:
+        summary: Run Python Podman Unit Tests on upstream PRs
+        discover+:
+            test: /tests/unittest
+
+    /integration:
```
lsm5, line 43: just confirming that `/tests/integration` and `/tests/all_integration` are different.

plans/main.fmf
```diff
@@ -0,0 +1,59 @@
...
+      - pip3 install -r test-requirements.txt
+
+  - name: ssh configuration
+    how: shell
+    script:
+        - ssh-keygen -t ecdsa -b 521 -f /root/.ssh/id_ecdsa -P ""
+        - cp /root/.ssh/authorized_keys /root/.ssh/authorized_keys%
+        - cat /root/.ssh/id_ecdsa.pub >>/root/.ssh/authorized_keys
+
+/upstream:
```
lsm5, line 27: unless you need to run all the below test plans in parallel, you can just use `filter: tag:upstream` and have this plan specification simplified by a lot I guess.

plans/main.fmf
inknos: yeah, bad naming maybe. /integration runs coverage for integration tests on a single python version while /all_integrations runs the actual tests on all python versions

plans/main.fmf
```diff
@@ -0,0 +1,64 @@
...
+            test: /tests/integration
+
+    adjust+:
+        enabled: false
+        when: initiator is not defined or initiator != packit
+
+/downstream:
+    summary: Run Python Podman Tests on bodhi / errata and dist-git PRs
+    discover+:
+        test: /tests/integration_tests
```
lsm5, line 60: When these actually run downstream, I guess they'll need access to the podman-py test files in the upstream source, and I don't think we're packaging those yet.

You could either do the fetch tests from upstream URL that you've done on dnf IIRC. Or else you could add: 
```
  dist-git-source: true
  dist-git-merge: true
```
in this discover+ section.

Might be worth creating a draft PR on python-podman dist-git and verifying before we merge this.

tests/main.fmf
```diff
@@ -0,0 +1,23 @@
...
+    summary: Run Linting on the whole codebase
+    test: cd .. && make lint
+
+/integration:
+    tag: [ upstream, downstream ]
+    summary: Run integration tests
+    test: cd .. && make integration
+
+/unittest:
+    tag: [ upstream, downstream ]
```
lsm5, line 16: do you plan to run unittest downstream?

plans/main.fmf
inknos: yup.

now I am trying to simplify the matrix a little bit, but I still didn't figured how.

so we have 4 runs to do.

- `make lint`: runs the linting. we want to run this on fedora stable only
- `make integration`: coverage for integration tests, upstream only and ideally fedora stable only
- `make unittest`: coverage for unit tests, upstream only and fedora stable only
- `make tests`: will run all integration and unit tests. this requires python3.9->3.13 to be installed and should run everywhere

inknos: and I am so confused if I should tag the tests or rather divide the tests virtually.

.packit.yaml
```diff
@@ -77,3 +77,36 @@ jobs:
...
+      - fedora-latest-stable
+    skip_build: true
+
+  # Test integration on all the targets
+  - job: tests
+    trigger: pull_request
+    tmt_plan: /upstream
+    packages: [python-podman-fedora]
+    targets:
+      - fedora-all
```
lsm5, line 97: Given you need to depend on the podman rpm, here's a  suggestion:

1. I guess you'd want upstream main of this repo to work with upstream main of podman. So, you might wanna test upstream PRs with podman installed from the [podman-next copr](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/packages/) . You can enable this repo for packit tasks using: https://github.com/containers/crun/blob/main/.packit.yaml#L99-L103

2. Then you can configure your testing env based on distro / initiator like so: https://github.com/containers/crun/blob/main/plans/main.fmf#L6-L20 
Note: I have enabled epel there only to install bats.

This way, on downstream tests, only the official distro packages will be used, nothing from the copr.

Your call, feel free to leave this for a followup.

lsm5 (commented):
Great to see this @inknos !!

Some suggestions added. Looking good so far.

plans/main.fmf
lsm5: I'd say let's go with what you find to be simplest. Also, feel free to enable just 1 environment `fedora-latest-stable` for now. I'll be happy to merge with just that 1 env :) .

You can add other distros and versions in followups.

plans/main.fmf
lsm5: Actually, scratch my previous comment. Let's get ourselves comfortable upstream. We can deal with downstream later.

tests/main.fmf
inknos: no. and again, this is coverage of tests, so it's not even a test run. it's good to measure how much the codebase is covered

.packit.yaml
inknos: > you might wanna test upstream PRs with podman installed from the [podman-next copr](https://copr.fedorainfracloud.org/coprs/rhcontainerbot/podman-next/packages/) . You can enable this repo for packit tasks using: https://github.com/containers/crun/blob/main/.packit.yaml#L99-L103

yup that'd be nice although podman py syncs versions after podman. but to catch some regressions that's worth installing podman-next

inknos:
@lsm5 is there a way to mark tests as ok to fail? the integration coverage tests are currently failing, but that's ok because we were not running them anyway in the past (🤷🏻) and to cover for 80% or more tests it will require a separate PR.

lsm5:
> @lsm5 is there a way to mark tests as ok to fail? the integration coverage tests are currently failing, but that's ok because we were not running them anyway in the past (🤷🏻) and to cover for 80% or more tests it will require a separate PR.

Not that I know of. I had requested something like `neutral status` buttons for packit tests, but don't think that's available or even planned yet.

Is it possible to include only those tests that pass in this PR and leave the rest for followup?
> 
> Edit: I am also rebasing into a single commit since now I'd like to enable fedora-stable only.

SGTM.

lsm5 (approved):
TMT test results LGTM RE: lint tests.

Other tests and envs can be enabled in followup PRs.

LGTM

PR #507, opened 2025-02-03
Add **kwargs to Network.connect call

inknos:
This and other methods don't pass **kwargs to requests
and therefore it is not possible to enable the compatible
endpoint.


jwhonce (approved):
LGTM

PR #510, opened 2025-02-12
Bump release to 5.4.0

inknos:


inknos:
@jwhonce PTAL

we see the _infinite job flake_ again but the tests are passing

jwhonce:
LGTM

Luap99 (commented):
just as general feedback as you have asked for a review, I cannot follow any of the changes besides the obvious version number changes.

I at least would think the github action fix and tmt change must in a separate commit and explain why these are changes are made in the commit message. Like this I have no idea why they are needed, there is no link to an issue, etc... And if someone goes back later and git blame points we at something like that I would have nothing to go by why this was done.

Now I do so that the indentation in the github action but simply saying this int he commit message goes a long way for a reviewer like me.
Anyway I am not a maintainer here so merge if you like.

TomSweeneyRedHat:
LGTM
I will second @luap99 on the version being in it's own commit.  Please do that going forward.

PR #517, opened 2025-02-19
Bump release to 5.4.0.1

inknos:
The podman pyproject.toml was updated to fix a packaging issue that prevented the `import podman.api` to work.

This caused an issue in the package versioning. Python-podman cannot be bumped to version 5.4.1 since it will try to use the libpod/v5.4.1/ endpoint which is not released yet. Python-podman uses the same versioning scheme and so the package need to add a fourth digi in the version scheme to be released on PyPI. This can be avoided in distribution packages with the option of a new release or a patch downstream, but on GitHub and PyPI this is the workaround.

Source for packaging Python software
https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,9 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "podman"
-dynamic = ["version"]
+# TODO: remove the line version = ... on podman-py > 5.4.0 releases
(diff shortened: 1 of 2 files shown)


Gauravtalreja1 (approved):
/approve
/lgtm

PR #518, opened 2025-02-24
fix(doc): correction of the example for the registry at the login

dklimpel:
Fix a long standing bug in docs / function comment.

IMO this was an oversight at:
- #196 

The example "https://quay.io/v" is part of `registry: URL for registry access. For example,` and not from `dockercfg_path`


inknos (approved):

inknos:
Thanks @dklimpel 

/lgtm
/approve

PR #519, opened 2025-02-24
fix: set return type for `ImagesManager.load()` to `Image`

dklimpel:
IMO this was an oversight at:
- #434 

The PR (#434) changed the return type from `Generator[Image, None, None]` to `Generator[bytes, None, None]`

The generator was created with the return value from `get`. `get` returns an `Image`

https://github.com/containers/podman-py/blob/2a29132efab4a183580a30d9c18b9bdb61b5ab49/podman/domain/images_manager.py#L162-L168

https://github.com/containers/podman-py/blob/2a29132efab4a183580a30d9c18b9bdb61b5ab49/podman/domain/images_manager.py#L83

### DIFF
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -123,7 +123,7 @@ class ImagesManager(BuildMixin, Manager):
 
     def load(
         self, data: Optional[bytes] = None, file_path: Optional[os.PathLike] = None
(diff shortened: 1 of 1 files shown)


inknos (approved):

inknos:
Thanks @dklimpel

/lgtm
/approve

PR #520, opened 2025-02-24
Set verbose: true to debug twine uploads to PypI

inknos:
Sometimes push to Test PyPI fails with 404. It is necessary to add verbose to see what's happening.

Some examples:
https://github.com/containers/podman-py/actions/runs/13407882925/job/37451651001
https://github.com/containers/podman-py/actions/runs/13494294813/job/37698067450


inknos:
@jwhonce it should be merged so issues like https://github.com/containers/podman-py/actions/runs/14039762818/job/39306792075?pr=529 can be investigated

PR #521, opened 2025-02-24
Implement container.update()

inknos:

PR #522, opened 2025-02-27
Enable tmt downstream

inknos:
Cross referencing https://src.fedoraproject.org/rpms/python-podman/pull-request/40

@lsm5 PTAL

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -4,6 +4,26 @@
 
 upstream_tag_template: v{version}
 
+files_to_sync:
+  - src: rpm/gating.yml
+    dest: gating.yml
+    delete: true
+  - src: pyproject.toml
+    dest: pyproject.toml
+    delete: true
+  - src: plans/
+    dest: plans/
+    delete: true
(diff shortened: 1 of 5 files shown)


lsm5:
Looks like there was infrastructure issues here and on zuul and fedora-ci. Maybe retriggering would help. Changes look alright AFAICT.

inknos:
I pushed a commit from https://github.com/containers/podman-py/pull/521 that increases the timeout for python tests. It will need to be removed after https://github.com/containers/podman-py/pull/521 gets merged, I just want to see if we get the green lights.

lsm5:
@inknos f40 is failing

inknos:
@lsm5 sometimes tests are flaky. now all should be good to go

lsm5 (approved):
LGTM

lsm5:
my bad on the LGTM without ensuring https://src.fedoraproject.org/rpms/python-podman/pull-request/40 was all green. We can do a followup to resolve pending issues.

PR #524, opened 2025-03-10
skip tests when the test image is not available (e.g. 32-bit)

eighthave:
Debian and other distros still ship for 32-bit systems.  But some of the test images are not available for 32-bit archs:

* https://ci.debian.net/packages/p/python-podman/testing/armel/58665732/
* https://ci.debian.net/packages/p/python-podman/testing/armhf/58665757/
* https://ci.debian.net/packages/p/python-podman/testing/i386/58561143/

### DIFF
--- a/podman/tests/integration/test_images.py
+++ b/podman/tests/integration/test_images.py
@@ -15,6 +15,7 @@
 """Images integration tests."""
 
 import io
+import platform
 import tarfile
 import types
 import unittest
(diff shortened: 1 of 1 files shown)


inknos (changes requested):
packit jobs should be fixed now. There is one failure on Fedora Rawhide, but it should be a flake and I believe it would go away with a rerun.

@eighthave please, sign your commit and push again, so DCO bot will be happy :) this will also trigger the needed rerun. then it's good to go

PR #525, opened 2025-03-13
feat: expose `*ns` keys to consumers

josegomezr:
`cgroupns`, `ipcns`, `pidns`, `userns`, `utsns` now support setting the `value` attribute.

This allows to mimic the behavior of `--userns`[0] cli arg via this API.

[0]: https://docs.podman.io/en/stable/markdown/podman-create.1.html#userns-mode

Maybe related to #499

---

We spotted this missing behavior when using `salt-podman-formula`, we realized that setting:

```yaml
container:
  userns_mode:
     nsmode: keep-id
     value: "uid=1000"
```

Was not bringing the `.container.userns_mode.value` to the actual invocation. The patch allows consumers to send a dict with the options, there's potential to do string parsing (to mimic 1:1 the parsing that the cli does) if desired.

### DIFF
--- a/podman/domain/containers_create.py
+++ b/podman/domain/containers_create.py
@@ -436,6 +436,13 @@ class CreateMixin:  # pylint: disable=too-few-public-methods
         def pop(k):
             return args.pop(k, None)
 
+        def normalize_nsmode(
+            mode: Union[str, MutableMapping[str, str]],
+        ) -> dict[str, str]:
+            if isinstance(mode, dict):
+                return mode
+            return {"nsmode": mode}
+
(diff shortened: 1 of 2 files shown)


josegomezr:
Hmm, I took the lint suggestions from `black` but seems like the CI uses `ruff` instead?

inknos:
> Hmm, I took the lint suggestions from black but seems like the CI uses ruff instead?

Sorry got late to the party. Where's the suggestion from black? we migrated recently to ruff (which uses their black implementation) so it might be a leftover from the past workflows

inknos (approved):

Issue #533, opened 2025-04-06
Add volume inspect


rkozlo:
Podman api already exposes volume inspect under `volumes/{name}` would be nice to have it in podman-py as well

inknos:
@rkozlo are you interested in taking this? I will put it in the backlog otherwise and intend to do it in the future.

rkozlo:
@inknos crated PR

PR #535, opened 2025-04-08
Add inspect volume

rkozlo:
#533

### DIFF
--- a/podman/domain/volumes.py
+++ b/podman/domain/volumes.py
@@ -35,6 +35,23 @@ class Volume(PodmanResource):
         """
         self.manager.remove(self.name, force=force)
 
+    def inspect(self, **kwargs) -> dict:
+        """Inspect this volume
+
+        Keyword Args:
+            tls_verify (bool) - Require TLS verification. Default: True.
+
+        Returns:
+            Display attributes of volume.
+
+        Raises:
+            APIError: when service reports an error
+        """
+        params = {"tlsVerify": kwargs.get("tls_verify", True)}
+        response = self.client.get(f"/volumes/{self.id}/json", params=params)
+        response.raise_for_status()
+        return response.json()
+
 
 class VolumesManager(Manager):
     """Specialized Manager for Volume resources."""
--- a/podman/tests/unit/test_volume.py
+++ b/podman/tests/unit/test_volume.py
@@ -39,6 +39,15 @@ class VolumeTestCase(unittest.TestCase):
         volume.remove(force=True)
         self.assertTrue(adapter.called_once)
 
+    @requests_mock.Mocker()
+    def test_inspect(self, mock):
+        adapter = mock.get(
+            tests.LIBPOD_URL + "/volumes/dbase/json?tlsVerify=False", json=FIRST_VOLUME
+        )
+        vol_manager = VolumesManager(self.client.api)
+        actual = vol_manager.prepare_model(attrs=FIRST_VOLUME)
+        self.assertEqual(actual.inspect(tls_verify=False)["Mountpoint"], "/var/database")
+
 
 if __name__ == '__main__':
     unittest.main()


podman/domain/volumes.py
```diff
@@ -34,6 +34,16 @@ def remove(self, force: Optional[bool] = None) -> None:
             APIError: when service reports an error
         """
         self.manager.remove(self.name, force=force)
+    
+    def inspect(self) -> dict:
+        """Inspect this volume
+        
+        Raises:
+            APIError: when service reports an error
+        """
```
inknos, line 43: Please, could you something like podman/domain/images_manager.py:L365?
```python
params = {
    "tlsVerify": kwargs.get("tlsVerify"),
}
```

let's try to honor the option in the [docs](https://docs.podman.io/en/latest/_static/api.html#tag/manifests/operation/ManifestInspectLibpod)

podman/domain/volumes.py
```diff
@@ -34,6 +34,16 @@ def remove(self, force: Optional[bool] = None) -> None:
             APIError: when service reports an error
         """
         self.manager.remove(self.name, force=force)
+    
+    def inspect(self) -> dict:
+        """Inspect this volume
+
```
inknos, line 40: There are a bunch of white spaces that the linter is complaining about. could you please remove those?

inknos (changes requested):
Thanks for the PR, a couple of comments on how to improve this PR

1.

Tests, would you mind writing a unit test for inspect in `podman/tests/unit/test_volume.py

2.

Please sign your contribution with `git commit -s` and please add `Fixes: <link to GH issue>` for future convenience :)

podman/domain/volumes.py
rkozlo: Good point. Added

podman/domain/volumes.py
rkozlo: Cleaned. Currently using a bit raw editor.

rkozlo:
> Thanks for the PR, a couple of comments on how to improve this PR
> 
> 
> Tests, would you mind writing a unit test for inspect in `podman/tests/unit/test_volume.py
> 
> 
> Please sign your contribution with `git commit -s` and please add `Fixes: <link to GH issue>` for future convenience :)

1. Created one unit test 'test_inspect' 
2.  Sure. Next commit is signed and more appropriate commit message. I'm sorry for basic mistakes, but it is my first contribution to open project.

inknos (changes requested):
code LGTM, let's fix the automatic checks.

1. you still have one commit unsigned. I would suggest you to squash both commits into one, then sign and force push to your branch. Do you know how to?

2. there are complaints from our linter, so you should also run it to fix your contribution. we use [pre-commit](https://github.com/containers/podman-py/blob/main/CONTRIBUTING.md#coding-conventions) which will run automatically if you install it in the repo

rkozlo:
> 1. would suggest you to squash both commits into one, then si

Seems like managed to squash both commits without braking repo.
Also cleaned up code with linter
Thanks for patience and guidance. Really appreciate this

podman/domain/volumes.py
```diff
@@ -34,6 +34,33 @@ def remove(self, force: Optional[bool] = None) -> None:
             APIError: when service reports an error
         """
         self.manager.remove(self.name, force=force)
+    
+    def inspect(self) -> dict:
```
inknos, line 38: I think in your squashing you added your changes twice, please just add the second part.

(make the change, `git commit -as --amend`, fix the message, then it should look better)

inknos (changes requested):
> Seems like managed to squash both commits without braking repo.

Thanks, that's awesome. I should be more precise about how we craft commits which it's usually with a title, description if necessary (in your case I think only title is enough), references and signature, so maybe you could polish that a little bit :)

```
Add inspect volume

(Description if you feel like but, imho unneded here)

Fixes: ref link...

Signed-off-by:...
```

> Also cleaned up code with linter

I think you did but also left the old code too, see the comments above

> Thanks for patience and guidance. Really appreciate this

No problem

podman/domain/volumes.py
rkozlo: Ehh... Something went clearly wrong here. Removed this section.

rkozlo:
> Thanks, that's awesome. I should be more precise about how we craft commits which it's usually with a title, description if necessary (in your case I think only title is enough), references and signature, so maybe you could polish that a little bit :)

Changed commit message. I promise my next contribution will not be such problematic ;)

inknos (approved):
> Changed commit message. I promise my next contribution will not be such problematic ;)

None of your contributions was problematic, thanks @rkozlo

PR #537, opened 2025-04-09
Fix: fix timeout tipe as int

inknos:
Documentation and Annotations require timeout to be a float, but the API requires an integer.

@jwhonce do you know why the type was annotated as `Union[None, float, tuple[float, float], tuple[float, None]]` and not just `Union[None, float]`? Was that an historical reason?

### DIFF
--- a/podman/domain/pods.py
+++ b/podman/domain/pods.py
@@ -5,7 +5,7 @@ from typing import Any, Optional, Union
 
 from podman.domain.manager import PodmanResource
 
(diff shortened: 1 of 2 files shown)


jwhonce:
@inknos Short answer: Lost to Time, longer answer: Everywhere the _Timeout alias is used it has the same value. 

My guess is the alias was defined as a superset of all values across code base. Given it is declared in each file where it is used, specialization should not be an issue.

/lgtm

Issue #459, opened 2024-10-28
Container object uses wrong key path for ports on listed containers.


MattBelle:
The container object is currently pulling ports from `Container.NetworkSettings.Ports`:

https://github.com/containers/podman-py/blob/8a4cbf3c8aaf90bfc36b5611288b9de9becca469/podman/domain/containers.py#L59

This works when you are inspecting an individual container like:

```python
container = podman_client.containers.get(some_container_id)
debug(container.ports)
```

but fails when you get containers from a list operation eg:

```python
for container in podman_client.containers.list():
    print(container.ports)
```

Some brief playing around with the raw RestAPI makes it look like the object schema returned from those 2 endpoints differs dramatically. The correct spot in the list schema appears to be `Container.Ports`, but the format of the data would still need to be massaged if we want a coherent schema in the SDK.

Single container inspect returns:

```python
{
'18630/tcp': [
            {
                'HostIp': '0.0.0.0',
                'HostPort': '45753',
            },
        ]
}
```

while the listed container payload is:

```python
 [
        {
            'host_ip': '',
            'container_port': 18630,
            'host_port': 45753,
            'range': 1,
            'protocol': 'tcp',
        },
    ]
```

MattBelle:
Assuming we want `Container` objects with identical `ID`s to have equivalent property values regardless of their source there are couple different solutions I can think of:

1. Manually attempt to build the format provided by the inspect command. Only issue I can see with this is that the `host_ip` doesn't seem to be getting properly populated and there doesn't seem to be another spot in the response to pull that data from.
2. Post-process what is returned by the list endpoint with individual `GET` operations. Eg something like:
```python
[podman_client.containers.get(container.id) for container in podman_client.containers.list()]
```
This obviously has the downside of processing overhead. I could also see this being unwanted behavior if someone explicitly wanted the schema returned by the `LIST` object.

3. Some kinda of background upgrade procedure. This would be similar to solution 2, but would be lazily evaluated. EG if someone tried to access the ports for example we would "upgrade" the object by doing the `GET` behind the scenes and update the field states accordingly.
4. Make the list object a different type of python object to be more transparent to the users that they need to be treated differently.

I'd be happy to implement the fix, but am relatively new to contributing to and using this project so I would be interested in hearing the thoughts/preferences of people more experienced before I settle on a particular type of solution.

inknos:
[...]

Issue #446, opened 2024-10-14
TypeError when trying to get container instance status


JacobCallahan:
When trying to get the status of a container instance running on a remote podman host, I get the following error.

```python
Out[5]: <Container: 859c3172e0>

In [6]: container_inst.id
Out[6]: '859c3172e0de5e27934ba7bcbd190d0949227bfe973ded77658d25f30a1585c2'

In [7]: container_inst.status
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[7], line 1
----> 1 container_inst.status

File ~/Programming/broker/.venv/lib64/python3.12/site-packages/podman/domain/containers.py:54, in Container.status(self)
     52 """Literal["running", "stopped", "exited", "unknown"]: Returns status of container."""
     53 with suppress(KeyError):
---> 54     return self.attrs["State"]["Status"]
     55 return "unknown"

TypeError: string indices must be integers, not 'str'
```

When checking the value of attrs for the instance, I see that status isn't nested under state.
```python
In [1]: self.attrs                                                                                                                                                                            
Out[1]:                                                                                                                                                                                       
{'AutoRemove': False,                                                                                                                                                                         
 'Command': ['/bin/sh', '-c', '/tmp/startup.sh'],
...
 'StartedAt': 1728929152,
 'State': 'running',
 'Status': ''}
```

There is also a possibility that this is due to a difference in major podman versions.

Local Podman: podman-5.2.3-1.fc40.x86_64
Local podman-py: 5.0.0
Remote Podman: podman-4.9.4-4

inknos:
@JacobCallahan thanks for the report, I think it could be due to Podman 4 on the host and Podman 5 on the client. Do you have the chance to try with the same major versions?

JacobCallahan:
@inknos unfortunately, I don't have a remote podman host now, but could potentially set one up on Friday or next week.

inknos:
@JacobCallahan , I tried to reproduce it and it works on my system with podman 5 in the host and podman 4 in the server. I think trying with podman 5 in both is not necessary.

```
>>> url = "ssh://morla@192.168.1.166/run/user/1000/podman/podman.sock"
>>> from podman import PodmanClient
>>> client = PodmanClient(base_url=url, identity="/home/nsella/.ssh/id_rsa")
>>> client.containers.list()
[<Container: fd7dec82e8>, ... ]
>>> c = client.containers.get("fd7dec82e8")
>>> c.id
[...]

PR #540, opened 2025-04-14
Implement sparse keyword for containers.list()

inknos:
@jwhonce I need your guidance here. I made this pr mirroring what the docker behaviour is but this is a change for us. what's your opinion? this is a slowdown, would you rather see `sparse=True` by default to keep the behavior consistent with the current one?


jwhonce:
@inknos We could do something crazy like default to True for libpod and False when compatible is True. Is that too odd for developers? It doesn't break current libpod scripts and it will work as expected for docker scripts.

The other option is to wait for podman 6 and announce it as a breaking change.

inknos:
> We could do something crazy like default to True for libpod and False when compatible is True. Is that too odd for developers? It doesn't break current libpod scripts and it will work as expected for docker scripts.

Right, I am in for the oddness, as it's not odd at all.

_(I'll put it down for future-me, who will forget about it)_

Having `sparse=True` has two big benefits. First, not breaking anything, second, it's faster and generally a safer call (especially for scripting). Furthermore, splitting the podman/docker's behaviours is just a developer's decision. We sacrifice 1-1 compatibility for consistency, faster `list` calls, and fewer failures by retrievals of statuses for hanging containers. In my opinion, giving the option to `sparse=False` is the best. 

> The other option is to wait for podman 6 and announce it as a breaking change.

We can always do this if we get a push from the community to unify the behaviours. Then we only need to flip the default.

inknos:
Fixes: https://github.com/containers/podman-py/issues/459
Fixes: https://github.com/containers/podman-py/issues/446

jwhonce:
LGTM

@Edward5hen Please review and add the "slash lgtm" command in a comment if you agree with the change. TIA

PR #541, opened 2025-04-15
misc: add mypy for type checking

dklimpel:
Follow up to: #294 
Related to: #293 

I think that type checking helps to avoid errors and bugs.

In #294 you can see how extensive the topic is and cannot be realised with one PR. Therefore the attempt to make it incremental.

Here pipeline, configuration and a few fixes, so that it becomes transparent which files still need to be fixed in the future.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,3 +19,10 @@ repos:
   rev: 1.39.0
   hooks:
     - id: tmt-lint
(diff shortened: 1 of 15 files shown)


lsm5:
@inknos @jwhonce PTAL. 

I guess f40 test failures can be ignored given it'll be EOL soon.

inknos:
LGTM once the tests pass (I think they will since the failures are unrelated)

inknos (approved):

PR #544, opened 2025-04-25
Correctly load configuration when instantiating new client without kwargs

anishasthana:
The init function of PodmanClient was never loading the system configuration when no kwargs were specified. This change makes it so that we load the core data (base_url and identity) from the config file.


lsm5:
@anishasthana The `testing-farm:..`  tests are failing across the board in the same manner. See: https://artifacts.dev.testing-farm.io/1b8ffa29-81a8-4923-8dba-ae02e308326b/ for fedora rawhide logs.

@inknos @jwhonce PTAL.

podman/client.py
```diff
@@ -62,15 +62,21 @@ def __init__(self, **kwargs) -> None:
...
+                connection = config.attrs["Connection"]["Connections"][default_connection_name]
+                api_kwargs["base_url"] = connection["URI"]
+                api_kwargs["identity"] = connection["Identity"]
+        else:
+            if "connection" in api_kwargs:
+                connection = config.services[api_kwargs.get("connection")]
+                api_kwargs["base_url"] = connection.url.geturl()
+                # Override configured identity, if provided in arguments
+                api_kwargs["identity"] = kwargs.get("identity", str(connection.identity))
+            elif "base_url" not in api_kwargs:
```
umohnani8, line 77: @jwhonce any reason why this is technically the default case (this will always be true if api_kwargs is empty) and not the contents we get from the podman config?

podman/client.py
```diff
@@ -62,15 +62,21 @@ def __init__(self, **kwargs) -> None:
...
-        if "connection" in api_kwargs:
-            connection = config.services[api_kwargs.get("connection")]
-            api_kwargs["base_url"] = connection.url.geturl()
-
-            # Override configured identity, if provided in arguments
-            api_kwargs["identity"] = kwargs.get("identity", str(connection.identity))
-        elif "base_url" not in api_kwargs:
-            path = str(Path(get_runtime_dir()) / "podman" / "podman.sock")
-            api_kwargs["base_url"] = "http+unix://" + path
+        if not api_kwargs:
```
umohnani8, line 65: I think add a check here to ensure that `config` is also not empty and see if that helps fix the failing tests.

podman/client.py
```diff
@@ -62,15 +62,20 @@ def __init__(self, **kwargs) -> None:
...
-        if "connection" in api_kwargs:
-            connection = config.services[api_kwargs.get("connection")]
-            api_kwargs["base_url"] = connection.url.geturl()
-
-            # Override configured identity, if provided in arguments
-            api_kwargs["identity"] = kwargs.get("identity", str(connection.identity))
-        elif "base_url" not in api_kwargs:
-            path = str(Path(get_runtime_dir()) / "podman" / "podman.sock")
-            api_kwargs["base_url"] = "http+unix://" + path
+        if not api_kwargs and "Connection" in config.attrs:
```
inknos, line 65: I think this PR is bringing in a lot of value.

I would take the chance to rework this logic, making use of a more linear if/else chain. Here's a POC of what I mean.

wdyt @jwhonce @umohnani8 @anishasthana ?

```python
    # Case 1: Use named connection from kwargs if specified
    if "connection" in api_kwargs:
        connection = config.services[api_kwargs.get("connection")]
        api_kwargs["base_url"] = connection.url.geturl()
        # Override configured identity, if provided in arguments
        api_kwargs["identity"] = kwargs.get("identity", str(connection.identity))
    
    # Case 2: No kwargs provided - try to load from system configuration
    elif not api_kwargs:
        default_connection_name = config.attrs["Connection"]["Default"]
        connection = config.attrs["Connection"]["Connections"][default_connection_name]
        api_kwargs["base_url"] = connection["URI"]
        api_kwargs["identity"] = connection["Identity"]
    
    # Case 3: Fallback - if no base_url is specified, use default socket path
    if "base_url" not in api_kwargs:
        path = str(Path(get_runtime_dir()) / "podman" / "podman.sock")
        api_kwargs["base_url"] = "http+unix://" + path
```

I would also take the chance to add some logging, and, of course, a couple more tests are needed to ensure that the config priority is maintained.

anishasthana: I was worried about breaking existing behaviour, but I'm happy to move to a more linear chain!

inknos: my suggestion is to **enforce** the behaviour with integration tests. you can proceed this way if you like:

1. write tests for the current behaviour (I didn't check the test status right now), make a pr, and we merge it.
2. write the new code (in this PR) and use any approach. then we'll have tests to help understand if the behaviour is breaking

we need to ensure this is tested anyway. it's quite critical for this feature :)

inknos: @anishasthana thanks for moving it forward

> we need to ensure this is tested anyway. it's quite critical for this feature :)

I'd like to have tests for all the if/else cases in this PR. do you need help with them?

anishasthana: Hey @inknos, I'm in the middle of some travel so it'll be slow going on my end (as and when I have time). I'll reach out if I find myself stuck/confused!

Honny1:
@anishasthana Any updates on this?

Issue #547, opened 2025-04-30
Testing against podman-next breaks test_container_mounts


inknos:
This recent upstream breaking change impacted the tests https://github.com/containers/podman/pull/25942.

Probably podman-py testing should reflect only podman release branches

Impacts:
https://github.com/containers/podman-py/pull/513
https://github.com/containers/podman-py/pull/544
https://github.com/containers/podman-py/pull/546
https://github.com/containers/podman-py/pull/540

And other test runs

inknos:
cc: @lsm5

inknos:
> Probably podman-py testing should reflect only podman release branches

A better solution could be to duplicate the tests runs and run the test suite against distribution `podman` and `podman-next`. `podman-next` is used to test against features that are coming in the next release, so it's crucial to be able to test against upstream changes sometimes.
By making use of pytest [skipping](https://docs.pytest.org/en/7.1.x/how-to/skipping.html) we could run the scenarios for stable features and for future features when we have one (which would also be a considerably small run).

PR #548, opened 2025-05-02
Improve testing against distro and podman-next

inknos:
1. Split upstream into distro/pnext test runs

A new CI run targets tests that are exclusive for podman-next. /upstream plans are now split into /distro and /pnext to be explicit against which podman version is used in the tests. Also, python version used are now called base_python and all_python, to be more self explainatory.

2. Add pytest marker `pnext` to filter tests

Tests that should run against podman-next are now skipped by default in tox runs. They can be enabled with --pnext and filtered in using `-m pnext`. These two options should be used together since `pnext` and `non-pnext` tests are thought to be incompatible.

3. Split `test_container_mounts` to test a breaking change and update docs

The scenario `test_container_mounts` was split to demonstrates the usage of the marker against an upstream breaking change. CONTRIBUTING.md now reflects the new feature and explains how to use tox locally to leverage on this feature.

4. Add manual trigger on packit

Since the tests cover a corner case and serve the purpose of testing
unreleased features of libpod (which is slightly out of the scope of
podman-py) the testing jobs are optional and should run for certain PRs
only. The command to run the pnext tests is the following.

/packit test --labels {pnext,podman_next}

Fixes: https://github.com/containers/podman-py/issues/547

cc: @lsm5


.packit.yaml
```diff
@@ -130,30 +131,38 @@ jobs:
...
-    tmt_plan: /upstream/all
+    identifier: distro-fedora-all
+    tmt_plan: /distro/all_python
     packages: [python-podman-fedora]
     targets:
       - fedora-all
+
+  # run tests for the rawhide python version using podman-next packages
+  - job: tests
+    trigger: pull_request
```
lsm5, line 142: I guess this should be changed to `ignore` unless we're ok with them failing often.

lsm5 (approved):
LGTM apart from the `podman-next-fedora-base` failing often. Left a comment about that.

.packit.yaml
lsm5: Alternatively, if you use this: https://packit.dev/posts/manual-triggering#triggering-test-jobs-based-on-labels-and-identifiers , you could run the default set of tests in all cases and run the podman-next test only when the pomdman-next / pnext label is added to the PR. That can be a manual action. When the label is not present, that test will simply not run and will show a rather misleading `waiting for RPM build to succeed` status and just not run. Maybe better than a failed test though.

inknos: that is perfect!

inknos: they should be passing now, though, so I'll fix them first

inknos:
@lsm5 all is good now (except for cirrus which is red because it uses `podman-next`)

as said privately, this is a problem that we cannot get around that easily for cirrus. disabling the copr, is surely a quick fix, but it would prevent us from testing upstream changes

tests/main.fmf
```diff
@@ -24,6 +24,12 @@ require:
...
 
+    /base_python_pnext:
+        tag: [ pnext ]
+        summary: Run all tests on the base python version and podman-next
+        test: cd .. && make tests-ci-base-python-podman-next
+        duration: 5m
+
     /all_python:
         tag: [ matrix ]
         summary: Run all tests for all ptyhon versions available
```
lsm5, line 35:
```suggestion
        summary: Run all tests for all python versions available
```

Nit

lsm5 (approved):
/lgtm

lsm5 (approved):
/lgtm
/hold

lsm5 (approved):
/lgtm

PR #549, opened 2025-05-14
Bump release to 5.5.0 and enable updates-testing repos on Fedora

inknos:
This PR should enable testing on updates-latest, so we can test on 5.5.0 binaries before they get stable karma


inknos:
PTAL @jwhonce ,
Also, not a blocker, but I'd like to see [this](https://github.com/containers/podman-py/pull/546) getting merged first :)

PR #550, opened 2025-05-18
Don't try to parse the response json when not_modified

jarovo:
In case of not_modified (HTTP 304), there is no json in the response
body. Json parsing fails like:
```python
        if response.status_code == requests.codes.not_modified:
            if kwargs.get("ignore", False):
                return

        body = response
>       raise APIError(body["cause"], response=response, explanation=body)
E       TypeError: 'APIResponse' object is not subscriptable

.venv/lib/python3.12/site-packages/podman/domain/containers.py:477: TypeError
```
This patch fixes this in the way that APIError is rised with content
from the text of the requests response

### DIFF
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -472,6 +472,10 @@ class Container(PodmanResource):
         if response.status_code == requests.codes.not_modified:
             if kwargs.get("ignore", False):
                 return
+            else:
+                raise APIError(
+                    response.text, response=response, explanation="Container already stopped."
+                )
 
         body = response.json()
         raise APIError(body["cause"], response=response, explanation=body["message"])
--- a/podman/tests/integration/test_containers.py
+++ b/podman/tests/integration/test_containers.py
@@ -142,7 +142,8 @@ class ContainersIntegrationTest(base.IntegrationTest):
 
             # Try stopping the already stopped.
             # See https://github.com/containers/podman-py/pull/550 for more info.
-            top_ctnr.stop()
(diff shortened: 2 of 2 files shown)


podman/domain/containers.py
```diff
@@ -472,6 +472,10 @@ def stop(self, **kwargs) -> None:
         if response.status_code == requests.codes.not_modified:
             if kwargs.get("ignore", False):
                 return
+            else:
+                raise APIError(
+                    response.text, response=response, explanation="Container already stopped."
```
inknos, line 477: would you consider populating the response with `response.json()["message"]` rather than a hard coded explanation?

inknos (changes requested):
hey, @jarovo thanks for the contribution.

Would you consider adding some tests to this PR? I think this will greatly help to reproduce it too.

podman/domain/containers.py
jarovo: IIRC The problem this PR attempted to solve was that podman compose tried to stop stopped container. There was no json-parseable body in the response and therefore the explanation couldn't be populated from the json in that case.

I found a place in the integration tests where I can stop the stopped container and expect the APIError. I've added the test in preceeding commit so we can see how it fails. Perhaps the fix should in the API -- to always produce json-parseable body.

Thank you for suggestion to write the tests. Is there some manual to help me running the tests? @inknos

Honny1:
Hi @jarovo, can you please rebase onto upstream main and squash the commits? Thanks.

Issue #551, opened 2025-05-19
Custom contexts not supported with image builds


aseering:
I have code that constructs a tarball in memory using the `tarfile` library.  My context is relatively small (though not empty) and I already have all of its contents as variables in memory, so it's faster and simpler for me to copy them around once more in memory than it is to write a bunch of files to a temporary directory and then have podman read them back in.

I'd like to pass my generated tarball directly into podman, rather than extracting it into a temporary directory and having podman-py tar it back up for me.

This is supported by the Docker Client via the `custom_context` argument to the `build()` method.  podman-py has this argument but documents it as `(ignored)` currently.

PR #552, opened 2025-05-19
Add support for custom contexts

aseering:
A "context", as pertaining to image builds, is a Dockerfile/Containerfile plus all local files that can be referenced via commands such as COPY, etc, in the Dockerfile.  The context is typically serialized as a tarball.

Podman-py's `build()` method already offers a couple of convenient ways to automatically construct this serialized tarball.

This PR adds the ability to directly pass in an already-constructed context tarball.  This addresses #551 .


aseering:
This is my first PR here, so I'm sure I'm doing something wrong!  Let me know; happy to take a look.

inknos (approved):
thanks!

/lgtm
/approve

PR #555, opened 2025-06-13
fix: move digest label to the repository instead of tag

Alex-Izquierdo:
When parsing the repository for [pull image](https://github.com/containers/podman-py/blob/main/podman/domain/images_manager.py#L376), the digest label is part of the tag and it sends a wrong repository. This is wrong and not accepted by podman which raises an error:
`APIError: 500 Server Error: Internal Server Error ({"message":"invalid reference format"}`

The digest label (algo) is before the `:` and is expected to the be in the repo, not in the tag. 

This PR includes:

- The fix in helper function, to returns the correct tuple to be composed later. 
- Update test cases, they were defined unvalid urls with digest
- Another fix in registry_data, this is not part of the issue, I found it accidentally. It calls `parse_repository` and ignoring that it returns a tuple.


podman/domain/registry_data.py
```diff
@@ -40,7 +40,7 @@ def pull(self, platform: Optional[str] = None) -> Image:
         Args:
             platform: Platform for which to pull Image. Default: None (all platforms.)
         """
-        repository = api.parse_repository(self.image_name)
+        repository, _ = api.parse_repository(self.image_name)
```
Alex-Izquierdo, line 43: Not sure what is the usecase of this, just I found It is calling the helper ignoring that it returns a tuple. Even without this change, this code is wrong and it would be passing a tuple instead of a string.

Alex-Izquierdo:
[testing-farm:fedora-rawhide-x86_64:distro-fedora-all](https://github.com/containers/podman-py/pull/555/checks?check_run_id=44049691054) check is failing, but it does not seem related to my change. It seems some networking issue.

inknos:
/approve
/lgtm

Thanks @Alex-Izquierdo

Issue #545, opened 2025-04-28
Accessing 'podman machine' API


timonviola:
Hi

Is it possible to somehow interact with the `podman machine` commands through this package?

inknos:
Hey, not at the moment.

What's your use case?

Would you like to work on the implementation? (I can help with PR reviews or discussions it if you like)

timonviola:
@inknos Thanks for the quick answer.

> What's your use case?

Nothing I cannot solve with the CLI right now, so ultimately I can accept relying on that.


> Would you like to work on the implementation? (I can help with PR reviews or discussions it if you like)

Sounds interesting, I tried to find the API spec, is it this one? https://docs.podman.io/en/latest/_static/api.html
I don't see anything for the `machine`s, can you give some high-level pointers on how to implement this?

inknos:
The link you provided brings to the docs for the `podman` (or rather `libpod`) API, that's absolutely right.

`podman machine` is not part of it. Why? While `machine` is a core part of `podman`, it also solves specific use cases. I would say that `podman machine` on Linux solves no problem, but it enables us to run containers on Windows and macOS.

So here's the thing. `podman-py` is a package that mostly (or solely? I am not sure) conveniently exposes the `libpod`'s RESTful API to python bindings. TBH I am not sure if `podman-py` does other things which are not part of the API. Although it could, maybe it just does not.

I believe your question is valid, although it has to find a purpose in the implementation, which requires some considerations.

Evaluate a typical use case for `podman machine` on [Windows](https://github.com/containers/podman/blob/main/docs/tutorials/podman-for-windows.md#machine-init-process), for example.
The user has to `init`, `start` the machine, then run `podman` from inside the machine (ie interacting with the unix socket, being in the Linux Kernel user space).

Where does `podman-py` fit? Say that you install `podman-py` on Windows to conveniently start the machine. Then another installation would be required inside the machine to run it. So that's tricky and maybe beyond the scope of `podman-py` (or at least I would subpackage this functionality and run something like `pip install podman[machine/utils]` to split it from the API package).

Definitely an interesting conversation to have about more `podman-py` functionalities and features for version 6

timonviola:
Thanks for the detailed answer - it makes a lot of sense that `podman machine` is only relevant for macOS and Windows.
I also see your point (and agree) that podman-py defines itself as a `library of bindings to use the RESTful API`.

apyrgio:
[...]

PR #558, opened 2025-06-19
Introduce PodmanCommand

apyrgio:
Introduce the `PodmanCommand` class, which is responsible for providing a Pythonic interface over the `podman` binary. The only prerequisite is that the `podman` binary must be installed in the user's system.

More specifically, this class brings the following improvements to the codebase:

1. Allows users to run arbitrary Podman commands, in a more friendly interface than subprocess.run().
2. Provides a Pythonic interface for `podman machine` commands, where arguments are type-checked and results are native Python objects.
3. Allows Linux users to easily start a Podman REST API using a new context manager (`with PodmanCommand.service()`)

As an example, Linux users can now do the following:

```python
import podman
import os

XDG_RUNTIME_DIR = os.environ.get("XDG_RUNTIME_DIR", "/tmp")
SOCK_URI = f"unix:///{XDG_RUNTIME_DIR}/podman/podman.sock"

# Run an arbitrary Podman command.
command = podman.PodmanCommand()
version = command.run(["--version"])
print(f"Podman version is: {version}")

# Start an API service and connect the PodmanClient to it.
with command.service(uri=SOCK_URI) as p:
    with podman.PodmanClient(base_url=SOCK_URI) as client:
        containers = client.containers.list()
        print(f"Containers: {containers}")
```

Also, macOS users can work more easily with Podman machines:

```python
import podman

# Run `podman machine` commands through a Pythonic interface.
command = podman.PodmanCommand()
command.options.log_level = "debug"
command.machine.init("dz", now=True, capture_output=False)
machines = command.machine.list()
print(f"Machines: {machines}")

# Or run arbitrary Podman commands for which podman-py has no support yet.
connection = command.run(["system", "connection", "list"]).split("\n")[1]
uri = connection.split()[1]
identity = connection.split()[2]
print(f"Using socket URI {uri}")
print(f"Using identity {identity}")

# Start an API service and connect the PodmanClient to it.
with podman.PodmanClient(base_url=uri, identity=identity) as client:
    containers = client.containers.list()
    print(f"Containers: {containers}")
```

This PR is not totally complete yet, since tests are missing for `PodmanCommand.machine`. Also, I'm not sure if the docstrings are in the style that this repo expects. In any case, before polishing this PR, I'd like to get some feedback about this approach. Let me know if it's something that works for you, or if you prefer a different way.

Refs #545


inknos:
@jwhonce please take a look

inknos:
Thanks for the PR.

This will require some time and discussion, so here's my initial comment. Please, @apyrgio , correct me if I understand your PR wrong.

If I am getting it right, you want to run `podman machine` as part of Podman Py calls. The new class will fill the gap between the RESTful API and `podman`'s command line arguments (from the top of my mind, the gaps are only `podman machine` and `podman compose`, but there could be a couple more). The way you thought about it, is by separating the logic into `PodmanCommand` class. This will reduce the overlap between `PodmanClient` and `PodmanCommand` to minimal or none.

If this is correct, I would **very soon** consider splitting this feature in a subpackage. This is a nice feature, but podman-py provides binding to the API, so it's not a core feature. So the final goal **has to be** to differentiate between `pip install podman`, and `pip install podman[cli]`. I am quite convinced that this should be a hard requirement for `podman-py`

apyrgio:
Hi @inknos! Yeap, I think you got the gist of it (although I'm not really sure what you mean with "This will reduce the overlap between PodmanClient and PodmanCommand to minimal or none.")

As for splitting this in a subpackage, that's makes sense to me, I do get that the repo mainly involves around Podman's REST API. One thing I wanted to note is that I've purposefully added the core features necessary to start the REST API service, either via `PodmanCommand.machine` on Windows/macOS or `PodmanCommand.service` on Linux. So, if the logic were to go in a `podman[cli]` subpackage, it would be very barebones. Perhaps a `podman[bootstrap]` subpackage would be more fitting? 

One last question, I wonder how users across OSes currently bootstrap this API. In practice, Windows/macOS users have to work with `podman machine` commands to bring it up. Linux users have to run `podman system service` to do the same. I guess people do this manually, or have their own CLI glue, but let me know if I'm missing something.

inknos:
> This will reduce the overlap between PodmanClient and PodmanCommand to minimal or none

It means that I really don't want to have duplicate features. The cli features should be implemented as part of api calls only, and having `PodmanCommand` could potentially introduce overlaps and double some maintenance costs.

> I've purposefully added the core features necessary to start the REST API service, either via PodmanCommand.machine on Windows/macOS or PodmanCommand.service on Linux.

yup, it does make sense from a design perspective, but `machine` on Linux is useless.

> Linux users have to run podman system service to do the same. I guess people do this manually, or have their own CLI glue, but let me know if I'm missing something

Right, under the hood, `podman` has a client server implementation. We provide `podman` and `podman-remote` binaries, which differ only for how they interact with libpod underneath. for `podman-remote` you have to enable and run `podman.socket`. technically speaking, you always run `podman-remote` when you type `podman` in win/macos. that's one of the reasons why everything is done in a `podman machine` (many reasons actually, but one of them is that you can talk with the server via socket). Python Podman does essentially the same and talks as a client, so needs a socket (systemd on linux and machine on win/macos).

apyrgio:
> It means that I really don't want to have duplicate features. The cli features should be implemented as part of api calls only, and having PodmanCommand could potentially introduce overlaps and double some maintenance costs.

Ok, got it. I'm totally behind that, I don't want to reimplement a feature of `PodmanClient` in `PodmanCommand`.

Also, thanks for the rest of the clarifications :slightly_smiling_face: .

inknos:
One more thing that comes to mind, and then I'll let the PR rest for a bit, is that in podman-py we don't test on win/macos. The reason is that the userbase of podman-py does not interact with machines, and so the code is run within linux all the time, whether it's a VM or on someone's machine. I am not sure how tricky it would be to start testing on different OS, surely interesting, but could be complex. Maybe the tests could be run on Podman's CI. 🤔

apyrgio:
Well, for better or worse, we (the [Dangerzone](https://github.com/freedomofpress/dangerzone/) team) are going to make use of `podman-py` on Windows and macOS. Which means that we will encounter the same issue, i.e., how do we make sure that what we've put together works.

What we plan to do in the immediate feature is take advantage of the [WSL2 support](https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md) in the `windows-2025` GitHub runner image, and use it to test the `podman machine` commands.

If this code splits into a different subpackage and you're ok with integration testing on GHA, I'd be more than happy to assist there and do some knowledge sharing.

apyrgio:
Hey folks. Let me know if there's any actionable item I can do from my side to get this moving, or if you want to shelf it. It will help us find a proper place for this code internally.

apyrgio:
Update regarding this part:

> What we plan to do in the immediate feature is take advantage of the [WSL2 support](https://github.com/actions/runner-images/blob/main/images/windows/Windows2025-Readme.md) in the windows-2025 GitHub runner image, and use it to test the podman machine commands.

We have actually used the code in this PR successfully, to run our CI tests that require nested virtualization on GitHub Actions. To our surprise, it worked not only on `windows-2025` runners, but on `windows-2022` and `macos-13` runners as well. Kudos go to the Podman devs, our wrapper didn't add any extra logic to make it happen.

inknos:
> We have actually used the code in this PR successfully, to run our CI tests that require nested virtualization on GitHub Actions. To our surprise, it worked not only on windows-2025 runners, but on windows-2022 and macos-13 runners as well. Kudos go to the Podman devs, our wrapper didn't add any extra logic to make it happen.

glad to know

apologies for the delay with this PR but I am a bit overwhelmed and couldn't use much time on podman-py. I didn't forget, about i, I'll come back to it.

Andre-85:
Hi,
i just tested today your great work.

I used the command.service context manager from the example you given and also PodmanCommand for using "podman import" which does not have a pythonic interface. Worked like a charm, the only thing i noticed is that command.service as no default for uri, but PodmanClient has one for base_url: https://github.com/containers/podman-py/blob/df7bca46b979b15e96124af4dce5e08f10709b12/podman/client.py#L80

Greetings,
André

Honny1:
@apyrgio Any update on this?

apyrgio:
> @apyrgio Any update on this?

Uhm, the last update I know is from @inknos: https://github.com/containers/podman-py/pull/558#issuecomment-3160469746. I assumed that the Podman team may not be super interested on having this functionality, which is why I haven't commented something since :shrug: .

With regards to our project, a good chunk of this code is shipped to Windows/macOS/Linux users, and no bug has been reported so far. If there's interest in merging it, or if you have any review comments, I'd be happy to make the necessary changes. Well, for as long as we keep our Python code that is, because we will switch to Rust in the near-term future, and it will not make any sense to maintain it.

podman/errors/__init__.py
```diff
@@ -21,18 +22,24 @@
     'NotFound',
     'NotFoundError',
     'PodmanError',
+    'PodmanNotInstalled',
+    'ServiceTerminatedServiceTimeout',
```
Honny1:
```suggestion
 'ServiceTerminated',
 'ServiceTimeout',
```

podman/tests/integration/utils.py
```diff
@@ -41,54 +41,27 @@ def __init__(
...
-        if podman_exe is None:
-            raise errors.PodmanNotInstalled()
+        self.podman = PodmanCommand(path=podman_path, privileged=privileged)
 
+        self.timeout = timeout
+        self.socket_uri: str = socket_uri
         self.socket_file: str = socket_uri.replace('unix://', '')
         self.log_level = log_level
 
         self.proc: Optional[subprocess.Popen[bytes]] = None
```
Honny1: This is never assigned.

Honny1 (commented):
I got some comments.

podman/errors/__init__.py
apyrgio: Ouch. Fixed in 47a9b4a81bed65bc0c954c7d05dc4e7e7a76fe7e.

podman/tests/integration/utils.py
apyrgio: Sorry, this is an artifact of a rewrite I did right before the PR. I've fixed it in 2b564f116aefa6de376472885fb7cb39167cfca1, but I want to make sure that CI tests run.

apyrgio:
Thanks for the quick review @Honny1. I've pushed some fixups in the branch based on the failing tests (that I saw of), and rebased on top of the latest `main` branch.

Honny1:
I re think idea and agree with @inknos. I think it should exists in some sort of subpackape. WDYT? @containers/podman-maintainers @timcoding1988

Issue #528, opened 2025-03-24
Image build doesn't use caches


moneytoo:
When I repeatedly attempt to build an image, it always assembles it without using any cached layer. I tried various options (`nocache=False`, `layers=True` etc.) but was unsuccessful.

My Dockerfile/Containerfile looks for example like this:

```Dockerfile
FROM docker.io/library/fedora:latest

RUN dnf update -y

RUN echo 'Hey'
```

Building the image using this library always re-executes all the steps:

```python3
import podman

client = podman.PodmanClient(base_url="unix:///tmp/podman.sock")

print(client.version())

image, logs = client.images.build(
	dockerfile="Dockerfile",
	tag="localhost/test-podman",
	path="src/",
	rm=False,
)
for log in logs:
	print(log)
```

However if I use the [docker](https://pypi.org/project/docker/) library instead, it uses caches:

```python3
import docker

client = docker.DockerClient(base_url='unix:///tmp/podman.sock')

print(client.version())

image, logs = client.images.build(
	dockerfile="Dockerfile",
	tag="localhost/test-docker",
	path="src/",
	rm=False,
)

for log in logs:
	print(log)
```

I'm experiencing this issue with podman version 5.2.2 (on Alma Linux 9), version 4.9.3 (Ubuntu 24.04 in WSL), version 5.4.1 (on mac OS).

```console
$ time python run-podman.py

real    0m11.543s
user    0m0.118s
sys     0m0.027s
$ time python run-podman.py

real    0m12.186s
user    0m0.109s
sys     0m0.033s
$ time python run-docker.py

real    0m12.512s
user    0m0.147s
sys     0m0.011s
$ time python run-docker.py

real    0m0.323s
user    0m0.132s
sys     0m0.000s
```

fulminemizzega:
If it helps, I've just encountered the same on fedora 42, it looks like docker-py is using the compat API (/build) and podman-py uses /libpod/build.

fulminemizzega:
I compared how podman-remote works to this library, by defaults it sends many more things. Using this command:
```
podman-remote --build-arg base=quay.io/fedora-ostree-desktops/base-atomic:42 -v $(pwd)/dnf-cache:/var/cache/libdnf5:z -t sidecar  .
```
I see all these parameters sent in the http request
```
{'buildargs': '{"base":"quay.io/fedora-ostree-desktops/base-atomic:42"}',
          'compatvolumes': '0',
          'dockerfile': '["Containerfile"]',
[...]

PR #559, opened 2025-06-20
fix image build not using cache

fulminemizzega:
This commit fixes issue #528 by adding a default value to parameters layers and outputformat. This change aligns the behavior with podman-remote.

@inknos take a look, I've also added a test for this PR

### DIFF
--- a/podman/domain/images_build.py
+++ b/podman/domain/images_build.py
@@ -204,8 +204,16 @@ class BuildMixin:
         if "labels" in kwargs:
             params["labels"] = json.dumps(kwargs.get("labels"))
 
-        if params["dockerfile"] is None:
-            params["dockerfile"] = f".containerfile.{random.getrandbits(160):x}"
+        def default(value, def_value):
+            return def_value if value is None else value
+
+        params["outputformat"] = default(
+            params["outputformat"], "application/vnd.oci.image.manifest.v1+json"
+        )
+        params["layers"] = default(params["layers"], True)
+        params["dockerfile"] = default(
+            params["dockerfile"], f".containerfile.{random.getrandbits(160):x}"
+        )
 
         # Remove any unset parameters
         return dict(filter(lambda i: i[1] is not None, params.items()))
--- a/podman/tests/integration/test_images.py
+++ b/podman/tests/integration/test_images.py
@@ -15,6 +15,7 @@
 """Images integration tests."""
 
 import io
+import json
 import platform
 import tarfile
 import types
@@ -143,6 +144,19 @@ class ImagesIntegrationTest(base.IntegrationTest):
         image, stream = self.client.images.build(fileobj=buffer)
         self.assertIsNotNone(image)
         self.assertIsNotNone(image.id)
+
+    def test_build_cache(self):
+        """Check that building twice the same image uses caching"""
+        buffer = io.StringIO("""FROM quay.io/libpod/alpine_labels:latest\nLABEL test=value""")
+        image, _ = self.client.images.build(fileobj=buffer)
+        buffer.seek(0)
+        _, stream = self.client.images.build(fileobj=buffer)
+        for line in stream:
+            # Search for a line with contents "-> Using cache <image id>"
+            parsed = json.loads(line)['stream']
+            if "Using cache" in parsed:
+                break
+        self.assertEqual(parsed.split()[3], image.id)
 
     def test_build_with_context(self):
         context = io.BytesIO()


podman/domain/images_build.py
```diff
@@ -204,8 +204,16 @@ def _render_params(kwargs) -> dict[str, list[Any]]:
         if "labels" in kwargs:
             params["labels"] = json.dumps(kwargs.get("labels"))
 
-        if params["dockerfile"] is None:
-            params["dockerfile"] = f".containerfile.{random.getrandbits(160):x}"
+        def default(value, def_value):
+            return def_value if value is None else value
```
inknos: having a function for this is overkill, parameters can be defined as default in kwargs calls.

consider using `kwargs.get(value, default)` like [here](https://github.com/containers/podman-py/blob/main/podman/domain/containers.py#L436-L437)

podman/domain/images_build.py
```diff
@@ -204,8 +204,16 @@ def _render_params(kwargs) -> dict[str, list[Any]]:
...
+
+        params["outputformat"] = default(
+            params["outputformat"], "application/vnd.oci.image.manifest.v1+json"
+        )
+        params["layers"] = default(params["layers"], True)
+        params["dockerfile"] = default(
+            params["dockerfile"], f".containerfile.{random.getrandbits(160):x}"
+        )
```
inknos: Once you get the parameters with `get` these lines are not needed. you need to also document the defaults in the lines above for the `build` function

podman/tests/integration/test_images.py
```diff
@@ -144,6 +145,19 @@ def test_build(self):
         self.assertIsNotNone(image)
         self.assertIsNotNone(image.id)
 
+    def test_build_cache(self):
```
inknos, line 155: more tests are probably needed to ensure that the defaults are passed, and images.build also accepts the other parameters

inknos (changes requested):
thanks @fulminemizzega for this PR. please address my comments and we'll move forward :) also don't hesitate to ask questions or raise concerns

podman/domain/images_build.py
fulminemizzega: I did not know of this. Makes more sense

podman/domain/images_build.py
fulminemizzega: What about the dockerfile? Just document that if omitted it is generated randomly? I do not understand why it is like this or what it is for

podman/tests/integration/test_images.py
fulminemizzega: There are defaults like those above, and then there is dockerfile that is generated randomly... maybe this is something that should be done in a unit test?

podman/domain/images_build.py
inknos: > What about the dockerfile? Just document that if omitted it is generated randomly? I do not understand why it is like this or what it is for

`dockerfile` docs are fine imho, you are changing the parameters of `layers` and `outputformat` so you should ensure that the defaults are documented

podman/tests/integration/test_images.py
inknos: > There are defaults like those above, and then there is dockerfile that is generated randomly... maybe this is something that should be done in a unit test?

I think you can do all through integration tests by inspecting the request that are passed. if you want I can write you some pseudo-code to help understand how I would design it

fulminemizzega: Yes please, I’m not familiar with this kind of stuff… how would I inspect what is requested? I’ve seen how it is mocked in unit tests, is it related?

inknos: > Yes please, I’m not familiar with this kind of stuff… how would I inspect what is requested? I’ve seen how it is mocked in unit tests, is it related?

Actually my bad, I was looking into many things at the same time and got confused. You are correct, unit tests is the place where you check you function. Also yes, you said it right, you need to mock a request and test that your image build calls the parameters correctly in the default/non-default cases.

podman/domain/images_build.py
fulminemizzega: This should be ok now. I'm a bit on the fence on this, I started thinking about using an enum type for the outputformat parameter, so I've played a bit with the code in another branch in my fork and removed all the kwargs thing and moved all the keys to function parameters, with type annotations and default values. Why do so many functions in this project use kwargs and a _render_param function? Is this something that could be useful or is it better if I leave it alone?

podman/domain/images_build.py
fulminemizzega: This should be ok, I've kept the comments very short.

podman/tests/integration/test_images.py
fulminemizzega: > more tests are probably needed to ensure that the defaults are passed, and images.build also accepts the other parameters

About the other parameters, you mean that, besides the new defaults, there should be a test should that uses all of them and inspects the request? In that case, there is another issue: while experimenting with the parameter-argument conversion (see above), I discovered that _render_params parses the "remote" key, it is not documented, it can not be used alone because _render_params requires either "path" or "fileobj" (and both are meaningless if remote is used), but other than that the build function is able to handle this, because body is initialized with None, and when all the if/elif fail it stays None which is right for remote. But if dockerfile is None, then it is generated randomly in _render_params and this will be yet another issue: "If the URI points to a tarball and the dockerfile parameter is also specified, there must be a file with the corresponding path inside the tarball" from https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImageBuildLibpod
I understand this is going quite out of the "cache" issue scope, I apologize in advance.

fulminemizzega: I've added a unit test that checks the default parameters for images.build, with it also I corrected the mock URL for two other test cases

podman/domain/images_build.py
inknos: > Why do so many functions in this project use kwargs

Historical reasons probably?

@jwhonce might give us a good answer

jwhonce: @fulminemizzega kwargs was a good escape hatch for podman-py to ease the migration for scripts from docker-py. It allowed developers access to podman features without having to port their whole scripts. From there I suspect there are areas where new development followed the old form even if it didn't make as much sense. 

The flexibility of args / kwargs have always been very Pythonic and pre-date type hinting.

fulminemizzega: I see, thanks for your answer. Would there be any value in a PR for moving most of the kwargs to typed parameters for at least images.build? I've tried to do it but my changes do break existing code (I made all the strings that are used for paths become a pathlib.Path, so the argument has to be Path("something") instead of just "something").

podman/tests/integration/test_images.py
```diff
@@ -144,6 +145,19 @@ def test_build(self):
...
+        """Check that building twice the same image uses caching"""
+        buffer = io.StringIO("""FROM quay.io/libpod/alpine_labels:latest\nLABEL test=value""")
```
Honny1: You can use FROM scratch so you don't rely on an external image. (This is nonblocking)
```suggestion
        buffer = io.StringIO("""FROM scratch\nLABEL test=value""")
```

fulminemizzega: I've followed your advice and also applied it in test_build

Honny1:
It seems like there is an issue with the tests. Any ideas? @inknos @fulminemizzega

```
=================================== FAILURES ===================================
____________________ ImagesIntegrationTest.test_image_crud _____________________

self = <podman.tests.integration.test_images.ImagesIntegrationTest testMethod=test_image_crud>

    def test_image_crud(self):
        """Test Image CRUD.
    
        Notes:
            Written to maximize reuse of pulled image.
        """
    
        with self.subTest("Pull Alpine Image"):
            image = self.client.images.pull("quay.io/libpod/alpine", tag="latest")
            self.assertIsInstance(image, Image)
            self.assertIn("quay.io/libpod/alpine:latest", image.tags)
            self.assertTrue(self.client.images.exists(image.id))
    
        with self.subTest("Inspect Alpine Image"):
            image = self.client.images.get("quay.io/libpod/alpine")
            self.assertIsInstance(image, Image)
            self.assertIn("quay.io/libpod/alpine:latest", image.tags)
    
        with self.subTest("Retrieve Image history"):
            ids = [i["Id"] for i in image.history()]
>           self.assertIn(image.id, ids)
E           AssertionError: '961769676411f082461f9ef46626dd7a2d1e2b2a38e6a44364bcbecf51e66dd4' not found in ['f88416a00aabae4c9520ee61e1692b50faf551e0d08e2196b736a9e0e63ad060', '<missing>']

podman/tests/integration/test_images.py:69: AssertionError
```

fulminemizzega:
> It seems like there is an issue with the tests. Any ideas? @inknos @fulminemizzega

For this sub case I still do not have any idea. I have had it working in 2 different systems, after resetting my podman environment (with podman system reset, since they've deprecated the boltDB db without any migration tool). Before resetting it I also had this test failing, and now I've just checked, it would fail again. I really have no idea why the libpod/alpine image's history does not contain the id of the image itself.
Just to be sure, I've run again podman system reset, and now after a pull I get:
```
# podman history libpod/alpine
ID            CREATED      CREATED BY                                     SIZE        COMMENT
961769676411  6 years ago  /bin/sh -c #(nop)  CMD ["/bin/sh"]             0B
<missing>     6 years ago  /bin/sh -c #(nop) ADD file:fe64057fbb83dcc...  5.84MB
```
which is ok.
On another system it is:
```
$ podman history libpod/alpine
ID            CREATED      CREATED BY                                     SIZE        COMMENT
d054a073bda4  6 years ago  /bin/sh -c #(nop)  CMD ["/bin/sh"]             0B
<missing>     6 years ago  /bin/sh -c #(nop) ADD file:fe64057fbb83dcc...  5.84MB
```
There is another failing test where I have a theory, and it is related to caching: subtest "Deleted unused Images" in test_images.py:test_image_crud (line 112) will fail because test_containers.py:test_container_commit leaves behind an image built from libpod/alpine. When the sub test tries to delete all the unused images it fails because the libpod/alpine image is in use by the new localhost/busybox.local:unittest.
I did not catch this because I do not run all the tests on my dev machine since it destroys all running containers... but the image history test case is a mistery.

fulminemizzega:
Disregard what I've written yesterday. The issue is related to caching and an intermediary image left over by test_containers.py:test_container_rm_anonymous_volume. This and some other steps performed in test_images.py are enough to explain everything, I think. I want to reproduce the steps with just podman and see if the results are the same.

fulminemizzega:
I have written most of the details in the last commit, I do not know if how I solved it is reasonable. I think that the cleanup should happen even if test_containers.py:test_container_rm_anonymous_volume fails, otherwise if this test fails leaving behind the built image, then the same tests in test_images.py will fail. On the other hand, the tested behavior (podman deleting anonymous volumes) is not really a concern of podman-py...

Honny1:
PTAL @inknos

fulminemizzega:
I've investigated a bit the pre-commit failing check, the tmt lint step reports an error. I can reproduce it using python 3.14 and tmt 1.39, it works fine with python 3.13. If it is not an issue, updating tmt to the latest version in .pre-commit-config.yaml (I tested 1.62.1 with python 3.14.0) fixes it.

Honny1:
It seems this needs an update. I will do update.

Honny1:
PR with bump: https://github.com/containers/podman-py/pull/606

Honny1:
It should be fixed. Can you rebase onto main?

fulminemizzega:
> It should be fixed. Can you rebase onto main?

Yes. Can I also use the bot commands to restart tests or do those require more privileges than just being the author of a PR in this repo?

Honny1 (approved):
LGTM, thanks. 

@inknos, can you do the final review?

inknos:
I would prefer the commits to be squashed together, but we don't have it in the contributing doc, (I'll make a change for the future) so it's not a blocker.

Otherwise it's clean, LGTM

fulminemizzega:
> I would prefer the commits to be squashed together, but we don't have it in the contributing doc, (I'll make a change for the future) so it's not a blocker.
> 
> Otherwise it's clean, LGTM

I can squash them. Could you give me a rough guideline on how you would like them to be?
I think doing a single commit would look a bit weird, there are things that make sense (first 3 commits, with the actual fix, docs and kwargs.get usage), others less, like the fix in https://github.com/containers/podman-py/pull/559/commits/d79786a2622bd0960b86c7ba8a9a0792d49ad8b1, or the last one that touches the container running integration test. Maybe I can squash everything in 3 commits, one for code and docs, one for tests and another for the "ancillary" changes that happened while touching this PR? Or instead squash code, docs and tests in one and the other things in another commit? What should the commit message be?

inknos:
My personal take is that every commit should not break the code, so no partial changes per commit. For example, [here](https://github.com/containers/podman-py/pull/559/commits/9dc5b8360bad644937cb5c962cb4c548a51d4c5a) you `import random` but do not use it in the same commit. I know, it's a minor thing, but it should be grouped logically. I think it would be better to have at least the bug fix in one commit, and the small fixes in a separate one.

For the commit message [this](https://github.com/containers/podman-py/pull/559/commits/c1cd8e9b4bfad07f189fe52769f13399a9945b8e) is a perfectly valid description, and it references the issue too. Other pieces of info that you left in the commit bodies are very useful too, so feel free to concatenate the body commits as you squash them.

Finally, if you end up in a rebase/squash nightmare, I am fine with merging it as is or as a single commit :)

Honny1:
```
self = <podman.tests.integration.test_containers.ContainersIntegrationTest testMethod=test_container_rm_anonymous_volume>

        def test_container_rm_anonymous_volume(self):
            with self.subTest("Check anonymous volume is removed"):
                container_file = """
    FROM alpine
    VOLUME myvol
    ENV foo=bar
    """
                tmp_file = tempfile.mkstemp()
>               file = open(tmp_file, 'w')
E               TypeError: expected str, bytes or os.PathLike object, not tuple

podman/tests/integration/test_containers.py:193: TypeError
```

fulminemizzega:
> ```
> self = <podman.tests.integration.test_containers.ContainersIntegrationTest testMethod=test_container_rm_anonymous_volume>
> 
>         def test_container_rm_anonymous_volume(self):
>             with self.subTest("Check anonymous volume is removed"):
>                 container_file = """
>     FROM alpine
>     VOLUME myvol
>     ENV foo=bar
>     """
>                 tmp_file = tempfile.mkstemp()
> >               file = open(tmp_file, 'w')
> E               TypeError: expected str, bytes or os.PathLike object, not tuple
> 
> podman/tests/integration/test_containers.py:193: TypeError
> ```

The ugliest commit ever created, ignore all of this. Should never have existed.

fulminemizzega:
I have sorted a bit the mess, now everything is split in 3 commits, let me know if this is reasonable or if it should be squashed more or differently. Half of the tests failed to provision, I do not know if this is my fault or something else's.

fulminemizzega:
> /packit retest-failed

Can you give it another try? There is still one left

Issue #532, opened 2025-03-27
Published documentation is out-of-date


zmc:
I was in the process of filing a bug on the broken example in the published documentation, but the source file was correct! It does look like that [has been fixed](https://github.com/containers/podman-py/pull/405), but the production docs site appears to be claiming v4.9.0 is the "latest", so it looks like the site just hasn't been update in quite a while.

inknos:
@zmc that's right!

@jwhonce do you have access to readthedocs?

dklimpel:
The issue seems not resolved. The docs are currently out of date.

dklimpel:
This issue is fixed:
https://app.readthedocs.org/projects/podman-py/builds/29450204/

PR #562, opened 2025-06-22
fix: broken configuration for readthedocs

dklimpel:
Part of:
- #532 

Follow up to:
- #546 

Docs and example: https://docs.readthedocs.com/platform/stable/config-file/v2.html#packages

Current failure: https://app.readthedocs.org/projects/podman-py/builds/28555767/

```
Config file validation error Config validation error in python.install. Expected a list, got type dict ({'path': '.', 'method': 'pip', 'extra_requirements': ['docs']}).
```


inknos:
LGTM

Issue #564, opened 2025-08-05
podman-py doesn't support pull policy like the podman cli go binary does


mkanoor:
When calling the podman-py code to start a container we can't specify the pull policy. But when we run this via the podman cli it has a command line option https://docs.podman.io/en/v5.0.1/markdown/podman-run.1.html#pull-policy to specify the pull policy.
For parity we would like to have the same functionality be available in podman-py as in podman cli.

inknos:
@mkanoor please test the code in the linked pr if you can. I think the fix should be enough :D

PR #536, opened 2025-04-09
Set X-Registry-Auth header on manifest push and bump to new API

dklimpel:
- Allows the setting of an authentication header similar to the push of images (related to #164)
- The current api to push manifests has been deprecated since podman 4.0 ([`http://podman.io/libpod/manifests/{name}/push`](https://docs.podman.io/en/latest/_static/api.html#tag/manifests/operation/ManifestPushV3Libpod))
- Use the "new" API: [`http://podman.io/libpod/manifests/{name}/registry/{destination}`](https://docs.podman.io/en/latest/_static/api.html#tag/manifests/operation/ManifestPushLibpod)

I was already unable to use the old API with Podman 4.9

The current implementation for image push:

https://github.com/containers/podman-py/blob/a75e8c2b848d4f48af6f72c3170b5aa02bdff092/podman/domain/images_manager.py#L241-L268

PR #565, opened 2025-08-05
Implement policy option for image.pull

inknos:
Fixes: https://github.com/containers/podman-py/issues/564


podman/domain/images_manager.py
```diff
@@ -337,6 +337,7 @@ def pull(
             decode (bool) – Decode the JSON data from the server into dicts.
                 Only applies with ``stream=True``
             platform (str) – Platform in the format os[/arch[/variant]]
+            policy (str) - Pull policy. "always" (default), "missing", "never", "newer"
```
mkanoor, line 340: @inknos should this go into the container run instead of the image manager. the podman go cli uses it on the podman run command. Fetching images is akin to podman pull images

inknos: I see your point.

I only implemented `client.images.pull(policy=)` which defaults to "always" in the [`/libpod/images/pull`](https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImagePullLibpod) endpoint.

Now I noticed that the cli implements the policy on different commands.

- `podman run --pull-policy` with `missing` as default
- `podman build --pull` with `missing` as default

(different defaults come from [here](https://github.com/containers/podman/pull/26478))

I'd say that `podman-py` aims to provide API bindings, not cli. Therefore, would this be enough for your case?

```python
client.images.pull(..., policy=yourpolicy)
client.containers.run(...)
```

mkanoor: @inknos In the CLI workflow we have to do
- podman login.  (to specify registry credentials)
- podman run      (this can pull image if missing)

The podman-py is different in the sense it can be talking to a remote podman service and the podman login doesn't help, since podman login is creating the auth file locally.

So in the podman-py when we pull an image we have the option to pass in the [auth_config](https://github.com/containers/podman-py/blob/7e834d3cbee16b18628c1cf35ae879bd7116a9a4/podman/domain/images_manager.py#L332) with the username and password, which is an alternative to do the podman login.
The podman-run doesn't have the ability to pass in the auth_config, so there seems to be a disconnect, it would be nice if the podman-run can use the auth_config along with the pull policy. So we would need the pull policy to be passed into both image pull and the container run. If the container run gets passed in the auth_config if the image is missing it doesn't pass in the [auth_config](https://github.com/containers/podman-py/blob/7e834d3cbee16b18628c1cf35ae879bd7116a9a4/podman/domain/containers_run.py#L69) and if there are credentials needed it will fail. It expects that podman login has happened but that wont work across the API.

mkanoor: So given the current state we always pull the image first pass in the credentials via auth_config and then run the container create to overcome this issue.

inknos: fair enough, now I have the full picture. so now the `client.containers.run()` method passes the `policy` parameter to `containers.images.pull()`. The defaults are the same as the cli, "missing" for `run()` and "always" for `pull()`. What do you think?

podman/domain/containers_run.py
```diff
@@ -44,7 +44,10 @@ def run(
                 side. Default: False.
 
         Keyword Args:
-            - See the create() method for keyword arguments.
+            - These args are directly used to pull an image when the image is not found
```
mkanoor: @inknos can we pass in auth_config here so that the image pull can use it

inknos: oh, sure

inknos:
@mkanoor I updated the commit message, and I believe that's ready for your review and opinions again :)

inknos:
fwiw, the failing tests are related to an upstream change that will never get to downstream before podman6 and it's constantly causing issues to tests that use nightly. should be skipped though, but I'll take a look

inknos:
tests are fixed in https://github.com/containers/podman-py/pull/567

podman/domain/containers_run.py
```diff
@@ -44,7 +44,14 @@ def run(
...
```
TomSweeneyRedHat: nit
```suggestion
            - These args are directly used to pull an image when the image is not found.
```

podman/domain/containers_run.py
```diff
@@ -44,7 +44,14 @@ def run(
...
+                auth_config (Mapping[str, str]): Override the credentials that are found in the
+                config for this request. auth_config should contain the username and password
+                keys to be valid.
+                platform (str): Platform in the format os[/arch[/variant]]
+                policy (str): Pull policy. "missing" (default), "always", "never", "newer"
```
TomSweeneyRedHat, line 52: Per https://docs.podman.io/en/latest/markdown/podman-pull.1.html#policy, `always` is default.  Is the man page wrong, or are you doing something different in Python?

Alex-Izquierdo: For podman run the default is different. 
https://docs.podman.io/en/v5.0.1/markdown/podman-run.1.html#pull-policy

inknos: > I only implemented client.images.pull(policy=) which defaults to "always" in the [/libpod/images/pull](https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImagePullLibpod) endpoint.
>
> Now I noticed that the cli implements the policy on different commands.
>
>    podman run --pull-policy with missing as default
>    podman build --pull with missing as default

Quoting myself in https://github.com/containers/podman-py/pull/565#discussion_r2256748355

there is a difference in how the pull policy is handled if it's run from `image pull` or `run --pull-policy`. the docs are correct and `client.images.pull` has policy=always as default

TomSweeneyRedHat:
LGTM
once the tests are hip

podman/domain/images_manager.py
```diff
@@ -365,6 +366,7 @@ def pull(
         }
```
hsong-rh, line 366: When I tested to pull images with `auth_config`, it failed with:
```
   image = self.client.images.pull(
             ^^^^^^^^^^^^^^^^^^^^^^^^
   File "/app/venv/lib64/python3.11/site-packages/podman/domain/images_manager.py", line 365, in pull
    "X-Registry-Auth": api.encode_auth_header(auth_config) if auth_config else ""
                       ^^^^^^^^^^^^^^^^^^^^^^
 AttributeError: module 'podman.api' has no attribute 'encode_auth_header'
```

hsong-rh: The previous version works: https://github.com/containers/podman-py/pull/536/files#diff-85825cbd525470577e722fecdb9188caca30be38075e9f689358c63fb8b39595

PR #567, opened 2025-08-07
Add os_release() to conditionally run some tests

inknos:


podman/tests/utils.py
```diff
@@ -0,0 +1,13 @@
+import re
+import pathlib
+import csv
+
+
+def get_os_release() -> dict:
+    path = pathlib.Path("/etc/os-release")
```
Honny1: Why didn't you use the [`platform`](https://docs.python.org/3/library/platform.html#module-platform) library?

inknos: I thought you couldn't get the os_release with it but with the docs now I see it, let me change it

Honny1: `platform.freedesktop_os_release()`

TomSweeneyRedHat:
Once @Honny1 's comment is addressed
LGTM

podman/tests/utils.py
inknos: thanks @Honny1, I changed the tests using `platform`

Honny1 (approved):
LGTM

inknos:
whops, looks like `platform.freedesktop_os_release` was added in `3.10` and we have to run on much earlier. @Honny1 I'd go back to the previous method. wdyt?

(edit)

I reverted the old behavior partially, we'll try to use `platform` when available and use a fallback code in other cases. the goal is to drop the fallback code in the future.

Honny1:
> I reverted the old behavior partially, we'll try to use `platform` when available and use a fallback code in other cases. the goal is to drop the fallback code in the future.

I think it's a good solution.

podman/tests/utils.py
```diff
@@ -0,0 +1,18 @@
...
+        """This is a fallback for platforms that don't have the freedesktop_os_release function.
+        Python < 3.10
+        """
+        path = pathlib.Path("/etc/os-release")
+        with open(path, "r") as f:
+            reader = csv.reader(f, delimiter="=")
+            return dict(reader)
+
+
+os_release = freedesktop_os_release()
```
Honny1: This appears to be a constant, so it should be in uppercase to follow PEP 8. https://peps.python.org/pep-0008/#constants

inknos: Addressed. As a note, the linter does not catch [`C0103`](https://pylint.readthedocs.io/en/latest/user_guide/messages/convention/invalid-name.html) or `N815`, so we might be doing slightly wrong in other instances too.

Honny1: The version of the ruff linter in the tox file is 0.8, but the latest is 0.12. That could be the issue.

inknos: I believe it's not even [upstream](https://github.com/astral-sh/ruff/issues/2964)

version is 0.8 same as our pre-commit. I'll update it in a subsequent PR since I don't know how many things might break.

inknos: let's see, I hit no problem locally. https://github.com/containers/podman-py/pull/569

Honny1: I think I know what the problem is—it's disabled.. https://github.com/containers/podman-py/blob/3c624222bc951fde816b3d360353e56adbf9cb68/pyproject.toml#L129

inknos: yup, that rule includes `N80...`. I tested it against the uppercase for constants. it didn't catch `os_release`, but it caught `tearUp`, instead, and I would like not to touch that function

inknos: technically, I could plan to fix all the `N`s and enable the rule, just to be sure that once it's implemented it will start to lint

Luap99 (approved):
/lgtm

I think this is the wrong approach long term though, you don't want to skip on the OS unless the feature is actually being dictated by the OS. In this case all these changes seem to be caused by the podman version so skipping on that seem most logical.

inknos:
> I think this is the wrong approach long term though, you don't want to skip on the OS unless the feature is actually being dictated by the OS. In this case all these changes seem to be caused by the podman version so skipping on that seem most logical.

Fair point. To give context, in the prev team, I remember we used skipping on version in some tests, and it led to issues with testing coprs (or something, I don't remember exactly what was the issue, but it was annoying). so I am burned by the amount of failures I had to fix and in the end skip on distro (for fedora, centos, rhel) was better. So then, yeah, the solution is arbitrary and might not be optimal

PR #569, opened 2025-08-13
Update Ruff to 0.12.8

inknos:
cc: @Honny1

### DIFF
--- a/podman/api/client.py
+++ b/podman/api/client.py
@@ -18,7 +18,7 @@ from requests.adapters import HTTPAdapter
 from podman.api.api_versions import VERSION, COMPATIBLE_VERSION
 from podman.api.ssh import SSHAdapter
 from podman.api.uds import UDSAdapter
-from podman.errors import APIError, NotFound
+from podman.errors import APIError, NotFound, PodmanError
 from podman.tlsconfig import TLSConfig
 from podman.version import __version__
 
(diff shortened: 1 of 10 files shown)


Honny1:
I'm not sure if the errors are caused by the Ruff update. Anyway, what do you think about enabling linters to run in a GitHub Action? I think it would be faster than running everything in Packit. I can set it up.

inknos:
> I'm not sure if the errors are caused by the Ruff update.

the errors are fixed in 567

> Anyway, what do you think about enabling linters to run in a GitHub Action? I think it would be faster than running everything in Packit. I can set it up.

linters are already in a github.com/containers/podman-py/blob/main/.github/workflows/pre-commit.yml run by pre-commit. the packit action runs pre-commit on the whole codebase in a tox environment.

Honny1:
> > Anyway, what do you think about enabling linters to run in a GitHub Action? I think it would be faster than running everything in Packit. I can set it up.
> 
> linters are already in a github.com/containers/podman-py/blob/main/.github/workflows/pre-commit.yml run by pre-commit. the packit action runs pre-commit on the whole codebase in a tox environment.

Thanks for the explanation.

inknos:
I just did a quick check and a lot of ruff issues were fixed with normal linting with time, so I removed a bunch of them and fixed the few issues in the code.

There's one to not in the tests which was a mistyped `tearDown` function, it was called `tearUp`. The linter caught it right away but it was there since 2021. I decided to include it in this linting PR + ruff upgrade

Honny1 (approved):
Good job, LGTM

Honny1 (approved):
/LGTM

Issue #571, opened 2025-08-20
Implement skip on version in tests


inknos:
Some tests rely on podman features that might/might not be released prior to some version. A recent PR was implemented to skip tests based on distro, but the [skip on version approach is worth considering](https://github.com/containers/podman-py/pull/567#pullrequestreview-3136560942) and possibly be beneficial.

PR #572, opened 2025-08-21
tests: Fix tests to reflect removal of rw as default option

ricardobranco777:
Behaviour changed upstream with https://github.com/containers/podman/pull/25942

Verification run: https://openqa.opensuse.org/tests/5261848

Fixes:

```
# Test messages # test_container_mounts
# failure: 

AssertionError: 'size=456k,rprivate,nosuid,nodev,tmpcopyup' != 'size=456k,rw,rprivate,nosuid,nodev,tmpcopyup'
- size=456k,rprivate,nosuid,nodev,tmpcopyup
+ size=456k,rw,rprivate,nosuid,nodev,tmpcopyup
?          +++
self = <podman.tests.integration.test_container_create.ContainersIntegrationTest testMethod=test_container_mounts>

    def test_container_mounts(self):
        """Test passing mounts"""
        with self.subTest("Check bind mount"):
            mount = {
                "type": "bind",
                "source": "/etc/hosts",
                "target": "/test",
                "read_only": True,
                "relabel": "Z",
            }
            container = self.client.containers.create(
                self.alpine_image, command=["cat", "/test"], mounts=[mount]
            )
            self.containers.append(container)
            self.assertIn(
                f"{mount['source']}:{mount['target']}:ro,Z,rprivate,rbind",
                container.attrs.get('HostConfig', {}).get('Binds', list()),
            )
    
            # check if container can be started and exits with EC == 0
            container.start()
            container.wait()
    
            self.assertEqual(container.attrs.get('State', dict()).get('ExitCode', 256), 0)
    
        with self.subTest("Check tmpfs mount"):
            mount = {"type": "tmpfs", "source": "tmpfs", "target": "/test", "size": "456k"}
            container = self.client.containers.create(
                self.alpine_image, command=["df", "-h"], mounts=[mount]
            )
            self.containers.append(container)
>           self.assertEqual(
                container.attrs.get('HostConfig', {}).get('Tmpfs', {}).get(mount['target']),
                f"size={mount['size']},rw,rprivate,nosuid,nodev,tmpcopyup",
            )
E           AssertionError: 'size=456k,rprivate,nosuid,nodev,tmpcopyup' != 'size=456k,rw,rprivate,nosuid,nodev,tmpcopyup'
E           - size=456k,rprivate,nosuid,nodev,tmpcopyup
E           + size=456k,rw,rprivate,nosuid,nodev,tmpcopyup
E           ?          +++

podman/tests/integration/test_container_create.py:274: AssertionError
```

### DIFF
--- a/podman/tests/integration/test_container_create.py
+++ b/podman/tests/integration/test_container_create.py
@@ -316,7 +316,7 @@ class ContainersIntegrationTest(base.IntegrationTest):
             self.containers.append(container)
             self.assertEqual(
                 container.attrs.get('HostConfig', {}).get('Tmpfs', {}).get(mount['target']),
-                f"size={mount['size']},rw,rprivate,nosuid,nodev,tmpcopyup",
+                f"size={mount['size']},rprivate,nosuid,nodev,tmpcopyup",
             )
 
             container.start()


inknos:
Hey, @ricardobranco777 
Thanks for the PR, I'd like to fix the tests with [this approach](https://github.com/containers/podman-py/issues/571), would this be a good fix for you?

ricardobranco777:
> Hey, @ricardobranco777 Thanks for the PR, I'd like to fix the tests with [this approach](https://github.com/containers/podman-py/issues/571), would this be a good fix for you?

I believe it's not the upstream's job to do this given the sheer amount of different distros out there. I think downstream should backport and patch tests however it suits them and that's the approach I'm taking for testing on openSUSE & SUSE.

Patching based on podman version is also not perfect because the podman patch was backported.

Yu're the maintainer, so your call. :)

inknos:
Right, we have the option already for filtering on `/etc/os-release` (which is also not perfect) but should fit the case for backporting and might help with downstream patches too.

Recently, we discussed if we should filter tests based on version or distro, and I'd prefer to filter based on upstream versioning of `podman` and `podman-py`.

inknos:
hey @ricardobranco777 , [this](https://github.com/containers/podman-py/pull/567) is making tests pass. would you be able to use this fix and `pytest.mark.skipif` with `OS_RELEASE` for suse?

I am closing this PR since it does not apply to upstream, but I am happy to accept a new PR filter the tests for suse

ricardobranco777:
> hey @ricardobranco777 , [this](https://github.com/containers/podman-py/pull/567) is making tests pass. would you be able to use this fix and `pytest.mark.skipif` with `OS_RELEASE` for suse?
> 
> I am closing this PR since it does not apply to upstream, but I am happy to accept a new PR filter the tests for suse

That PR just skips the tests and we don't want to skip any tests. The root cause continues to be the change made by podman in https://github.com/containers/podman/pull/25942

This PR was about podman-py catching up to that change.  I don't understand the current approach at all.

ricardobranco777:
> Recently, we discussed if we should filter tests based on version or distro, and I'd prefer to filter based on upstream versioning of `podman` and `podman-py`.

This is the way to go... Why skip tests based on distro version?

inknos:
> That PR just skips the tests and we don't want to skip any tests. The root cause continues to be the change made by podman in https://github.com/containers/podman/pull/25942

The test is run for the old behavior **and** the new behavior. see [test_container_create.py in this commit](https://github.com/containers/podman-py/commit/f3b8f1d9825cbcdc5dba6c8a7157fbd9223a8624)

`test_container_mounts_without_rw_as_default` is a clone of `test_container_mounts` but tests the new option. therefore I believe we are testing everything.

the tests are simply skipped based on the podman version installed, so the correct behavior is always tested.

ricardobranco777:
> the tests are simply skipped based on the podman version installed, so the correct behavior is always tested.

Ok.  That makes sense.

The podman PR was backported to older podman versions in SLES though, so the podman version ain't perfect.  But podman-py is not packaged for SLES anyway.  It's only available on openSUSE Tumbleweed.  We'll use this PR until next version comes.  It's used in openQA tests only and not shipped in the RPM.

Thanks for your patience.

inknos:
> The podman PR was backported to older podman versions in SLES though, so the podman version ain't perfect. But podman-py is not packaged for SLES anyway. It's only available on openSUSE Tumbleweed. We'll use this PR until next version comes. It's used in openQA tests only and not shipped in the RPM.

I see, I was not aware if podman-py is packaged or how. thanks for the explanation.

If more changes need to be done upstream to fit SUSE needs, don't hesitate to reach out, and thanks to you for your contributions 🎉

PR #573, opened 2025-08-22
fixed reading status from self.attrs

sekyHC:
This is a fix for the following error:
```
  File "/usr/local/lib/python3.11/dist-packages/podman/domain/containers.py", line 61, in status
    return self.attrs["State"]["Status"]
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
TypeError: string indices must be integers, not 'str'
```

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -9,7 +9,7 @@ jobs:
     env:
       SKIP: no-commit-to-branch
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
     - uses: actions/setup-python@v5
       with:
         python-version: |
--- a/.github/workflows/publish-to-test-pypi.yml
+++ b/.github/workflows/publish-to-test-pypi.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
@@ -46,7 +46,7 @@ jobs:
 
     steps:
       - name: Download all the dists
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v5
         with:
           name: python-package-distributions
           path: dist/
@@ -68,7 +68,7 @@ jobs:
 
     steps:
       - name: Download all the dists
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v5
         with:
           name: python-package-distributions
           path: dist/
@@ -114,7 +114,7 @@ jobs:
 
     steps:
       - name: Download all the dists
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@v5
         with:
           name: python-package-distributions
           path: dist/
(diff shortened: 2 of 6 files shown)


sekyHC:
As far as I understood from the openshift-ci bot workflow I have to tag reviewers and approvers in a comment:
@ashley-cui 
@baude 
@inknos

inknos:
@sekyHC , thanks for the PR.

A bunch of changes were made recently that improved testing a little bit. Would you mind rebasing your branch? That should filter out few test errors.

The PR is very simple and looks good by itself, but we need to be sure that nothing breaks

sekyHC:
@inknos, only errors I can see in the logs of failed tests are:
1. can't find module tomli
2. changelog entry has to start with *

inknos:
mmh, for every `testing-farm` looks like I am getting this

```
        with self.subTest("Run Container"):
            top_ctnr = self.client.containers.run(
                self.alpine_image, "/usr/bin/top", name="TestRunPs", detach=True
            )
>           self.assertEqual(top_ctnr.status, "running")
E           AssertionError: {'OciVersion': '1.2.1', 'Status': 'runnin[425 chars]00Z'} != 'running'

podman/tests/integration/test_containers.py:125: AssertionError
```

let me investigate

inknos:
I confirm there's a failing test. can you reproduce it locally with `tox -e py -- -k test_container_crud`?

inknos:
yup, looks like `self.attrs["State"]["Status"]` gets just the status, ie, `running`, and you are returning the whole `"State"` structure. so tests are right to fail

If you can give me a reproducer of how you got the error in the PR description I can help you with the investigation

sekyHC:
```python
    async def list_containers(self, request):
        def sync_list():
            with PodmanClient(base_url="unix:///tmp/podman.sock") as client:
                result = []
                for c in client.containers.list(all=True):
                    result.append({
                        "Id": c.id,
                        "Name": c.name,          # container name
                        "Image": c.image.tags[0],
                        "Status": c.status,   # podman-py gives status directly
                    })
                return result

        containers = await asyncio.to_thread(sync_list)
        return web.json_response(containers)
```

This was giving me the error:

```
  File "/usr/local/lib/python3.11/dist-packages/podman/domain/containers.py", line 61, in status
    return self.attrs["State"]["Status"]
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
TypeError: string indices must be integers, not 'str'
```

in order to fix it and actually get status, I had to delete ["Status"] in the library, like I did in this PR

inknos:
Aha! then I know!

- here's the issue you are having:
  https://github.com/containers/podman-py/issues/446

- and here's the fix that will be released in the next versoin and merged just few mins ago:
  https://github.com/containers/podman-py/pull/540
  
Now, you have 2 options:

1. know the usage of list, and call reload manually as per the README. https://github.com/containers/podman-py/blob/main/README.md?plain=1#L55-L61
2. use the upstream version (or the one I am planning to releaes today or tomorrow, 5.6.0) and use `list(sparse=False)` which will reload under the hood.

to add a little bit of context, the state of containers is not retrieved automatically on each `list` call because the `podman inspect` call (which is what gets you the correct status) is expensive, so by default we don't do that.

hope that helps :)

PR #578, opened 2025-09-01
Skip tests conditionally based on version

inknos:
Fixes: https://github.com/containers/podman-py/issues/571

### DIFF
--- a/podman/tests/utils.py
+++ b/podman/tests/utils.py
@@ -1,5 +1,7 @@
 import pathlib
 import csv
+import re
+import subprocess
 
 try:
     from platform import freedesktop_os_release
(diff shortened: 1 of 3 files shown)


inknos:
Hey, @Luap99 and @Honny1 . based on the comments of https://github.com/containers/podman-py/pull/567#pullrequestreview-3136560942 I decided to open this PR to skip based on Podman's version. I decided to leave the os_release code because it could help testing downstream backports sometimes.

what do you think?

podman/tests/utils.py
```diff
@@ -15,4 +17,14 @@ def freedesktop_os_release() -> dict[str, str]:
             return dict(reader)
 
 
+def podman_version() -> tuple[int, ...]:
+    process = subprocess.Popen(
```
Honny1: I would use `with subprocess.Popen(...) as process:` because without the `with` statement, you are responsible for cleanup yourself.

Honny1 (commented):
LGTM, I have one minor suggestion.

podman/tests/utils.py
inknos: yup, I wasn't sure that `subprocess.Popen` supported context manager for the versions we support, but it's since `py3.2` so we are good. thanks

Honny1 (approved):
LGTM

baude:
i wasnt sure if this repo blocks on testing-farm so i added a hold label while that completes.

inknos:
> i wasnt sure if this repo blocks on testing-farm so i added a hold label while that completes.

actually good point, I need to update GH to block on fedora 43. currently it only blocks up to f42

PR #581, opened 2025-09-04
Bump release to 5.6.0

inknos:


inknos:
@Honny1 I believe the job yellow icons are stuck, because copr reports [all rpms are built](https://copr.fedorainfracloud.org/coprs/packit/containers-podman-py-581/build/9525044/). tests also run and passed, which depend on rpms, so we can move on I guess

Honny1:
Sure, pack it is not having good day. 
/LGTM

PR #582, opened 2025-09-05
Remove GH release from pypi workflow

inknos:

The workflow now mirrors the one from packaging.python.org page

https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#the-whole-ci-cd-workflow


inknos:
Closing in favor of https://github.com/containers/podman-py/pull/583

Creating a PR from GH interface is **very** confusing and makes my workflow useless when I want to update a PR. hence, I re-created it from my fork

PR #583, opened 2025-09-05
Remove GH release from pypi workflow

inknos:

Overall, the workflow mirrors the one from packaging.python.org page but with few differences:
  - "if" condition in publish-to-test-pypi, to run on main branch only
  - skip-existing test.pypi releases is true to try and reduce the pypi usage

https://packaging.python.org/en/latest/guides/
  publishing-package-distribution-releases-using-
  github-actions-ci-cd-workflows/#the-whole-ci-cd-workflow

### DIFF
--- a/.github/actions/build-python-package/action.yml
+++ b/.github/actions/build-python-package/action.yml
@@ -1,0 +1,54 @@
+name: 'Build Python Package'
+description: 'Build Python distribution packages (wheel and source tarball)'
+inputs:
+  python-version:
+    description: 'Python version to use for building'
+    required: false
+    default: '3.x'
+  upload-artifact:
+    description: 'Whether to upload the built packages as artifacts'
+    required: false
+    default: 'true'
+  artifact-name:
+    description: 'Name for the uploaded artifact'
+    required: false
+    default: 'python-package-distributions'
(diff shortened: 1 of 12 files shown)


inknos:
looks like packit is sleeping today, therefore /hold

inknos:
> Also, I am not sure if signing the Python package is required.

Quoting from [here](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#publishing-the-distribution-to-pypi) in the tips box

> Starting with version [v1.11.0](https://github.com/pypa/gh-action-pypi-publish/releases/tag/v1.11.0), [pypa/gh-action-pypi-publish](https://github.com/pypa/gh-action-pypi-publish) generates and uploads [PEP 740](https://peps.python.org/pep-0740/)-compatible attestations for each distribution by default. No additional manual signing steps are required.

we are good with signing

inknos:
I am ok to split the files, I'll work on it later

inknos:
I opened this PR with GH code editor and frankly, it made a mess creating branches. Closing in favor of https://github.com/containers/podman-py/pull/590

PR #586, opened 2025-09-08
Include test files in the source distribution

### LINKED ISSUE #585: Please include tests in PyPI source distribution
We would like to use the PyPI source distribution for downstream packaging in Gentoo to benefit from the signed attestations available there. However, we also need to be able to run the test suite as part of packaging. Could you please consider including the test files in the sdist? I can submit a PR if you agree.

See: https://packaging.python.org/en/latest/discussions/downstream-packaging/#provide-complete-source-distributions

mgorny:
Include the test files and `tox.ini` in the source distribution to facilitate downstream testing.

Fixes #585

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,2 @@
+recursive-include podman/tests *.py *.md
+include tox.ini


inknos:
@mgorny I rebased your branch since another PR went through. I'll approve it and hold for tests to pass.

/approve
/lgtm
/hold

mgorny:
Thank you!

PR #587, opened 2025-09-11
network.id should look for attrs "Id" and "id"...

### LINKED ISSUE #584: [BUG] Getting a network by ID does not work
See repro test [here](https://github.com/nlacasse/podman-py/commit/2cad8baf2859966276e177a6b1cc3a605cf9ba2a).

When we create a new network, podman returns it's `id`. However, the [`Network.id`](https://github.com/containers/podman-py/blob/main/podman/domain/networks.py#L31-L40) method tries to return `self.attrs["Id"]`, but that always fails, so it falls back to returing a synthetic `id` based on the network name. If we pass this `id` to `network.get()`, the call fails, since Podman does not know any networks with that id.

Lowercasing `Id` -> `id` in [this line](https://github.com/containers/podman-py/blob/main/podman/domain/networks.py#L34) fixes the issue, and causes `Network.id` to return the correct `id` assiged by Podman.  But I'm not sure if that's the correct fix.

nlacasse:
...before falling back to a synthetic id based on the hash of the network name.

Checking both is needed to support the Docker-compatible format ("Id") and Libpod-native format ("id").

This is analagous to how podman-py handles network.name -- we check attrs "Name" and "name" in that order.

Fixes #584

### DIFF
--- a/podman/domain/networks.py
+++ b/podman/domain/networks.py
@@ -30,8 +30,11 @@ class Network(PodmanResource):
     @property
     def id(self):  # pylint: disable=invalid-name
         """str: Returns the identifier of the network."""
-        with suppress(KeyError):
+        if "Id" in self.attrs:
             return self.attrs["Id"]
+
+        if "id" in self.attrs:
+            return self.attrs["id"]
 
         with suppress(KeyError):
             sha256 = hashlib.sha256(self.attrs["name"].encode("ascii"))
--- a/podman/tests/integration/test_networks.py
+++ b/podman/tests/integration/test_networks.py
@@ -64,6 +64,13 @@ class NetworksIntegrationTest(base.IntegrationTest):
             names = [i.name for i in nets]
             self.assertIn("integration_test", names)
 
+        with self.subTest("Get by ID"):
+            network = self.client.networks.get("integration_test")
+            net_id = network.id
+            assert isinstance(net_id, str)
+            network_by_id = self.client.networks.get(net_id)
+            self.assertEqual(network.name, network_by_id.name)
+
         with self.subTest("Delete network"):
             network = self.client.networks.get("integration_test")
             network.remove(force=True)


nlacasse:
Personally, I don't think we should ever generate a fake network id [like this](https://github.com/containers/podman-py/blob/0d88ef891638f3adf2bec6af7b54521bc8c43022/podman/domain/networks.py#L40-L41), but some tests depend on that behavior, so I left it in.

nlacasse:
> @nlacasse what about returning Id or id depending on the endpoint. the libpod api and docker's are essentially the same. they differ in the /libpod prefix. If this is a fix for the docker api only and breaks libpod behavior, I'd suggest to return it via the compatible endpoint

I'm not sure I understand the suggestion. Does the `Network` resource know whether is was created (or gotten) via a libpod endpoint?  Or is the idea that when we initialize a new `Network`, we normalize the `Id`->`id` attr based on the endpoint before initailizing?  That seems doable but might require a bit of plumbing.

Also, note that my change does not break docker (or any) behavior, it only fixes `Network.id` for libpod endpoints.  The approach I took is the same way we handle docker/libpod compatability in `Network.name`.  First we check for `attrs["Name"]` then `attrs["name"]`:
https://github.com/containers/podman-py/blob/main/podman/domain/networks.py#L54-L58

I think it's best to handle `Network.id` the same way we handle `Network.name`.  Maybe we can move forward with this PR, since it fixes a bug, and later clean up the upper/lowercase compatability for both `id` and `name` together?

podman/domain/networks.py
```diff
@@ -30,9 +30,12 @@ class Network(PodmanResource):
     @property
     def id(self):  # pylint: disable=invalid-name
         """str: Returns the identifier of the network."""
-        with suppress(KeyError):
+        if "Id" in self.attrs:
```
inknos, line 33: I'd stick with the old behavior and simply return `id`. that's what we want.

podman/tests/integration/test_networks.py
```diff
@@ -64,6 +64,13 @@ def test_network_crud(self):
             names = [i.name for i in nets]
             self.assertIn("integration_test", names)
 
+        with self.subTest("Get by ID"):
```
inknos, line 67: does this test actually fail if `id` or `Id`?

inknos:
hey, @nlacasse. Any update? do you need help or clarification from my side?

PR #588, opened 2025-09-18
Enable tests downstream via packit/tmt

inknos:
no effect upstream, changes will impact downstream only.

see here https://src.fedoraproject.org/rpms/python-podman/pull-request/52


jwhonce:
LGTM

inknos:
@Honny1 all good now

PR #590, opened 2025-09-24
Inknos update publish to pypy steps

inknos:

In the first commit, the workflow mirrors the one from packaging.python.org page but with few differences:

    "if" condition in publish-to-test-pypi, to run on main branch only
    skip-existing test.pypi releases is true to try and reduce the pypi usage

https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#the-whole-ci-cd-workflow

---

In the second commit, the build steps are split in an action and three separate workflows that should
- build on every PR (and don't upload artifacts as default)
- build and push to test pypi on every merge to main
- build and push to pypi on every tag push

Updates https://github.com/containers/podman-py/pull/583

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,18 @@
+name: Test Build Python distribution 📦
+
+on:
+  pull_request:
+
+jobs:
+  build:
+    name: Test Build Python distribution 📦
+    runs-on: ubuntu-latest
+    environment:
+      name: build
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Build Python package
+        uses: ./.github/actions/build-python-package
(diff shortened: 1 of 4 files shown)


inknos:
@Honny1 ptal :)

.github/workflows/build.yml
```diff
@@ -0,0 +1,18 @@
+name: Test Build Python distribution 📦
+
+on:
+  pull_request:
+
+jobs:
+  build:
+    name: Test Build Python distribution 📦
+    runs-on: ubuntu-latest
+    environment:
```
inknos, line 10: @Honny1 I added two lines here. `environment: name: build`. the reason why I did it is to enfore a new protection rule on dev environment targeting just the `build` job. not sure if I have to add it to the other `build` jobs in pypi and testpypi. wdyt?

image

Honny1: I think it is good.

PR #591, opened 2025-09-25
Update testing section

inknos:
Add common issues section
Add docs on how to skip tests based on podman's version and os


inknos:
> I would add an example of how to run tests with specific versions

Good suggestion, I reworked the docs a bit to reflect the changes you asked

Issue #574, opened 2025-08-22
Add support for podman volume import and export


Riushda:
[Podman 5.6.0](https://github.com/containers/podman/releases/tag/v5.6.0) introduces 2 new commands for the volumes: podman volume import and export. It would be good to add support for them in the sdk. This shouldn't be too complicated, I think this only requires adding both functions under `podman/domain/volumes.py.`

inknos:
@Riushda are you interested in working on that? if not I'll find some time to implement it

Riushda:
I already made a pull request ( reference in the this ticket history)

inknos:
> I already made a pull request ( reference in the this ticket history)

missed it, sorry. I'll check it now

PR #593, opened 2025-10-01
Add support for export and import commands

Riushda:
Podman 5.6.0 introduces 2 new commands: podman volume import and export. I was willing to add support for these 2 commands in the python sdk because they are very useful in some specific volumes backup scenario I have in mind. 

Closes https://github.com/containers/podman-py/issues/574


podman/domain/volumes.py
```diff
@@ -173,3 +173,33 @@ def remove(self, name: Union[Volume, str], force: Optional[bool] = None) -> None
             name = name.name
         response = self.client.delete(f"/volumes/{name}", params={"force": force})
         response.raise_for_status()
+
+    def export_archive(self, name: Union[Volume, str]) -> bytes:
```
inknos, line 178: why not just `def export/import`?

podman/domain/volumes.py
```diff
@@ -173,3 +173,33 @@ def remove(self, name: Union[Volume, str], force: Optional[bool] = None) -> None
...
+        Raises:
+            APIError: when service reports an error
+        """
+        if isinstance(name, Volume):
+            name = name.name
+        response = self.client.get(f"/volumes/{name}/export")
+        response.raise_for_status()
+        return response._content
+
+    def import_archive(self, name: Union[Volume, str], data: bytes):
```
inknos: I am thinking if it would be better to have `path: str` and `data: bytes` and make them exclusively required. It could be convenient to pass the path to a tarball. what do you think?

podman/domain/volumes.py
Riushda: Because import is a python reserved keyword and for consistency I also renamed the export. If I you have a better name I wouldn't mind changing them.

podman/domain/volumes.py
Riushda: I think this is a good idea, I will add this feature.

podman/domain/volumes.py
inknos: mmmh, correct, I thought it was ok to use it as a method. let's keep _archive

podman/domain/volumes.py
```diff
@@ -173,3 +174,49 @@ def remove(self, name: Union[Volume, str], force: Optional[bool] = None) -> None
...
+            APIError: when service reports an error
+        """
+        if isinstance(name, Volume):
+            name = name.name
+        response = self.client.get(f"/volumes/{name}/export")
+        response.raise_for_status()
+        return response._content
+
+    def import_archive(
+        self, name: Union[Volume, str], data: Optional[bytes] = None, path: Optional[str] = None
```
inknos, line 194: I was about to comment on this line about `|` before python 3.10, but you found it

inknos:
code looks good now! let me be pedantic, but could you add a test for `path` and `data` too? you can mock the API call in the unit tests, check `podman/tests/unit/test_volumesmanager.py`

podman/tests/unit/test_volumesmanager.py
```diff
@@ -137,6 +137,28 @@ def test_prune(self, mock):
...
+            self.client.volumes.import_archive("dbase", path="/path/to/archive.tar")
+
+    @requests_mock.Mocker()
+    def test_export(self, mock):
+        mock.get(
+            tests.LIBPOD_URL + "/volumes/dbase/export",
+            content=b'exported_archive',
+        )
+        actual = self.client.volumes.export_archive("dbase")
+        self.assertIsInstance(actual, bytes)
```
Riushda, line 160: To be honest, I'm not sure about the value of these mock tests (for the import and export). I wouldn't have added them if you didn't ask for it.

inknos: My motivation is primarily code coverage. In addition to this, tests help me a lot with future features working with the community, even when they are simple.

I'd like to hear your opinion: why you are not sure of the value for these specific tests. Do you mean they are too trivial?

thanks a lot for the patience and for updating the pr

Riushda: I was mainly talking about the functions I added. There isn't much to test without an actual volume and archive file. But don't get me wrong, I think mock tests in general are great to add more test coverage in particular for complex scenario that are not easy to reproduce in actual integration test.

inknos: I think it would be very nice to do something like this in python, but we are missing other options in the python bindings to do so

```bash
podman volume create testvol
podman volume export testvol --output testvol.tar
mkdir testvol
tar -xvf testvol.tar -C testvol/
touch  testvol/test
tar -rf testvol.tar testvol/test
cat testvol.tar | podman volume import testvol -
podman unshare
podman volume mount testvol
ls /home/nsella/.local/share/containers/storage/volumes/testvol/_data/testvol/test
```

So I think your tests are now good

Riushda:
Thanks for the review !

PR #596, opened 2025-10-06
Add direct support to `--manifest` when building container images.

### LINKED ISSUE #592: Add direct support to `--manifest` for building multi-architecture container images.
Hi,

I am using `podman‑py` to automate the creation of multi‑architecture container images with `podman`.

My existing CLI workflow looks like this:

```bash
podman manifest create example:v1.2.3
podman build --manifest example:v1.2.3 --platform linux/amd64,linux/arm .
```

When translating these commands into calls to `podman-py`, I've noticed that the [`--manifest` argument](https://docs.podman.io/en/latest/markdown/podman-build.1.html#manifest-manifest) is **not** supported by the [`client.images.build` function](https://github.com/containers/podman-py/blob/main/podman/domain/images_build.py#L171).

I am aware that `podman-py` already supports creating manifests and adding images to them.
However, I am left wondering if supporting the `--manifest` would not give developers an experience closer to the existing `podman` CLI and thus simplify the creation of scripts, as the creation of manifests and addition of images could all be done in a single call.

My proposal, for the same example CLI workflow above, is to allow something like this:

```python
# ...
  image, _ = client.images.build(
    path=".",
    dockerfile="Containerfile",
    manifest="example:v1.2.3",  # <-- here's the missing field - the manifest is created if it does not exist
    platform=["linux/amd64", "linux/arm64"],
  )
```

Here, the new optional `manifest` _kwarg_ would accept the name of the manifest and automatically add the newly built image(s) to it.

What are your opinions on this feature?

Thanks a lot for your attention and for the project!

MisterOwlPT:
**Closes #592.**

With the proposed changes it is now possible to create manifests and automatically add images to them _as part of the build process_ (equivalent to `podman build --manifest ...`). 

```python
# ...
  image, _ = client.images.build(
    path=".",
    dockerfile="Containerfile",
    manifest="example:v1.2.3",  # <-- this field is now supported 
    platform=["linux/amd64", "linux/arm64"],
  )
```

Running the previous code, all built images are automatically added to the provided manifest (which is also created if it does not exist already).

#### **Changes**:
- Mapped the `--manifest` argument into the expected `manifest` query parameter when making requests;
- Updated documentation for the `client.images.build` function as to refer to the newly support field. Description text extracted from the Podman CLI itself;
- Updated existing unit tests to ensure that `--manifest` argument is mapped into the expected query parameter;
- Implemented new integration test to ensure that a manifest is automatically created if required. A complementary cleanup mechanism was also implemented to ensure that manifests created during tests are deleted after their execution.

Looking forward to your feedback.

### DIFF
--- a/podman/domain/images_build.py
+++ b/podman/domain/images_build.py
@@ -66,6 +66,8 @@ class BuildMixin:
             layers (bool) - Cache intermediate layers during build.
             output (str) - specifies if any custom build output is selected for following build.
             outputformat (str) - The format of the output image's manifest and configuration data.
+            manifest (str) - add the image to the specified manifest list.
+                Creates manifest list if it does not exist.
 
         Returns:
             first item is the podman.domain.images.Image built
@@ -173,6 +175,7 @@ class BuildMixin:
             "forcerm": kwargs.get("forcerm"),
             "httpproxy": kwargs.get("http_proxy"),
             "networkmode": kwargs.get("network_mode"),
+            "manifest": kwargs.get("manifest"),
             "nocache": kwargs.get("nocache"),
             "platform": kwargs.get("platform"),
             "pull": kwargs.get("pull"),
--- a/podman/tests/integration/test_images.py
+++ b/podman/tests/integration/test_images.py
@@ -38,6 +38,12 @@ class ImagesIntegrationTest(base.IntegrationTest):
         self.client = PodmanClient(base_url=self.socket_uri)
         self.addCleanup(self.client.close)
 
+        self.test_manifest_name = "dummy:v1.2.3"
+
+    def tearDown(self) -> None:
+        if self.client.manifests.exists(self.test_manifest_name):
+            self.client.manifests.remove(self.test_manifest_name)
+
     def test_image_crud(self):
         """Test Image CRUD.
 
@@ -143,6 +149,17 @@ class ImagesIntegrationTest(base.IntegrationTest):
         image, stream = self.client.images.build(fileobj=buffer)
         self.assertIsNotNone(image)
         self.assertIsNotNone(image.id)
+
+    def test_build_with_manifest(self):
+        buffer = io.StringIO("""FROM quay.io/libpod/alpine_labels:latest""")
+
+        self.assertFalse(self.client.manifests.exists(self.test_manifest_name))
+
+        image, _ = self.client.images.build(fileobj=buffer, manifest=self.test_manifest_name)
+        self.assertIsNotNone(image)
+        self.assertIsNotNone(image.id)
+
+        self.assertTrue(self.client.manifests.exists(self.test_manifest_name))
 
     def test_build_with_context(self):
         context = io.BytesIO()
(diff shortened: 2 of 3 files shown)


jwhonce:
/approve

LGTM

MisterOwlPT:
Hi @jwhonce,

Some more points concerning this PR.

- I have fixed the existing linting errors;
- I've shortened the two long commit messages (to fit the 72 character limit).
- Removed accidental signature lines from commits I didn’t author and only the commits that _are_ actually mine are now signed by me.

Thanks for the review, and really really sorry for the extra noise!

Edit: typos

jwhonce:
@MisterOwlPT It appears the commit linter is still not happy with your work.

MisterOwlPT:
@jwhonce,
I've amended the troublesome commit messages.
Everything should be fine now.

Let me know if everything looks good to you.

inknos:
@MisterOwlPT your code is clean and linter is happy, nice job.

I believe you need to rebase to main and fix the commits, I see the commit history is mixed with merged history, that's not clean and can't be merged.

I tried to rebase locally, and this is what imho looks nicer in a single patch. what do you think?

[0001-support-flag-manifest-when-building-container-images.patch](https://github.com/user-attachments/files/22740117/0001-support-flag-manifest-when-building-container-images.patch)

MisterOwlPT:
@inknos,

> **I believe you need to rebase to main and fix the commits, I see the commit history is mixed with merged history, that's not clean and can't be merged.**

Yes, sincere apologies for that!
As suggested, I've `git rebase` to `main` and fixed the commits.
I believe this branch is now clean as intended. 

Let me know if something else is required.
Thanks once again for your patience!

podman/domain/images_build.py
```diff
@@ -173,6 +175,7 @@ def _render_params(kwargs) -> dict[str, list[Any]]:
             "forcerm": kwargs.get("forcerm"),
             "httpproxy": kwargs.get("http_proxy"),
             "networkmode": kwargs.get("network_mode"),
+            "manifest": kwargs.get("manifest"),
```
inknos, line 178: I am not sure if what I say makes 100% sense, but I believe you should consider avoiding `get("manifest")` and rather use

```python
manifest = kwargs.get("manifest")
if manifest is not None:
    params["manifest"] = manifest
```

when manifest is not passed to the function, you are feeding `manifest=None` to the api and I am not sure how it's handled

so we should also test that behavior for params that don't have a default

MisterOwlPT: > ... you are feeding `manifest=None to the api and I am not sure how it's handled

As per my understanding of the code, all parameters are handled [here](https://github.com/containers/podman-py/blob/main/podman/domain/images_build.py#L213).
In essence, if a parameter is `None`, then it won't be included in the result of `_render_params` and therefore will be "hidden" from `build` and thus not fed to the API at all.

Therefore I think the condition you propose is redundant 🤔 
Also, why don't all other fields also require that additional guard condition?

> so we should also test that behavior for params that don't have a default

I adopted the same parsing strategy as for all other supported fields as I do not see in the code where's the difference (concerning the final API calls) between supporting a parameter which is set to `None` and not supporting it at all.

Let me know what you think of my reasoning.

If required, I'll make the suggested changes and update the tests.
Thanks for the feedback!

inknos: yeah makes sense, there are discrepancies in other parts of the code and a warning flag raised in my head, but you are doing things right

inknos:
@MisterOwlPT all is good. let me just nitpick on the commits. we don't squash on merge, and the change looks nice on a single commit since it's all part of one added feature. would you do squash them, please? then I'll merge :D

MisterOwlPT:
@inknos,

>  would you do squash them, please?

Done!
I got a bit carried away with "little, incremental commits" 😅

Issue #595, opened 2025-10-04
PidmanClient().version() fails while `podman version` works


Ark-kun:
podman-py version: 5.6.0
OS: Mac OS X 15.7

```
% podman version
Client:        Podman Engine
Version:       5.5.0
API Version:   5.5.0
Go Version:    go1.24.3
Built:         Tue May 13 11:59:13 2025
Build Origin:  brew
OS/Arch:       darwin/arm64

Server:       Podman Engine
Version:      5.4.0
API Version:  5.4.0
Go Version:   go1.23.5
Built:        Mon Feb 10 16:00:00 2025
OS/Arch:      linux/arm64
```
```
% podman info -f "{{.Host.RemoteSocket.Path}}"
unix:///run/podman/podman.sock
(backend-test-python) ark@127 bac
```

```
% uv run --python 3.11 python -c 'import podman; print(podman.PodmanClient().version())'
Traceback (most recent call last):
  File "/Users/aaa/proj-backend/.venv/lib/python3.11/site-packages/podman/api/uds.py", line 41, in connect
    super().connect(netloc)
FileNotFoundError: [Errno 2] No such file or directory

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/aaa/proj-backend/.venv/lib/python3.11/site-packages/urllib3/connectionpool.py", line 787, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "/Users/aaa/proj-backend/.venv/lib/python3.11/site-packages/urllib3/connectionpool.py", line 493, in _make_request
    conn.request(
  File "/Users/aaa/proj-backend/.venv/lib/python3.11/site-packages/urllib3/connection.py", line 445, in request
    self.endheaders()
  File "/Users/ark/.local/share/uv/python/cpython-3.11.12-macos-aarch64-none/lib/python3.11/http/client.py", line 1298, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "/Users/ark/.local/share/uv/python/cpython-3.11.12-macos-aarch64-none/lib/python3.11/http/client.py", line 1058, in _send_output
    self.send(msg)
  File "/Users/ark/.local/share/uv/python/cpython-3.11.12-macos-aarch64-none/lib/python3.11/http/client.py", line 996, in send
    self.connect()
  File "/Users/aaa/proj-backend/.venv/lib/python3.11/site-packages/podman/api/uds.py", line 86, in connect
[...]

PR #597, opened 2025-10-07
fix: use active service connection when no explicit connection specified

Honny1:
- Check for configured active service before falling back to local socket
- Fixes connection issues on macOS with podman machine setups

Fixes: https://github.com/containers/podman-py/issues/595


Honny1:
cc @inknos @Ark-kun

inknos:
Thanks for investigating and fixing on Mac OS, @Honny1 

Since the reporter confirmed [the fix](https://github.com/containers/podman-py/issues/595#issuecomment-3378556429) we can get it in.

/approve
/lgtm

Issue #589, opened 2025-09-21
Incorrect and incomplete type annotations for container.run()


jonathanunderwood:
The type annotation for `container.run()` [here](https://github.com/containers/podman-py/blob/main/podman/domain/containers_run.py#L28) is `Union[Container, Union[Generator[str, None, None], Iterator[str]]]`. 

The docstring for the return type states:
```python
        Returns:
            - When detach is True, return a Container
            - If stdout is True, include stdout from container in output
            - If stderr is True, include stderr from container in output
            - When stream is True, output from container is returned as a generator
            - Otherwise, an iterator is returned after container has finished
```

So, when `stream` is `True`, according to the return type, one would expect `Generator[str, None, None]` as the return type. In fact, what is returned is `[Generator[bytes, None, None]`. I am not sure if this is a bug in the type annotation, or incorrect implementation, and decoding to a string is actually expected.

In addition, those conditional return types really should be encoded as a set of `overlad` annotations for that method - the current type hint is pretty useless as is.

inknos:
@jonathanunderwood good catch, it's wrongly annotated. it should be `bytes`. would you like to open a PR?

Ark-kun:
Also, the class Mount does not exist in Podman (exists in Docker).

dklimpel:
It is fixed by:
- #599

PR #599, opened 2025-10-10
fix: typing in `podman.domain.containers_run`

dklimpel:
fixes:
- #589

### DIFF
--- a/podman/domain/containers_run.py
+++ b/podman/domain/containers_run.py
@@ -25,7 +25,7 @@ class RunMixin:  # pylint: disable=too-few-public-methods
         stderr=False,
         remove: bool = False,
         **kwargs,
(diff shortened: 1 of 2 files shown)


vrothberg:
@jwhonce PTAL

jwhonce:
@dklimpel Should the Iterator also be updated to return bytes, as copilot suggests? The ignores are acceptable to me.

dklimpel:
> @dklimpel Should the Iterator also be updated to return bytes, as copilot suggests? The ignores are acceptable to me.

I would only fix this one issue. For a decision if the iterator is a str or bytes it needs a deep dive.

EDIT:
From a conceptual point of view, it makes sense to have standardized return types. `container.logs` also returns `bytes`. I will adjust it.

jwhonce:
LGTM

PR #603, opened 2025-11-03
Add support to `--secrets` when building container images

### LINKED ISSUE #526: [QUESTION] Pass `--secret` option when building an image
Hi,

Is it possible to pass [the `--secret` option](https://docs.podman.io/en/latest/markdown/podman-build.1.html#secret-id-id-src-envorfile-env-env-type-file-env) when building an image with the python client? It works with podman CLI with or without the `--remote` option and I can mount the secret when building the image.

gsilva00:
**Closes #526**.

With the proposed changes it is now possible to pass secrets into the build context in a way that they won't be stored in the final image (equivalent to `podman build --secret ...`). They are temporarily mounted during the build process, in Containerfile `RUN --mount=type=secret,id=<some_id>` instructions, where the `<some_id>` matches the id specified in the corresponding `--secret` option of the `podman build` command.

**Changes:**
- Mapped the `--secrets` argument into the expected manifest query parameter when making requests;
- Updated documentation for the `client.images.build` function to include the added field. Description text extracted from the Podman CLI itself (with some additional info about the possibility of passing env variables)
- Updated existing unit tests to ensure that `--secret` argument is mapped into the expected query parameter;
- Implemented new integration test to ensure that a secret file is correctly passed and used by the build context, as well as testing that the file does not remain in the built image;

### DIFF
--- a/podman/domain/images_build.py
+++ b/podman/domain/images_build.py
@@ -68,6 +68,7 @@ class BuildMixin:
             outputformat (str) - The format of the output image's manifest and configuration data.
             manifest (str) - add the image to the specified manifest list.
                 Creates manifest list if it does not exist.
+            secrets (list[str]) - Secret files/envs to expose to the build
 
         Returns:
             first item is the podman.domain.images.Image built
@@ -208,6 +209,9 @@ class BuildMixin:
             params["extrahosts"] = json.dumps(kwargs.get("extra_hosts"))
         if "labels" in kwargs:
             params["labels"] = json.dumps(kwargs.get("labels"))
+
+        if "secrets" in kwargs:
+            params["secrets"] = json.dumps(kwargs.get("secrets"))
 
         if params["dockerfile"] is None:
             params["dockerfile"] = f".containerfile.{random.getrandbits(160):x}"
--- a/podman/tests/integration/test_images.py
+++ b/podman/tests/integration/test_images.py
@@ -15,16 +15,17 @@
 """Images integration tests."""
 
 import io
+import os
 import platform
 import tarfile
+import tempfile
 import types
 import unittest
 
 import podman.tests.integration.base as base
 from podman import PodmanClient
 from podman.domain.images import Image
-from podman.errors import APIError, ImageNotFound, PodmanError
-
+from podman.errors import APIError, ContainerError, ImageNotFound, PodmanError
 
 # @unittest.skipIf(os.geteuid() != 0, 'Skipping, not running as root')
 
@@ -196,6 +197,41 @@ class ImagesIntegrationTest(base.IntegrationTest):
         )
         self.assertIsNotNone(image)
         self.assertIsNotNone(image.id)
+
+    def test_build_with_secret(self):
+        with tempfile.TemporaryDirectory() as context_dir:
+            dockerfile_path = os.path.join(context_dir, "Dockerfile")
+            with open(dockerfile_path, "w") as f:
+                f.write("""
+                FROM quay.io/libpod/alpine_labels:latest
+                RUN --mount=type=secret,id=example cat /run/secrets/example > /output.txt
+                """)
+
+            secret_path = os.path.join(context_dir, "build-secret.txt")
+            with open(secret_path, "w") as f:
+                f.write("secret123")
+
+            image, _ = self.client.images.build(
+                path=context_dir,
+                secrets=["id=example,src=build-secret.txt"],
+                dockerfile="Dockerfile",
+            )
+
+        self.assertIsNotNone(image)
+        self.assertIsNotNone(image.id)
+
+        # Verify secret was passed and stored in file (NOT RECOMMENDED for real use cases)
+        container_out = self.client.containers.run(
+            image.id, command=["cat", "/output.txt"], remove=True, log_config={"Type": "json-file"}
+        )
+        self.assertIn(b"secret123", container_out)
+
+        # Verify mounted secret file is not present in image
+        with self.assertRaises(ContainerError) as exc:
+            self.client.containers.run(
+                image.id, command=["cat", "/run/secrets/example"], remove=True
+            )
+        self.assertIn("No such file or directory", b"".join(exc.exception.stderr).decode("utf-8"))
 
     @unittest.skipIf(platform.architecture()[0] == "32bit", "no 32-bit image available")
     def test_pull_stream(self):
--- a/podman/tests/unit/test_build.py
+++ b/podman/tests/unit/test_build.py
@@ -67,7 +67,8 @@ class TestBuildCase(unittest.TestCase):
                 "&cpuperiod=10"
                 "&extrahosts=%7B%22database%22%3A+%22127.0.0.1%22%7D"
                 "&labels=%7B%22Unittest%22%3A+%22true%22%7D"
-                "&manifest=example%3Av1.2.3",
+                "&manifest=example%3Av1.2.3"
+                "&secrets=%5B%22id%3Dexample%2Csrc%3Dpodman-build-secret123%22%5D",
                 text=buffer.getvalue(),
             )
             mock.get(
@@ -100,6 +101,7 @@ class TestBuildCase(unittest.TestCase):
                 extra_hosts={"database": "127.0.0.1"},
                 labels={"Unittest": "true"},
                 manifest="example:v1.2.3",
+                secrets=["id=example,src=podman-build-secret123"],
             )
             self.assertIsInstance(image, Image)
             self.assertEqual(image.id, "032b8b2855fc")


gsilva00:
Hi,

Now that most pipeline checks have finished, I'm stumped on the results of the failing ones:

- The `pre-commit` check is strange. I installed its hooks and it passed locally (otherwise, I wouldn't have been able to commit), and I also ran it manually. The `test` attribute in `/tests/coverage_integration` also seems to be set, despite the error message.
- The integration test fails in not raising an exception when the secret file is not found. However, when I run it locally, it raises the exception as expected. Given that `ls /non_existent_file` should always fail and return a non-zero code, I'm not sure why it passes in the pipeline.

Looking forward to your feedback!

EDIT: I've also noticed that I left a temporary `assert` in the integration test, that I should've replaced with a proper `unittest` `assert`. However, I thought it best to wait on pipeline feedback to not clutter the PR with very small commits, nor trigger it multiple times.

inknos:
> The pre-commit check is strange. I installed its hooks and it passed locally (otherwise, I wouldn't have been able to commit), and I also ran it manually. The test attribute in /tests/coverage_integration also seems to be set, despite the error message.

Weird indeed. Locally, it's a green check for me

> The integration test fails in not raising an exception when the secret file is not found. However, when I run it locally, it raises the exception as expected. Given that ls /non_existent_file should always fail and return a non-zero code, I'm not sure why it passes in the pipeline.

this however fails, consistently with CI.

I'll take a quick look now to see if I catch the issue, and if not I'll come back tomorrow with a fresh mind

inknos:
update: 
```
podman run -it --rm -v $(pwd):/devel:Z fedora:43
cd /devel
dnf install -y pre-commit
pre-commit run -a
```
reproduces the behavior of github.

The python env for that action (3.14) follows latest fedora (43) and probably tmt run into some issues since it's not up to date. the version in `.pre-commit-config.yaml` [should be bumped to `1.60.0`](https://github.com/containers/podman-py/blob/main/.pre-commit-config.yaml#L19), and that fixes the issue

Edit: typo

gsilva00:
Hi @inknos ,

**About `ContainerError` not being raised for `ls /run/secrets`**

I ran the tests on the same OSes as the testing farms (not exactly their setup, but the error showed up nonetheless).

It happened because **the `/run/secrets` dir inside the built images isn't deleted after the image is built**, so the `ls` command succeeds.
That's different from the behavior on my host. On my host, that folder is deleted, so the `ls` command fails as expected.
Still, the secret file is not present, so **altering the test to check for the file presence instead of the folder presence worked on all tested OSes.** It also makes the code more meaningful, as the important part is the file, not necessarily the folder.

I don't understand why this happens:

- Both the host and the testing farms are using the same Podman API endpoint version (the default - `v5.6.0`).
- I would think this difference in build behavior is the responsability of `buildah`, which AFAIK is the one handling the builds under the hood. So I went looking for differences related to this:
  - Some testing farms have older `podman` versions (`v5.6.0`) than my host (`v5.6.2`).
  - The behaviour could be different thanks to that (given that the API calls use the underlying host's installation, as the tests are starting a `podman` service connecting on a local unix socket, AFAIK).
  - I checked the `podman` releases and their respective `buildah` releases (between `v5.6.0` and `v5.6.2`), **but I couldn't find any mention of build-time secret mounts or changes to handling of temporary files/folders during builds.**
  - However, other testing farms have a higher `podman` versions and it still fails, so I doesn't seem to be related to that.

I didn't explore further than this, but would like to understand the root cause of this issue better.

In any case, the test should now pass.

gsilva00:
> LGTM. Please rebase onto main, squash your commits, and remove the separate CI fix commit. The issue should be fixed now. Thanks.

Done.
Missed the squashing the first time, my bad.

Honny1:
@inknos, can you do the final review?

ricardobranco777:
This is failing with podman 5.8.0 on openSUSE Tumbleweed on arm64 only:

- runc: https://openqa.opensuse.org/tests/5740044
- runc (rootless): https://openqa.opensuse.org/tests/5740045
- crun: https://openqa.opensuse.org/tests/5740050
- crun (rootless): https://openqa.opensuse.org/tests/5740049

```
# Test messages # test_build_with_secret
# failure: 

podman.errors.exceptions.BuildError: building at STEP "RUN --mount=type=secret,id=example cat /run/secrets/example > /output.txt": while running runtime: exit status 255
self = <podman.tests.integration.test_images.ImagesIntegrationTest testMethod=test_build_with_secret>

    def test_build_with_secret(self):
        with tempfile.TemporaryDirectory() as context_dir:
            dockerfile_path = os.path.join(context_dir, "Dockerfile")
            with open(dockerfile_path, "w") as f:
                f.write("""
                FROM quay.io/libpod/alpine_labels:latest
                RUN --mount=type=secret,id=example cat /run/secrets/example > /output.txt
                """)
    
            secret_path = os.path.join(context_dir, "build-secret.txt")
            with open(secret_path, "w") as f:
                f.write("secret123")
    
>           image, _ = self.client.images.build(
                path=context_dir,
                secrets=["id=example,src=build-secret.txt"],
                dockerfile="Dockerfile",
            )

podman/tests/integration/test_images.py:252: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <podman.domain.images_manager.ImagesManager object at 0xffff7f3e3650>
kwargs = {'dockerfile': 'Dockerfile', 'path': '/tmp/tmpxz4i7fst', 'secrets': ['id=example,src=build-secret.txt']}
params = {'dockerfile': 'Dockerfile', 'layers': True, 'outputformat': 'application/vnd.oci.image.manifest.v1+json', 'secrets': '["id=example,src=build-secret.txt"]'}
body = <_io.BufferedReader name='/tmp/podman_context1nh3pkwi.tar'>, path = None
filename = PosixPath('/tmp/tmpxz4i7fst/Dockerfile'), excludes = []
post_kwargs = {}

    def build(self, **kwargs) -> tuple[Image, Iterator[bytes]]:
        """Returns built image.
    
        Keyword Args:
            path (str) – Path to the directory containing the Dockerfile
            fileobj – A file object to use as the Dockerfile. (Or an IO object)
            tag (str) – A tag to add to the final image
            quiet (bool) – Whether to return the status
            nocache (bool) – Don’t use the cache when set to True
            rm (bool) – Remove intermediate containers. Default True
            timeout (int) – HTTP timeout
            custom_context (bool) – Optional if using fileobj
            encoding (str) – The encoding for a stream. Set to gzip for compressing (ignored)
            pull (bool) – Downloads any updates to the FROM image in Dockerfile
            forcerm (bool) – Always remove intermediate containers, even after unsuccessful builds
            dockerfile (str) – full path to the Dockerfile / Containerfile
            buildargs (Mapping[str,str) – A dictionary of build arguments
            container_limits (dict[str, Union[int,str]]) –
                A dictionary of limits applied to each container created by the build process.
                    Valid keys:
    
                    - memory (int): set memory limit for build
                    - memswap (int): Total memory (memory + swap), -1 to disable swap
                    - cpushares (int): CPU shares (relative weight)
                    - cpusetcpus (str): CPUs in which to allow execution, For example, "0-3", "0,1"
                    - cpuperiod (int): CPU CFS (Completely Fair Scheduler) period (Podman only)
                    - cpuquota (int): CPU CFS (Completely Fair Scheduler) quota (Podman only)
            shmsize (int) – Size of /dev/shm in bytes. The size must be greater than 0.
                If omitted the system uses 64MB
            labels (Mapping[str,str]) – A dictionary of labels to set on the image
            cache_from (list[str]) – A list of image's identifier used for build cache resolution
            target (str) – Name of the build-stage to build in a multi-stage Dockerfile
            network_mode (str) – networking mode for the run commands during build
            squash (bool) – Squash the resulting images layers into a single layer.
            extra_hosts (dict[str,str]) – Extra hosts to add to /etc/hosts in building
                containers, as a mapping of hostname to IP address.
            platform (str) – Platform in the format os[/arch[/variant]].
            isolation (str) – Isolation technology used during build. (ignored)
            use_config_proxy (bool) – (ignored)
            http_proxy (bool) - Inject http proxy environment variables into container (Podman only)
            layers (bool) - Cache intermediate layers during build. Default True.
            output (str) - specifies if any custom build output is selected for following build.
            outputformat (str) - The format of the output image's manifest and configuration data.
                Default to "application/vnd.oci.image.manifest.v1+json" (OCI format).
            manifest (str) - add the image to the specified manifest list.
                Creates manifest list if it does not exist.
            secrets (list[str]) - Secret files/envs to expose to the build
    
        Returns:
            first item is the podman.domain.images.Image built
    
            second item is the build logs
    
        Raises:
            BuildError: when there is an error during the build
            APIError: when service returns an error
            TypeError: when neither path nor fileobj is not specified
        """
    
        params = self._render_params(kwargs)
    
        body = None
        path = None
        if kwargs.get("custom_context"):
            if "fileobj" not in kwargs:
                raise PodmanError(
                    "Custom context requires fileobj to be set to a binary file-like object "
                    "containing a build-directory tarball."
                )
            if "dockerfile" not in kwargs:
                # TODO: Scan the tarball for either a Dockerfile or a Containerfile.
                # This could be slow if the tarball is large,
                # and could require buffering/copying the tarball if `fileobj` is not seekable.
                # As a workaround for now, don't support omitting the filename.
                raise PodmanError(
                    "Custom context requires specifying the name of the Dockerfile "
                    "(typically 'Dockerfile' or 'Containerfile')."
                )
            body = kwargs["fileobj"]
        elif "fileobj" in kwargs:
            path = tempfile.TemporaryDirectory()  # pylint: disable=consider-using-with
            filename = pathlib.Path(path.name) / params["dockerfile"]
    
            with open(filename, "w", encoding='utf-8') as file:
                shutil.copyfileobj(kwargs["fileobj"], file)
            body = api.create_tar(anchor=path.name, gzip=kwargs.get("gzip", False))
        elif "path" in kwargs:
            filename = pathlib.Path(kwargs["path"]) / params["dockerfile"]
            # The Dockerfile will be copied into the context_dir if needed
            params["dockerfile"] = api.prepare_containerfile(kwargs["path"], str(filename))
    
            excludes = api.prepare_containerignore(kwargs["path"])
            body = api.create_tar(
                anchor=kwargs["path"], exclude=excludes, gzip=kwargs.get("gzip", False)
            )
    
        post_kwargs = {}
        if kwargs.get("timeout"):
            post_kwargs["timeout"] = float(kwargs.get("timeout"))
    
        response = self.client.post(  # type: ignore[attr-defined]
            "/build",
            params=params,
            data=body,
            headers={
                "Content-type": "application/x-tar",
                # "X-Registry-Config": "TODO",
            },
            stream=True,
            **post_kwargs,
        )
        if hasattr(body, "close"):
            body.close()
    
        if hasattr(path, "cleanup"):
            path.cleanup()
    
        image_id = unknown = None
        marker = re.compile(r"(^[0-9a-f]+)\n$")
        report_stream, stream = itertools.tee(response.iter_lines())
        for line in stream:
            result = json.loads(line)
            if "error" in result:
>               raise BuildError(result["error"], report_stream)
E               podman.errors.exceptions.BuildError: building at STEP "RUN --mount=type=secret,id=example cat /run/secrets/example > /output.txt": while running runtime: exit status 255

podman/domain/images_build.py:151: BuildError
```

Issue #604, opened 2025-11-10
Python version requirements listed in pyproject.toml may not be correct


NicholasBHubbard:
The `pyproject.toml` file lists the `required-python` version as greater or equal to 3.9. The `dependencies` list also lists that we depend on `python_version<'3.11'`. This means that we depend on a version of Python greater or equal to 3.9 and less than 3.11. However, PR #401, which was merged into the main branch, added Python 3.12 support. PR #401 also closed issue #397 which was a request to add support for Python version 3.12 and greater.

The line in `pyproject.toml` that specifies we need a Python version less than 3.11 comes from this string: `"tomli>=1.2.3; python_version<'3.11'",`. Note that it is bundled with the dependency on [tomli](https://pypi.org/project/tomli/), which suggests that they are related. Newer versions of `tomli` do support Python versions greater or equal to 3.11 though, so perhaps we need to increase the minimum version of `tomli` ([Here](https://github.com/hukkin/tomli/blob/master/pyproject.toml) is a link to the `tomli` `pyproject.toml` file).

Is it incorrect that the `pyproject.toml` file says we require a Python version less than 3.11? Would any other steps need to be taken to support newer versions of Python?

I'd be happy to work on a PR for this once I better understand the situation. Thanks!

inknos:
@NicholasBHubbard thanks, it's a bug and we support python >= 3.9 and on. feel free to open a PR, and it would be a nice fix.

inknos:
Closing, see pr for reason. Apologies for the noise

PR #605, opened 2025-11-10
(issue #604) Remove dependency on python_version<3.11

NicholasBHubbard:
This PR aims to solve issue #604. 

I simply removed the dependency on `python_version<'3.11'` from the `dependencies` list. I determined that there was no need to change the minimum required version of `tomli`, due to `tomli` version 1.2.3 supporting any Python version greater or equal to 3.6. [Here](https://github.com/hukkin/tomli/blob/1.2.3/pyproject.toml) is a link to the `pyproject.toml` file for `tomli` version 1.2.3.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ classifiers = [
 # ~= with version numbers
 dependencies = [
     "requests >=2.24",
-    "tomli>=1.2.3; python_version<'3.11'",
+    "tomli>=1.2.3",
     "urllib3",
 ]
 


inknos:
Sorry for missing the point of the issue. I was tired and when I returned to it, I made a wrong evaluation.

The line `"tomli>=1.2.3; python_version<'3.11'"` is an [environment marker](https://packaging.python.org/en/latest/specifications/dependency-specifiers/#environment-markers) that will imply that `tomli` would be installed if pyhton version is below 3.11. [tomli is in python since 3.11](https://docs.python.org/3/library/tomllib.html)

PR #607, opened 2025-12-06
Delete custom cached_property in favor of functools.cached_property

yangdanny97:
~~Change try/except import to use sys.version_info check instead, so that type checkers can statically determine the type.~~

Since podman only supports Python 3.9+, the custom cached_property impl for older versions is no longer necessary.

See https://github.com/facebook/pyrefly/issues/1758 for context


yangdanny97:
Oh, we can just delete that file entirely then if Python 3.8 support is dropped

inknos:
yes, and the references in other files. do you plan to take care of it as part of this PR?

yangdanny97:
Yep, the updated version of the PR deletes the custom cached_property entirely and updates all the call sites to use functools.cached_property

inknos (approved):

PR #611, opened 2025-12-16
Remove deprecated mypy option

jwhonce:
* "no_strict_optional = true" has been deprecated by mypy. This change removes linting errors.

### DIFF
--- a/podman/client.py
+++ b/podman/client.py
@@ -127,11 +127,10 @@ class PodmanClient(AbstractContextManager):
         environment = environment or os.environ
         credstore_env = credstore_env or {}
 
-        if version == "auto":
-            version = None
+        api_version: Optional[str] = None if version == "auto" else version
 
         kwargs = {
-            'version': version,
+            'version': api_version,
             'timeout': timeout,
             'tls': False,
             'credstore_env': credstore_env,
--- a/podman/domain/containers_run.py
+++ b/podman/domain/containers_run.py
@@ -108,6 +108,6 @@ class RunMixin:  # pylint: disable=too-few-public-methods
             container.remove()
 
         if exit_status != 0:
-            raise ContainerError(container, exit_status, command, image_id, log_iter)
+            raise ContainerError(container, exit_status, command or [], image_id, log_iter)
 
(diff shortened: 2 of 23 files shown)


inknos:
I didn't find the deprecation notice but since ["it's evil"](https://mypy.readthedocs.io/en/stable/config_file.html#confval-strict_optional) let's remove it and work on fixing the type checks

Honny1:
It seems that disabling this option exposed several issues. I think we should address them.

PTAL @jwhonce

jwhonce:
@inknos @Honny1 Please review now that tests pass.

podman/domain/containers_run.py
```diff
@@ -118,6 +118,6 @@ def remove_container(container_object: Container) -> None:
             container.remove()
 
         if exit_status != 0:
-            raise ContainerError(container, exit_status, command, image_id, log_iter)
```
inknos, line 111: I am only confused about this line. what's the meaning of the error?

inknos: the rest looks good

podman/client.py
```diff
@@ -128,7 +128,7 @@ def from_env(
         credstore_env = credstore_env or {}
 
         if version == "auto":
-            version = None
+            version = None  # type: ignore[assignment]
```
Honny1: Maybe use Optional for version?

podman/domain/manager.py
```diff
@@ -39,7 +39,7 @@ def __init__(
             podman_client: PodmanClient() configured to connect to Podman object.
         """
         super().__init__()
-        self.client = client
+        self.client: APIClient = client  # type: ignore[assignment]
```
Honny1: Maybe use Optional for the client?

podman/domain/manager.py
```diff
@@ -113,7 +113,7 @@ def __init__(
...
```
Honny1: ditto

Honny1 (commented):
I have some nits about ignoring errors. Otherwise LGTM.

Honny1:
@jwhonce, can you please rebase on main? Thanks

jwhonce:
@Honny1 Done

Honny1 (approved):
LGTM

PTAL @inknos

PR #615, opened 2026-01-19
networks_manager: fix dns_enabled parameter description

### LINKED ISSUE #613: Misleading kwargs explanation for NetworksManager create function
[dns_enabled (bool): When True, do not provision DNS for this network.](https://github.com/containers/podman-py/blob/main/podman/domain/networks_manager.py#L42)

I have found that if I don't specify this as true, DNS for the network is disabled. So, I found this explanation confusing. I think the author meant that when true, the DNS is provisioned. English is my secondary language, but chatgpt agrees with my confusion.

sekyHC:
Fixes #613

### DIFF
--- a/podman/domain/networks_manager.py
+++ b/podman/domain/networks_manager.py
@@ -39,8 +39,8 @@ class NetworksManager(Manager):
         Keyword Args:
             attachable (bool): Ignored, always False.
             check_duplicate (bool): Ignored, always False.
-            dns_enabled (bool): DNSEnabled is whether name resolution is active for container on this Network. 
-                                Only supported with the bridge driver.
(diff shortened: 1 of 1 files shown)


inknos:
@sekyHC thanks, looks good and I would like to merge it. could you please squash the commits together? it would just make it a bit more clean in the log since is a single change. then I'll merge it

(I can clarify if what I mean is not clear)

inknos:
I found that I could squash from the webui. nothing is requested from you now, thanks @sekyHC :D

sekyHC:
> I found that I could squash from the webui. nothing is requested from you now, thanks @sekyHC :D

Sorry, I was not online. Yes, you can squash commits when merging, unless it has been disabled in GitHub repo settings.

PR #617, opened 2026-01-19
Add healthcheck to create and run

inknos:


podman/domain/containers_create.py
```diff
@@ -736,6 +766,34 @@ def parse_host_port(_container_port, _protocol, _host):
...
+            "health_startup_retries",
+            "health_startup_success",
+            "health_startup_timeout",
+            "health_timeout",
+        ]
+        # the healthcheck section of parameters accepted can be either no_healthcheck or a series
+        # of healthcheck parameters
+        if args.get("no_healthcheck"):
+            for command in health_commands_data:
+                if command in args:
```
Honny1: I think this can be simplified with the intersection of sets. 

```
if conflicts := set(args) & set(health_commands_data):
        raise ValueError(f"Cannot set {conflicts.pop()} when no_healthcheck is True")
```

Also, it could be used in the `else` branch.

inknos: Good, I also changed the type of `health_commands_data` to set so no casting is needed

Honny1 (approved):
LGTM

PR #622, opened 2026-02-02
Implement Quadlet, QuadletManager and GET calls

inknos:


TomSweeneyRedHat:
LGTM

podman/domain/quadlets.py
```diff
@@ -0,0 +1,165 @@
...
+        Raises:
+            NotFound: when quadlet could not be found
+            APIError: when service reports an error
+        """
+        response = self.client.get(
+            "/quadlets/json",
+            params={"filters": api.prepare_filters({"name": name})},
+        )
+        response.raise_for_status()
+        return self.prepare_model(attrs=response.json()[0])
```
jankaluza: Are we sure the response.json() is always a list with single item? If the list is empty, this would fail with IndexError. Not sure if that somehow translates to NotFound or not. But maybe it is guaranteed to always return non-empty list. Just something to think about.

jankaluza: I was thinking about doing something like that if needed:

```
data = response.json()
if not data:
    raise NotFound(f"Quadlet {name} not found")
return self.prepare_model(attrs=data[0])
```

podman/domain/quadlets.py
```diff
@@ -0,0 +1,165 @@
...
+        """Check if a quadlet exists.
+
+        Args:
+            key: Name of the quadlet to check.
+
+        Returns:
+            True if the quadlet exists, False otherwise.
+        """
+        response = self.client.get(f"/quadlets/{key}/exists")
+        return response.ok
```
jankaluza: This will hide any server-side error and tells the caller that the quadlet does not exist instead. I think we should still raise the error in case server was not able to decide whether the quadlet really exists. I don't know the real `/quadlets` API, but in general, I would do something like:

```
if response.status_code == 404:
    return False
response.raise_for_status()
return True
```

podman/domain/quadlets.py
```diff
@@ -0,0 +1,165 @@
...
+    def get_contents(self) -> str:
+        """Get the contents of this quadlet file.
+
+        Returns:
+            The complete quadlet file contents as a string.
+        """
+        return self.manager.get_contents(self.name)
+
+    def print_contents(self) -> None:
+        """Get the contents of this quadlet file.
```
jankaluza: The doc-string is forgotten copy-paste from `get_contents` :-). I think it needs to be updated.

jankaluza:
Just two things worth fixing and one to think about, otherwise LGTM.

inknos:
> Just two things worth fixing and one to think about, otherwise LGTM.

thanks I am on it

podman/domain/quadlets.py
inknos: fixed

.pre-commit-config.yaml
```diff
@@ -8,7 +8,7 @@ repos:
     - id: trailing-whitespace
 - repo: https://github.com/astral-sh/ruff-pre-commit
   # Ruff version.
-  rev: v0.12.8
+  rev: v0.14.13
```
Honny1, line 11: Non-blocking: We should automate these updates.

podman/domain/quadlets.py
inknos: added a change in the code and test `test_exists_error_status` to check the behavior

podman/domain/quadlets.py
inknos: added suggestion and `test_get_empty_list` to check it

.pre-commit-config.yaml
inknos: 100%. do you know how?

inknos: addressed by https://github.com/containers/podman-py/pull/626

Honny1:
CI is green

PR #623, opened 2026-02-03
Fix pytest flag for rootful test

inknos:
I found a test that needs to be run if root only

### DIFF
--- a/podman/tests/integration/test_container_create.py
+++ b/podman/tests/integration/test_container_create.py
@@ -1,11 +1,10 @@
 import unittest
 import re
-import os
 import pytest
 
 import podman.tests.integration.base as base
 from podman import PodmanClient
-from podman.tests.utils import PODMAN_VERSION
+from podman.tests.utils import PODMAN_VERSION, is_root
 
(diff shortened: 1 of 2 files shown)


podman/tests/integration/test_container_create.py
```diff
@@ -397,6 +397,7 @@ def test_container_mounts_without_rw_as_default(self):
                 f"size={mount['size']},rprivate,nosuid,nodev,tmpcopyup",
             )
 
+    @pytest.mark.skipif(os.geteuid() != 0, reason='Skipping, not running as root')
```
Honny1: I'm a little confused by the reason text and the PR description. Should it be `reason="Skipping, rootless user"`?

Non-blocking:
I would also suggest creating a helper function:

```Python
def is_root():
    return os.geteuid() == 0
```
This would be more readable and could be reused in other cases if needed.

inknos: I think I had twisted brain cells at the time of writing. also, your comment made sense so I implemented the helper, and finally, I changed commit message an fixed the reasons. Now it should be better

PR #634, opened 2026-03-05
Feature/add image import

Andre-85:
Hi @all,
while using podman-py i realized that there is no python-binding for podman import, so i created one by .myself.

I works straight forward:
```
import podman
c = podman.PodmanClient()
c.images.import_image(<tarfile name>, reference=<like in the podman import command>)
```

I tested it successfully on Ubuntu 24.04 with podman 4.9.3 and on Ubuntu 26.04 snapshot 3 (snapshot 4 does not work on qemu for some reason) and podman 5.7.0.

What do you think?

Greetings,
André

### DIFF
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -8,6 +8,7 @@ import os
 import urllib.parse
 from typing import Any, Literal, Optional, Union
 from collections.abc import Iterator, Mapping, Generator
+from contextlib import nullcontext
 from pathlib import Path
 import requests
 
@@ -165,6 +166,87 @@ class ImagesManager(BuildMixin, Manager):
 
         # Pass the response body to the generator
         return _generator(response.json())
+
+    def import_image(
+        self,
+        data: Optional[bytes] = None,
+        file_path: Optional[os.PathLike] = None,
+        url: Optional[str] = None,
+        **kwargs,
+    ) -> "Image":
+        """Import a tarball as an image (equivalent of 'podman import').
+
+        Args:
+            file_path: Path to the tarball to import.
+            data: tarball raw data (bytes)
+            url: Url to the tarball to import.
+
+        Keyword Args:
+            reference: Optional reference for the new image (e.g. 'myimage:latest').
+            message: Optional commit message.
+            changes: Optional list of Dockerfile-style instructions
+                     (e.g. ['CMD /bin/bash', 'ENV FOO=bar']).
+
+        Returns:
+            An Image object for the newly imported image.
+
+        Raises:
+            APIError: when service returns an error.
+        """
+        # Check that exactly one of the data or file_path is provided
+        if sum(x is not None for x in (data, file_path, url)) != 1:
+            raise PodmanError(
+                "Exactly one parameter should be set from 'data', 'file_path' and 'url' parameters."
+            )
+
+        # Check if url given it is supported
+        if url:
+            uri = urllib.parse.urlparse(url)
+            if uri.scheme not in api.APIClient.supported_schemes:
+                raise ValueError(
+                    f"The scheme '{uri.scheme}' must be one of {api.APIClient.supported_schemes}"
+                )
+
+        # Set the parameters
+        params = {}
+        if reference := kwargs.get("reference"):
+            params["reference"] = reference
+        if message := kwargs.get("message"):
+            params["message"] = message
+        if changes := kwargs.get("changes"):
+            params["changes"] = changes  # requests sends repeated keys as a list
+        if url:
+            params["url"] = url
+
+        # Get either from file or from raw data
+        post_data_context = None
+        if file_path is not None:
+            post_data_context = Path(file_path).open("rb")
+        elif data is not None:
+            post_data_context = io.BytesIO(data)
+        elif url is not None:
+            post_data_context = nullcontext()
+
+        # Post it
+        image_id = None
+        with post_data_context as post_data:
+            response = self.client.post(
+                "/images/import",
+                params=params,
+                data=post_data,
+                headers={"Content-Type": "application/x-tar"},
+            )
+            response.raise_for_status()
+
+            body = response.json()
+            image_id = body.get("Id")
+
+            if image_id is None:
+                raise APIError(
+                    response.url, response=response, explanation="No image id was returned"
+                )
+
+        return self.get(image_id)
 
     def prune(
         self,
--- a/podman/domain/images_manager.py
+++ b/podman/domain/images_manager.py
@@ -8,6 +8,7 @@ import os
 import urllib.parse
 from typing import Any, Literal, Optional, Union
 from collections.abc import Iterator, Mapping, Generator
+from contextlib import nullcontext
 from pathlib import Path
 import requests
 
@@ -157,6 +158,81 @@ class ImagesManager(BuildMixin, Manager):
             )
             response.raise_for_status()
             return _generator(response.json())
+
+    def import_image(
+        self,
+        data: Optional[bytes] = None,
+        file_path: Optional[os.PathLike] = None,
+        url: Optional[str] = None,
+        **kwargs,
+    ) -> "Image":
+        """Import a tarball as an image (equivalent of 'podman import').
+
+        Args:
+            file_path: Path to the tarball to import.
+            data: tarball raw data (bytes)
+            url: Url to the tarball to import.
+
+        Keyword Args:
+            reference: Optional reference for the new image (e.g. 'myimage:latest').
+            message: Optional commit message.
+            changes: Optional list of Dockerfile-style instructions
+                     (e.g. ['CMD /bin/bash', 'ENV FOO=bar']).
+
+        Returns:
+            An Image object for the newly imported image.
+
+        Raises:
+            APIError: when service returns an error.
+        """
+        # Check that exactly one of the data or file_path is provided
+        if sum(x is not None for x in (data, file_path, url)) != 1:
+            raise PodmanError(
+                "Exactly one parameter should be set from 'data', 'file_path' and 'url' parameters."
+            )
+
+        # Check if url given it is supported
+        if url:
+            uri = urllib.parse.urlparse(url)
+            supported_schemes = ["http", "https"]
+            if uri.scheme not in supported_schemes:
+                raise ValueError(f"The scheme '{uri.scheme}' must be one of {supported_schemes}")
+
+        # Set the parameters
+        params = {}
+        if reference := kwargs.get("reference"):
+            params["reference"] = reference
+        if message := kwargs.get("message"):
+            params["message"] = message
+        if changes := kwargs.get("changes"):
+            params["changes"] = changes  # requests sends repeated keys as a list
+        if url:
+            params["url"] = url
+
+        # Get either from file or from raw data and post it
+        with (
+                Path(file_path).open("rb")
+                if file_path is not None
+                else (io.BytesIO(data) if data is not None else nullcontext()) as post_data
+        ):
+
+            response = self.api.post(
+                "/images/import",
+                params=params,
+                data=post_data,
+                headers={"Content-Type": "application/x-tar"},
+            )
+            response.raise_for_status()
+
+            body = response.json()
+            image_id = body.get("Id")
+
+            if image_id is None:
+                raise APIError(
+                    response.url, response=response, explanation="No image id was returned"
+                )
+
+        return self.get(image_id)
 
     def prune(
         self,
--- a/podman/tests/integration/test_images.py
+++ b/podman/tests/integration/test_images.py
@@ -17,9 +17,11 @@
 import io
 import os
 import json
+import http.server
 import platform
 import tarfile
 import tempfile
+import threading
 import types
 import unittest
 import random
@@ -293,3 +295,94 @@ class ImagesIntegrationTest(base.IntegrationTest):
             e.exception.explanation,
             r"failed to connect: dial tcp: lookup fake\.ip\.addr.+no such host",
         )
+
+    def test_import_from_file(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            # Create test folder with test file
+            base_dir = os.path.join(tmpdir, "test")
+            os.mkdir(base_dir)
+            open(os.path.join(base_dir, "foobar"), "w").close()
+
+            # Pack the testfile with the test folder in a tar
+            tar_path = os.path.join(tmpdir, "test.tar.gz")
+            with tarfile.open(tar_path, "w:gz") as tar:
+                tar.add(base_dir, arcname="test")
+
+            # Import it
+            image = self.client.images.import_image(file_path=tar_path, message="test")
+            self.assertIsInstance(image, Image)
+            self.assertEqual(image.attrs.get("Comment"), "test")
+            container = self.client.containers.create(image, command=["."])
+
+            # Check the imported image
+            actual = container.get_archive("./test/foobar")
+            self.assertEqual(len(actual), 2)
+            self.assertEqual(actual[1]["linkTarget"], "/test/foobar")
+
+            # Clean up
+            container.remove(force=True)
+            image.remove(force=True)
+
+    def test_import_from_data(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            # Create test folder with test file
+            base_dir = os.path.join(tmpdir, "test")
+            os.mkdir(base_dir)
+            open(os.path.join(base_dir, "foobar"), "w").close()
+
+            # Pack the testfile with the test folder in a tar buffer
+            tar_buffer = io.BytesIO()
+            with tarfile.open(fileobj=tar_buffer, mode="w:gz") as tar:
+                tar.add(base_dir, arcname="test")
+            tar_buffer.seek(0)
+
+            # Import it
+            image = self.client.images.import_image(data=tar_buffer.read(), changes=["ENV FOO=bar"])
+            self.assertIsInstance(image, Image)
+            self.assertEqual(image.attrs.get("Config", {}).get("Env"), ["FOO=bar"])
+            container = self.client.containers.create(image, command=["."])
+
+            # Check the imported image
+            actual = container.get_archive("./test/foobar")
+            self.assertEqual(len(actual), 2)
+            self.assertEqual(actual[1]["linkTarget"], "/test/foobar")
+
+            # Clean up
+            container.remove(force=True)
+            image.remove(force=True)
+
+    def test_import_from_url(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            # Create test folder with test file
+            base_dir = os.path.join(tmpdir, "test")
+            os.mkdir(base_dir)
+            open(os.path.join(base_dir, "foobar"), "w").close()
+
+            # Pack the testfile with the test folder in a tar
+            tar_path = os.path.join(tmpdir, "test.tar.gz")
+            with tarfile.open(tar_path, "w:gz") as tar:
+                tar.add(base_dir, arcname="test")
+
+            # Serve it on a http server
+            def handler(*a):
+                return http.server.SimpleHTTPRequestHandler(*a, directory=tmpdir)
+
+            with http.server.HTTPServer(("", 0), handler) as httpd:
+                threading.Thread(target=httpd.serve_forever, daemon=True).start()
+
+                # Import it
+                image = self.client.images.import_image(
+                    url=f"http://localhost:{httpd.server_port}/test.tar.gz", message="test"
+                )
+                self.assertIsInstance(image, Image)
+                self.assertEqual(image.attrs.get("Comment"), "test")
+                container = self.client.containers.create(image, command=["."])
+
+                # Check the imported image
+                actual = container.get_archive("./test/foobar")
+                self.assertEqual(len(actual), 2)
+                self.assertEqual(actual[1]["linkTarget"], "/test/foobar")
+
+                # Clean up
+                container.remove(force=True)
+                image.remove(force=True)
--- a/podman/tests/integration/test_images.py
+++ b/podman/tests/integration/test_images.py
@@ -17,9 +17,11 @@
 import io
 import os
 import json
+import http.server
 import platform
 import tarfile
 import tempfile
+import threading
 import types
 import unittest
 import random
@@ -292,3 +294,94 @@ class ImagesIntegrationTest(base.IntegrationTest):
             e.exception.explanation,
             r"failed to connect: dial tcp: lookup fake\.ip\.addr.+no such host",
         )
+
+    def test_import_from_file(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            # Create test folder with test file
+            base_dir = os.path.join(tmpdir, "test")
+            os.mkdir(base_dir)
+            open(os.path.join(base_dir, "foobar"), "w").close()
+
+            # Pack the testfile with the test folder in a tar
+            tar_path = os.path.join(tmpdir, "test.tar.gz")
+            with tarfile.open(tar_path, "w:gz") as tar:
+                tar.add(base_dir, arcname="test")
+
+            # Import it
+            image = self.client.images.import_image(file_path=tar_path, message="test")
+            self.assertIsInstance(image, Image)
+            self.assertEqual(image.attrs.get("Comment"), "test")
+            container = self.client.containers.create(image, command=["."])
+
+            # Check the imported image
+            actual = container.get_archive("./test/foobar")
+            self.assertEqual(len(actual), 2)
+            self.assertEqual(actual[1]["linkTarget"], "/test/foobar")
+
+            # Clean up
+            container.remove(force=True)
+            image.remove(force=True)
+
+    def test_import_from_data(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            # Create test folder with test file
+            base_dir = os.path.join(tmpdir, "test")
+            os.mkdir(base_dir)
+            open(os.path.join(base_dir, "foobar"), "w").close()
+
+            # Pack the testfile with the test folder in a tar buffer
+            tar_buffer = io.BytesIO()
+            with tarfile.open(fileobj=tar_buffer, mode="w:gz") as tar:
+                tar.add(base_dir, arcname="test")
+            tar_buffer.seek(0)
+
+            # Import it
+            image = self.client.images.import_image(data=tar_buffer.read(), changes=["ENV FOO=bar"])
+            self.assertIsInstance(image, Image)
+            self.assertEqual(image.attrs.get("Config", {}).get("Env"), ["FOO=bar"])
+            container = self.client.containers.create(image, command=["."])
+
+            # Check the imported image
+            actual = container.get_archive("./test/foobar")
+            self.assertEqual(len(actual), 2)
+            self.assertEqual(actual[1]["linkTarget"], "/test/foobar")
+
+            # Clean up
+            container.remove(force=True)
+            image.remove(force=True)
+
+    def test_import_from_url(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            # Create test folder with test file
+            base_dir = os.path.join(tmpdir, "test")
+            os.mkdir(base_dir)
+            open(os.path.join(base_dir, "foobar"), "w").close()
+
+            # Pack the testfile with the test folder in a tar
+            tar_path = os.path.join(tmpdir, "test.tar.gz")
+            with tarfile.open(tar_path, "w:gz") as tar:
+                tar.add(base_dir, arcname="test")
+
+            # Serve it on a http server
+            def handler(*a):
+                return http.server.SimpleHTTPRequestHandler(*a, directory=tmpdir)
+
+            with http.server.HTTPServer(("", 0), handler) as httpd:
+                threading.Thread(target=httpd.serve_forever, daemon=True).start()
+
+                # Import it
+                image = self.client.images.import_image(
+                    url=f"http://localhost:{httpd.server_port}/test.tar.gz", message="test"
+                )
+                self.assertIsInstance(image, Image)
+                self.assertEqual(image.attrs.get("Comment"), "test")
+                container = self.client.containers.create(image, command=["."])
+
+                # Check the imported image
+                actual = container.get_archive("./test/foobar")
+                self.assertEqual(len(actual), 2)
+                self.assertEqual(actual[1]["linkTarget"], "/test/foobar")
+
+                # Clean up
+                container.remove(force=True)
+                image.remove(force=True)
--- a/podman/tests/unit/test_imagesmanager.py
+++ b/podman/tests/unit/test_imagesmanager.py
@@ -1,5 +1,6 @@
 import io
 import types
+import urllib.parse
 import unittest
 from unittest.mock import patch
 
@@ -43,6 +44,11 @@ SECOND_IMAGE = {
     "VirtualSize": 23855104,
     "SharedSize": 0,
     "Containers": 0,
+}
+
+IMPORTED_IMAGE = {
+    "Id": "b07571c5220ab38f303c854bd150aede9e5d1d1501a34b54f84916e9f5f8d000",
+    "Digest": "sha256:122ae1d154d021971cd7a739b5d6f029a80b627763b649adeb34c6f13f3c451b",
 }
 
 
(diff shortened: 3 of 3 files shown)


Honny1 (changes requested):
Please add tests for this. Ideally integration tests, since it doesn’t work on your machine with the latest version of Podman.

Andre-85:
> Please add tests for this. Ideally integration tests, since it doesn’t work on your machine with the latest version of Podman.


Yes i will do in the next days.

Greetings,
André

podman/domain/images_manager.py
```diff
@@ -166,6 +166,64 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+        data: Optional[bytes] = None,
+        file_path: Optional[os.PathLike] = None,
+        reference: Optional[str] = None,
+        message: Optional[str] = None,
+        changes: Optional[List[str]] = None,
+    ) -> "Image":
+        """Import a tarball as an image (equivalent of 'podman import').
+
+        Args:
+            source: Path to a tarball (str) or a file-like object opened in binary mode.
```
Honny1: This `arg` does not exist.

podman/domain/images_manager.py
```diff
@@ -166,6 +166,64 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+                     (e.g. ['CMD /bin/bash', 'ENV FOO=bar']).
+
+        Returns:
+            An Image object for the newly imported image.
+
+        Raises:
+            APIError: when service returns an error.
+        """
+        # Check that exactly one of the data or file_path is provided
+        if not data and not file_path:
```
Honny1: I would use `is None` check.

podman/tests/unit/test_imagesmanager.py
```diff
@@ -430,6 +430,53 @@ def test_load(self, mock):
...
+    @requests_mock.Mocker()
+    def test_import(self, mock):
+        with self.assertRaises(PodmanError):
+            self.client.images.import_image()
+
+        with self.assertRaises(PodmanError):
+            self.client.images.import_image(b'data', b'file_path')
+
+        with self.assertRaises(PodmanError):
+            self.client.images.import_image(data=b'data', file_path=b'file_path')
```
Honny1: Why is `file_path` a bytes literal?

podman/domain/images_manager.py
```diff
@@ -166,6 +166,64 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+        if data and file_path:
+            raise PodmanError(
+                "Only one parameter should be set from 'data' and 'file_path' parameters."
+            )
+
+        post_data = data
+        if file_path:
+            # Convert to Path if file_path is a string
+            file_path_object = Path(file_path)
+            post_data = file_path_object.read_bytes()  # Read the tarball file as bytes
```
Honny1: I would stream this data instead of reading a multiple GB file into memory.

Honny1:
Also, please squash your changes into one commit.

Andre-85:
> Also, please squash your changes into one commit.

Yes, will do and i will apply your suggestions.

Andre-85:
> > Also, please squash your changes into one commit.
> 
> Yes, will do and i will apply your suggestions.

I applied the requested changes, added extra tests for setting message and changes and squashed all changes in one commit.

Honny1 (changes requested):
```
ruff (legacy alias)......................................................Failed
- hook id: ruff
- exit code: 1

UP035 `typing.List` is deprecated, use `list` instead
  --> podman/domain/images_manager.py:9:1
   |
 7 | import os
 8 | import urllib.parse
 9 | from typing import Any, Literal, Optional, Union, List
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
10 | from collections.abc import Iterator, Mapping, Generator
11 | from pathlib import Path
   |

UP006 Use `list` instead of `List` for type annotation
   --> podman/domain/images_manager.py:175:27
    |
173 |         reference: Optional[str] = None,
174 |         message: Optional[str] = None,
175 |         changes: Optional[List[str]] = None,
    |                           ^^^^
176 |     ) -> "Image":
177 |         """Import a tarball as an image (equivalent of 'podman import').
    |
help: Replace with `list`

F841 Local variable `tar_path` is assigned to but never used
   --> podman/tests/integration/test_images.py:327:13
    |
325 |             # Pack the testfile with the test folder in a tar buffer
326 |             tar_buffer = io.BytesIO()
327 |             tar_path = os.path.join(tmpdir, "test.tar.gz")
    |             ^^^^^^^^
328 |             with tarfile.open(fileobj=tar_buffer, mode="w:gz") as tar:
329 |                 tar.add(base_dir, arcname="test")
    |
help: Remove assignment to unused variable `tar_path`

Found 3 errors.
No fixes available (2 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

Andre-85:
> ```
> ruff (legacy alias)......................................................Failed
> - hook id: ruff
> - exit code: 1
> 
> UP035 `typing.List` is deprecated, use `list` instead
>   --> podman/domain/images_manager.py:9:1
>    |
>  7 | import os
>  8 | import urllib.parse
>  9 | from typing import Any, Literal, Optional, Union, List
>    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> 10 | from collections.abc import Iterator, Mapping, Generator
> 11 | from pathlib import Path
>    |
> 
> UP006 Use `list` instead of `List` for type annotation
>    --> podman/domain/images_manager.py:175:27
>     |
> 173 |         reference: Optional[str] = None,
> 174 |         message: Optional[str] = None,
> 175 |         changes: Optional[List[str]] = None,
>     |                           ^^^^
> 176 |     ) -> "Image":
> 177 |         """Import a tarball as an image (equivalent of 'podman import').
>     |
> help: Replace with `list`
> 
> F841 Local variable `tar_path` is assigned to but never used
>    --> podman/tests/integration/test_images.py:327:13
>     |
> 325 |             # Pack the testfile with the test folder in a tar buffer
> 326 |             tar_buffer = io.BytesIO()
> 327 |             tar_path = os.path.join(tmpdir, "test.tar.gz")
>     |             ^^^^^^^^
> 328 |             with tarfile.open(fileobj=tar_buffer, mode="w:gz") as tar:
> 329 |                 tar.add(base_dir, arcname="test")
>     |
> help: Remove assignment to unused variable `tar_path`
> 
> Found 3 errors.
> No fixes available (2 hidden fixes can be enabled with the `--unsafe-fixes` option).
> ```

I fixed the issues.

podman/domain/images_manager.py
```diff
@@ -166,6 +166,67 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+        data: Optional[bytes] = None,
+        file_path: Optional[os.PathLike] = None,
+        reference: Optional[str] = None,
+        message: Optional[str] = None,
+        changes: Optional[builtins.list[str]] = None,
+    ) -> "Image":
+        """Import a tarball as an image (equivalent of 'podman import').
+
+        Args:
+            file_path: Path to a tarball (str) or a file-like object opened in binary mode.
```
Honny1: I think description doesn't match `file_path: Optional[os.PathLike] = None`.

podman/domain/images_manager.py
```diff
@@ -166,6 +166,67 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+            response = self.client.post(
+                "/images/import",
+                params=params,
+                data=post_data,
+                headers={"Content-Type": "application/x-tar"},
+            )
+            response.raise_for_status()
+
+            body = response.json()
+            image_id = body.get("Id") or body.get("id")
```
Honny1: What if `[Ii]d` is missing?

Honny1:
PTAL @inknos

podman/domain/images_manager.py
inknos: yup, this should be handled for failure of getting Id

podman/domain/images_manager.py
```diff
@@ -166,6 +166,67 @@ def _generator(body: dict) -> Generator[Image, None, None]:
         # Pass the response body to the generator
         return _generator(response.json())
 
+    def import_image(
+        self,
+        data: Optional[bytes] = None,
+        file_path: Optional[os.PathLike] = None,
+        reference: Optional[str] = None,
+        message: Optional[str] = None,
+        changes: Optional[builtins.list[str]] = None,
```
inknos: I think I'd like to see only the required args as named, the others being kwargs. It's an arbitrary approach, but the library is designed this way, so I would follow this design pattern.

```python
def import_image(
    self,     
    data: Optional[bytes] = None,
    file_path: Optional[os.PathLike] = None,
    **kwargs,
```

Kwargs being: `reference`, `message`, `changes`, and `url`, which is missing. Do you plan to implement it? if not, we'll have to document it with TODO and reference a new issue in the docstring.

podman/domain/images_manager.py
```diff
@@ -166,6 +166,67 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+            )
+
+        # Set the parameters
+        params = {}
+        if reference:
+            params["reference"] = reference
+        if message:
+            params["message"] = message
+        if changes:
+            params["changes"] = changes  # requests sends repeated keys as a list
```
inknos, line 208: Following the comment above, this can be simplified with the same logic of other functions that do `kwargs.get()`

podman/domain/images_manager.py
Andre-85: I fixed it.

podman/domain/images_manager.py
Andre-85: I found out the true field name is "Id" so i will leave the other one out.

I guess i found a similar error handling as requested in the load function:
```
        response.raise_for_status()  # Catch any errors before proceeding
  
          def _generator(body: dict) -> Generator[Image, None, None]:
              # Iterate and yield images from response body
              for item in body["Names"]:
                  yield self.get(item)
  
          # Pass the response body to the generator
          return _generator(response.json())
```

But being honest, I don't get what is happening here...

podman/domain/images_manager.py
Andre-85: i fixed the style. reference, message and changes are implemented and tested (in the integration test). 

The url parameter i did not implement (yet). Would be a solution with **with urllib.request.urlopen(source) as response: ...** ok for you?

podman/domain/images_manager.py
Andre-85: i fixed it

Andre-85:
@Honny1 
@inknos 

I will do the remaining fixes in the next days. After resolving all i will squeeze them in one commit.

podman/domain/images_manager.py
Andre-85: Or alternate with requests:  
```
  with requests.get('https://httpbin.org/get', stream=True) as r:
    for chunk in r.iter_content(chunk_size=8192):
        ....
```

Andre-85:
Hi,
I implemented the request changes and I implemented the missing url feature including unit and integration tests.

Greetings,
André

podman/domain/images_manager.py
```diff
@@ -166,6 +166,82 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+            params["reference"] = reference
+        if message := kwargs.get("message"):
+            params["message"] = message
+        if changes := kwargs.get("changes"):
+            params["changes"] = changes  # requests sends repeated keys as a list
+        if url:
+            params["url"] = url
+
+        # Get either from file or from raw data
+        post_data_context = Path(file_path).open("rb") if file_path else io.BytesIO(data)
```
Honny1: What if `io.BytesIO(None)`?

podman/tests/unit/test_imagesmanager.py
```diff
@@ -430,6 +431,86 @@ def test_load(self, mock):
...
+
+        # Check if url is valid
+        with self.assertRaises(ValueError):
+            self.client.images.import_image(url="not-an-url")
+
+        # Patch Path.read_bytes to mock the file reading behavior
+        with patch("pathlib.Path.open", return_value=io.BytesIO(b"mock tarball data")):
+            mock.post(
+                tests.LIBPOD_URL + "/images/import",
+                json={"Id": "quay.io/fedora:latest"},
```
Honny1: Will the actual import return an image tag or an image ID?

Andre-85:
Hi @all,
I implemented the requests.

To the requests and the fixes:
1. io.BytesIO(None): Instead of returning a empty temporary file, i  directly None which is the default value of the optional data kwarg parameter
2. Mock for images/import: Yes you were right, and id (or better digest) is returned.

Greetings,
André

podman/tests/integration/test_images.py
```diff
@@ -294,3 +294,80 @@ def test_scp(self):
...
+            open(os.path.join(base_dir, "foobar"), "w").close()
+
+            # Pack the testfile with the test folder in a tar
+            tar_path = os.path.join(tmpdir, "test.tar.gz")
+            with tarfile.open(tar_path, "w:gz") as tar:
+                tar.add(base_dir, arcname="test")
+
+            # Serve it on a http server
+            handler = lambda *a: http.server.SimpleHTTPRequestHandler(*a, directory=tmpdir)
+            with http.server.HTTPServer(("", 8000), handler) as httpd:
```
Honny1: I am not sure about hardcoding the port.

podman/tests/integration/test_images.py
```diff
@@ -294,3 +294,80 @@ def test_scp(self):
             e.exception.explanation,
             r"failed to connect: dial tcp: lookup fake\.ip\.addr.+no such host",
         )
+
+    def test_import_from_file(self):
```
Honny1, line 299: All new integration tests should do cleanups.

podman/domain/images_manager.py
```diff
@@ -166,6 +167,88 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+                data=post_data,
+                headers={"Content-Type": "application/x-tar"},
+            )
+            response.raise_for_status()
+
+            body = response.json()
+            image_id = body.get("Id")
+
+        if image_id is None:
+            raise APIError(response.url, response=response, explanation="No image id was returned")
```
Honny1: Response doesn't exist in this scope.

podman/tests/integration/test_images.py
Andre-85: But howto do in another way? The image is pulled by podman itself and not by podman-py. So I cannot use requests_mock

podman/tests/integration/test_images.py
Andre-85: i will do

podman/domain/images_manager.py
Andre-85: I will have a look

podman/tests/integration/test_images.py
Andre-85: I guess you meant setting port to 0 and let the os choose

podman/tests/integration/test_images.py
Andre-85: I did

podman/domain/images_manager.py
Andre-85: This was a indent mistake

podman/domain/images_manager.py
```diff
@@ -166,6 +167,90 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+            post_data_context = nullcontext()
+
+        # Post it
+        image_id = None
+        with post_data_context as post_data:
+            response = self.client.post(
+                "/images/import",
+                params=params,
+                data=post_data,
+                headers={"Content-Type": "application/x-tar"},
```
Honny1, line 222: Is this Header correct when the input is a URL?

podman/domain/images_manager.py
```diff
@@ -166,6 +167,90 @@ def _generator(body: dict) -> Generator[Image, None, None]:
         # Pass the response body to the generator
         return _generator(response.json())
 
+    def import_image(
```
Honny1, line 162: Nonblocking: I'm not sure if `podman-py` is a drop-in replacement for `docker-py`. Is it? @inknos

If so, we should probably match `docker-py`.

podman/domain/images_manager.py
```diff
@@ -166,6 +167,90 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+        self,
+        data: Optional[bytes] = None,
```
Honny1, line 164: What if the data will be empty `b""`?

Honny1 (commented):
I have last questions about changes.

podman/domain/images_manager.py
Andre-85: Hi,
I did some comparison to docker-py.
First i saw that docker-py is using a different endpoint, it uses images/create. But from the description (Link: https://docs.docker.com/reference/api/engine/version/v1.54/#tag/Image/operation/ImageCreate) i would say that looks like misuse of this endpoint, because the descriptions are all pointing to oci images and not to raw tar files.

I also did a comparison with command line podman (**podman --remote --debug import import.tar**). I can see here that command line podman is also using the equivalent of images/import like my patch. In general I would say that for podman-py it is more important to mimic the behavior of  command line podman than the behavior of docker-py.

Further I think that their design for the import_image functions have serious problems (Link: https://github.com/docker/docker-py/blob/main/tests/integration/api_image_test.py#L161) finding out for the src arg, what kind of source it is (file/url or raw bytes). So i think the approach with kwargs data, file_path and url is better and fits more to the load function in podman-py which uses already data.

But i think also providing the convenience wrappers import_image_from_url, import_image_from_data and import_image_from_file like docker-py does should be simple to do.

What do you think?

Greetings,
André

podman/domain/images_manager.py
Andre-85: I did some investigation:

I assume that podman --remote is using the API correctly.

Since importing images and using an alternate podman socket is not allow sametime (some error like "url and r parameter is not allowed sametime" is shown), put a MITM with socket and started the podman system service on a different socket:
```
podman system service unix:///tmp/podman.sock --time=0
socat -v UNIX-LISTEN:/run/user/1000/podman/podman.sock UNIX-CONNECT:/tmp/podman.sock
podman --remote import ...

```

The result for an empty file loaded from file_path is:
```
> 2026/03/27 09:59:38.000477279  length=78 from=0 to=77
GET /v4.9.3/libpod/_ping HTTP/1.1\r
Host: d\r
User-Agent: Go-http-client/1.1\r
\r
< 2026/03/27 09:59:38.000477854  length=380 from=0 to=379
HTTP/1.1 200 OK\r
Api-Version: 1.41\r
Builder-Version: \r
Buildkit-Version: \r
Cache-Control: no-cache\r
Docker-Experimental: true\r
Libpod-Api-Version: 4.9.3\r
Libpod-Buildah-Version: 1.33.7\r
Ostype: linux\r
Pragma: no-cache\r
Server: Libpod/4.9.3 (linux)\r
X-Reference-Id: 0xc000708010\r
Date: Fri, 27 Mar 2026 08:59:38 GMT\r
Content-Length: 2\r
Content-Type: text/plain; charset=utf-8\r
\r
OK
```

For an non-empty file loaded from file_path:
```
> 2026/03/27 10:03:13.000103092  length=78 from=0 to=77
GET /v4.9.3/libpod/_ping HTTP/1.1\r
Host: d\r
User-Agent: Go-http-client/1.1\r
\r
< 2026/03/27 10:03:13.000103678  length=380 from=0 to=379
HTTP/1.1 200 OK\r
Api-Version: 1.41\r
Builder-Version: \r
Buildkit-Version: \r
Cache-Control: no-cache\r
Docker-Experimental: true\r
Libpod-Api-Version: 4.9.3\r
Libpod-Buildah-Version: 1.33.7\r
Ostype: linux\r
Pragma: no-cache\r
Server: Libpod/4.9.3 (linux)\r
X-Reference-Id: 0xc000708010\r
Date: Fri, 27 Mar 2026 09:03:13 GMT\r
Content-Length: 2\r
Content-Type: text/plain; charset=utf-8\r
\r
OK
```
For an non-empty file loaded from url:
```
> 2026/03/27 09:53:33.000864558  length=78 from=0 to=77
GET /v4.9.3/libpod/_ping HTTP/1.1\r
Host: d\r
User-Agent: Go-http-client/1.1\r
\r
< 2026/03/27 09:53:33.000865215  length=380 from=0 to=379
HTTP/1.1 200 OK\r
Api-Version: 1.41\r
Builder-Version: \r
Buildkit-Version: \r
Cache-Control: no-cache\r
Docker-Experimental: true\r
Libpod-Api-Version: 4.9.3\r
Libpod-Buildah-Version: 1.33.7\r
Ostype: linux\r
Pragma: no-cache\r
Server: Libpod/4.9.3 (linux)\r
X-Reference-Id: 0xc000392028\r
Date: Fri, 27 Mar 2026 08:53:33 GMT\r
Content-Length: 2\r
Content-Type: text/plain; charset=utf-8\r
\r
OK
```

So in all cases the Content-Type is set to text/plain with looks odd, i would say setting in all cases application/x-tar like I did looks more resonable or?

What do you think?

Greetings,
André

podman/domain/images_manager.py
Andre-85: That's correct, I have overseen this case. Just an empty container should be created. I improved filling the context for the request with check `ìs not None`, means its not default.

```
        # Get either from file or from raw data
        post_data_context = None
        if file_path is not None:
            post_data_context = Path(file_path).open("rb")
        elif data is not None:
            post_data_context = io.BytesIO(data)
        elif url is not None:
            post_data_context = nullcontext()
```

Andre-85:
I have made comments about docker-py compatibility and about the `Content-Type` (see above). The `data=b""`  corner case I fixed

podman/domain/images_manager.py
```diff
@@ -166,6 +167,90 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+            APIError: when service returns an error.
+        """
+        # Check that exactly one of the data or file_path is provided
+        if data is None and file_path is None and url is None:
+            raise PodmanError("The 'data' or 'file_path' or 'url' parameter should be set.")
+
+        if (data and file_path) or (url and data) or (url and file_path):
+            raise PodmanError(
+                "Only one parameter should be set from 'data', 'file_path' and 'url' parameters."
+            )
```
Honny1:
```suggestion
       if sum(x is not None for x in (data, file_path, url)) > 1:
           raise PodmanError("Only one parameter should be set from 'data', 'file_path' and 'url' parameters.")
```

podman/tests/integration/test_images.py
```diff
@@ -294,3 +294,89 @@ def test_scp(self):
...
+            self.assertEqual(image.attrs.get("Config", {}).get("Env"), ["FOO=bar"])
+            container = self.client.containers.create(image, command=["."])
+
+            # Check the imported image
+            actual = container.get_archive("./test/foobar")
+            self.assertEqual(len(actual), 2)
+            self.assertEqual(actual[1]["linkTarget"], "/test/foobar")
+
+            # Clean up
+            image.remove(force=True)
```
Honny1, line 352: Don't forget about the container.

podman/tests/integration/test_images.py
```diff
@@ -294,3 +294,89 @@ def test_scp(self):
...
+
+    def test_import_from_data(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            # Create test folder with test file
+            base_dir = os.path.join(tmpdir, "test")
+            os.mkdir(base_dir)
+            open(os.path.join(base_dir, "foobar"), "w").close()
```
Honny1, line 331:
```suggestion
            Path(base_dir, "foobar").touch()
```

podman/tests/integration/test_images.py
Honny1: In all integration tests.

podman/tests/integration/test_images.py
Honny1: In all integration tests.

podman/domain/images_manager.py
Honny1: Note for me: 
> Check this

Honny1: This seems odd. The first thing I noticed is an older version of Podman (`4.9.3`). Can you please test it with version `5.8`?

Honny1 (changes requested):

Andre-85:
Ok, i will do the requested changes the next days

podman/domain/images_manager.py
Andre-85: Guess you did a minor mistake here: You change proposal covers only the more than one parameter set, but not the none of the parameters set. So i change check from `>1` to `!=1` and i adapted the error message a bit

podman/tests/integration/test_images.py
Andre-85: I thought since you are not using pathlib at all within your integration tests, I should not use it for a single statement and do a mixture of pathlib / os module file handling?

podman/tests/integration/test_images.py
Andre-85: i was not aware that deleting the image the containers are based on is not deleting them automatically. I will fix it

Andre-85:
I did the requested changes and/or modified them a bit, in case of using pathlib i omitted it, see above. I did no further adding of functions for docker-py compat, since you said that this should not block here. Perhaps we should do this in an own pull-request.

Greetings,
André

Andre-85:
Just did an rewrite of a lambda to make a function to pass "pre-commit"

podman/tests/integration/test_images.py
Honny1: I think this looks cleaner, but it is not blocking:

```python
base_dir = Path(tmpdir) / "test"
base_dir.mkdir(parents=True, exist_ok=True)
(base_dir / "foobar").touch()

# or

(Path(tmpdir) / "test").mkdir(parents=True, exist_ok=True)
(Path(tmpdir) / "test" / "foobar").touch()
```

podman/domain/images_manager.py
Honny1: Well, Podman tries to be a replacement of Docker, so podman-py should be able to replace docker-py. On one hand, I think it doesn't matter if it is called the libpod API or the compat API of Podman if is supported only for full replacement, but it would mean that with podman-py you cannot use Docker. I am not sure which way this compatibility should work. 

cc @inknos 


So definitely, I think convenience wrappers are a good idea, but for another PR.

Honny1 (approved):
Code LGTM. Just a non-blocking comment. Also, I'm not sure how strictly we should enforce compatibility between Podman and Docker.

Andre-85:
Any news?

Honny1:
PTAL @inknos

podman/domain/images_manager.py
```diff
@@ -146,17 +147,106 @@ def load(
...
+            post_data_context = Path(file_path).open("rb")
+        elif data is not None:
+            post_data_context = io.BytesIO(data)
+        elif url is not None:
+            post_data_context = nullcontext()
+
+        # Post it
+        image_id = None
+        with post_data_context as post_data:
             response = self.client.post(
```
Honny1: Is this correct? Should it be this: `response = self.api.post(`?

podman/domain/images_manager.py
```diff
@@ -146,17 +147,106 @@ def load(
...
+        # Check that exactly one of the data or file_path is provided
+        if sum(x is not None for x in (data, file_path, url)) != 1:
+            raise PodmanError(
+                "Exactly one parameter should be set from 'data', 'file_path' and 'url' parameters."
+            )
+
+        # Check if url given it is supported
+        if url:
+            uri = urllib.parse.urlparse(url)
+            if uri.scheme not in api.APIClient.supported_schemes:
```
Honny1: The `api.APIClient.supported_schemes` contians for Podman service connection. I am not sure if this is intended.

podman/domain/images_manager.py
```diff
@@ -146,17 +147,106 @@ def load(
...
         def _generator(body: dict) -> Generator[Image, None, None]:
             # Iterate and yield images from response body
             for item in body["Names"]:
                 yield self.get(item)
 
-        with Path(file_path).open("rb") if file_path else io.BytesIO(data) as stream:
+        # Pass the response body to the generator
+        return _generator(response.json())
+
+    def import_image(
```
Honny1, line 162: Does it support all args that the API implementation supports? Ref: https://github.com/podman-container-tools/podman/blob/f19c577355dd5fab3411aee8d12b47da61f2237b/pkg/api/handlers/libpod/images.go#L424

Honny1 (commented):
Sorry for the long inactivity. I have some additional comments.

podman/domain/images_manager.py
Andre-85: I made the response like all the other functions in images_manager. All use self.client.post()

podman/domain/images_manager.py
Andre-85: My idea behind this was that for this parameter (get an image from an external url) all schemes that are supported by the API itself should be possible too.

podman/domain/images_manager.py
Andre-85: Guess you are right. It looks like only http is supported (this is also the only test case i made for this function). I will the limit the url parameter to http and test it in the next days.

podman/domain/images_manager.py
Honny1: Well, I checked `images_manager.py` and it doesn't contain any `self.client.post`. Can you try rebasing on upstream main?

podman/domain/images_manager.py
Honny1: A user passing `url="https://example.com/image.tar"` would be rejected because https is not in `supported_schemes`, even though the Podman daemon can fetch from HTTPS URLs just fine. The `supported_schemes` contains API supported sechemas to connecting to the Podman `"daemon"`.

podman/domain/images_manager.py
Andre-85: Rebasing is done, and it's changed to self.api.post

podman/domain/images_manager.py
Andre-85: Ok, i fixed the scheme check as well. To what do you refer exactly when you write "even though the Podman daemon can fetch from HTTPS URLs just fine"?

podman/domain/images_manager.py
Andre-85: See above

Andre-85:
I made all requested changes. For details see my comments above.

podman/domain/images_manager.py
```diff
@@ -158,6 +159,88 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+            params["message"] = message
+        if changes := kwargs.get("changes"):
+            params["changes"] = changes  # requests sends repeated keys as a list
+        if url:
+            params["url"] = url
+
+        # Get either from file or from raw data
+        post_data_context = None
+        if file_path is not None:
+            post_data_context = Path(file_path).open("rb")
```
Honny1: I am afraid that this can leak a file handle. 

Maybe this aprach:
```python
with Path(file_path).open("rb") if file_path is not None else (
    io.BytesIO(data) if data is not None else nullcontext()
) as post_data:
```

podman/domain/images_manager.py
Honny1: @Andre-85 Did you check this? I would use: `headers = {"Content-Type": "application/x-tar"} if url is None else {}`

podman/domain/images_manager.py
```diff
@@ -158,6 +159,88 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+            uri = urllib.parse.urlparse(url)
+            supported_schemes = ["http", "https"]
+            if uri.scheme not in supported_schemes:
+                raise ValueError(
+                    f"The scheme '{uri.scheme}' must be one of {supported_schemes}"
+                )
+
+        # Set the parameters
+        params = {}
+        if reference := kwargs.get("reference"):
```
Honny1, line 205: What if `reference == ""`?

Andre-85:
"Did you check this? I would use: headers = {"Content-Type": "application/x-tar"} if url is None else {}".  I looked it up the podman API forces the content-type of "application/x-tar", see https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImageImportLibpod. But yes also in my option it should not necessary to set a content-type if you reference the image to import with an url

podman/domain/images_manager.py
Andre-85: I would the leave the checking of the correct format of reference to the podman itself. I don't know in detail but i guess there are further restrictions, then just no empty string not allowed

podman/domain/images_manager.py
Andre-85: I am not entirely sure if a file descriptor can be leaked with my implementation, since it puts just with context to a variable before using it in the with statement, so also a __exit__ should be called there and the file descriptors are closed. But I am ok with your proposal so I used it.

Andre-85:
I applied all the recommended changes and/or commented on them.

podman/domain/images_manager.py
Honny1: Also missing OS/Architecture/Variant kwargs.

podman/tests/integration/test_images.py
Honny1: What about this?

Andre-85:
Fixed formatting and squeezed in one commit

podman/domain/images_manager.py
```diff
@@ -158,6 +159,81 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+        if url:
+            params["url"] = url
+
+        # Get either from file or from raw data and post it
+        with (
+                Path(file_path).open("rb")
+                if file_path is not None
+                else (io.BytesIO(data) if data is not None else nullcontext()) as post_data
+        ):
+
```
Honny1, line 218: Ci is complaining abtou this white space.

podman/domain/images_manager.py
```diff
@@ -158,6 +159,81 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+        # Check if url given it is supported
+        if url:
+            uri = urllib.parse.urlparse(url)
+            supported_schemes = ["http", "https"]
+            if uri.scheme not in supported_schemes:
+                raise ValueError(f"The scheme '{uri.scheme}' must be one of {supported_schemes}")
+
+        # Set the parameters
+        params = {}
```
Honny1, line 202: I would also add these params: 

```python
if os_name := kwargs.get("os"):
    params["OS"] = os_name
if architecture := kwargs.get("architecture"):
    params["Architecture"] = architecture
if variant := kwargs.get("variant"):
    params["Variant"] = variant
```

podman/domain/images_manager.py
```diff
@@ -158,6 +159,81 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+        if reference := kwargs.get("reference"):
+            params["reference"] = reference
+        if message := kwargs.get("message"):
+            params["message"] = message
+        if changes := kwargs.get("changes"):
+            params["changes"] = changes  # requests sends repeated keys as a list
+        if url:
+            params["url"] = url
```
Honny1, line 210: API expect `URL`.
```suggestion
            params["URL"] = url
```

podman/domain/images_manager.py
```diff
@@ -158,6 +159,81 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+            changes: Optional list of Dockerfile-style instructions
+                     (e.g. ['CMD /bin/bash', 'ENV FOO=bar']).
+
+        Returns:
+            An Image object for the newly imported image.
+
+        Raises:
+            APIError: when service returns an error.
+        """
+        # Check that exactly one of the data or file_path is provided
```
Honny1, line 188:
```suggestion
        # Check that exactly one source is provided
```

podman/domain/images_manager.py
```diff
@@ -158,6 +159,81 @@ def _generator(body: dict) -> Generator[Image, None, None]:
...
+                Path(file_path).open("rb")
+                if file_path is not None
+                else (io.BytesIO(data) if data is not None else nullcontext()) as post_data
+        ):
+
+            response = self.api.post(
+                "/images/import",
+                params=params,
+                data=post_data,
+                headers={"Content-Type": "application/x-tar"},
```
Honny1, line 223: I would not send `Content-Type: application/x-tar` when using URL.

podman/domain/images_manager.py
Andre-85: Sorry, slipped through. I fixed it.

podman/domain/images_manager.py
Andre-85: To be honest, I don't know what you are referring to. In the API documentation (https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImageTreeLibpod) these parameters are not named. Can you provide link to your information?

podman/domain/images_manager.py
Andre-85: Same here. To what documentation you are referring to? I used https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImageTreeLibpod

podman/domain/images_manager.py
Andre-85: I fixed that.

podman/domain/images_manager.py
Andre-85: Referring to documentation (https://docs.podman.io/en/latest/_static/api.html#tag/images/operation/ImageImportLibpod) we should not do that. It says "Request Body schema: application/x-tar
required", so i would say leaving "application/x-tar" out in the case a url is used is off-spec (But weird, I agree)

Andre-85:
I fixed/commented all issues and squashed in one commit.

podman/domain/images_manager.py
Honny1: Implementation: https://github.com/podman-container-tools/podman/blob/f19c577355dd5fab3411aee8d12b47da61f2237b/pkg/api/handlers/libpod/images.go#L431

podman/domain/images_manager.py
Honny1: https://github.com/podman-container-tools/podman/blob/f19c577355dd5fab3411aee8d12b47da61f2237b/pkg/api/handlers/libpod/images.go

podman/tests/integration/test_images.py
Honny1: @Andre-85

PR #642, opened 2026-04-08
Add ssh parameter support for image builds

### LINKED ISSUE #602: [Bug] Pass --secret option when building an image
> Hi, thank you very much for taking the time to look into this!
> 
> Confirmed your findings. My wrong results came from an old Podman (v3.4.4) that supported `--secret` locally but not remotely.
> Updated to Podman `v5.6.2` — secrets are now correctly passed via API.
> 
> ---
> 
> **Question:** Why does your POST request use `v5.7.0` in the URL if the latest release is `v5.6.2`?
> Is it related to `podman-remote`?
> Are you using a dev build?
> 
> ---
> 
> Should I open a new issue to discuss `--ssh`?
> 
> ---
> 
> In any case, I'm interested in opening a PR for this.
> If you don't mind, I will tackle it as soon as I can.
>  

 _Originally posted by @gsilva00 in [#526](https://github.com/containers/podman-py/issues/526#issuecomment-3402928399)_

---

This issue is to track the implementation of `ssh` to `images.build()`

sowmya-sl:
Allow passing SSH agent sockets or keys to the build API, similar to the existing secrets parameter support.
Fixes: https://github.com/containers/podman-py/issues/602

### DIFF
--- a/podman/domain/images_build.py
+++ b/podman/domain/images_build.py
@@ -70,6 +70,7 @@ class BuildMixin:
             manifest (str) - add the image to the specified manifest list.
                 Creates manifest list if it does not exist.
             secrets (list[str]) - Secret files/envs to expose to the build
+            ssh (list[str]) - SSH agent socket or keys to expose to the build
 
         Returns:
             first item is the podman.domain.images.Image built
@@ -215,6 +216,9 @@ class BuildMixin:
 
         if "secrets" in kwargs:
             params["secrets"] = json.dumps(kwargs.get("secrets"))
+
+        if "ssh" in kwargs:
+            params["ssh"] = json.dumps(kwargs.get("ssh"))
 
         # Remove any unset parameters
         return dict(filter(lambda i: i[1] is not None, params.items()))
--- a/podman/domain/images_build.py
+++ b/podman/domain/images_build.py
@@ -70,6 +70,9 @@ class BuildMixin:
             manifest (str) - add the image to the specified manifest list.
                 Creates manifest list if it does not exist.
             secrets (list[str]) - Secret files/envs to expose to the build
+            ssh (list[str]) - SSH agent socket or keys to expose to the build.
+                Format is the same as ``podman build --ssh``, e.g.
+                ``["default"]`` or ``["src=/path/to/key"]``.
 
         Returns:
             first item is the podman.domain.images.Image built
@@ -215,6 +218,9 @@ class BuildMixin:
 
         if "secrets" in kwargs:
             params["secrets"] = json.dumps(kwargs.get("secrets"))
+
+        if "ssh" in kwargs and kwargs.get("ssh") != []:
+            params["ssh"] = json.dumps(kwargs["ssh"])
 
         # Remove any unset parameters
         return dict(filter(lambda i: i[1] is not None, params.items()))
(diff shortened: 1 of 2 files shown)


inknos:
@sowmya-sl thanks for the PR, please sign your commit as DCO bot says

inknos (commented):
Apologies for the noise. I think the PR looks good but something might be missing on Podman's side and an issue could be filed to add ssh parameter there too. I'll investigate that while tests run

sowmya-sl:
@inknos Let me know if anything needs to be added from the server side.

podman/domain/images_build.py
```diff
@@ -70,6 +70,7 @@ def build(self, **kwargs) -> tuple[Image, Iterator[bytes]]:
             manifest (str) - add the image to the specified manifest list.
                 Creates manifest list if it does not exist.
             secrets (list[str]) - Secret files/envs to expose to the build
+            ssh (list[str]) - SSH agent socket or keys to expose to the build
```
Honny1: I would add an example of how content should look, or refer that content should look like `podman build --ssh`.

podman/domain/images_build.py
```diff
@@ -216,5 +217,8 @@ def _render_params(kwargs) -> dict[str, list[Any]]:
         if "secrets" in kwargs:
             params["secrets"] = json.dumps(kwargs.get("secrets"))
 
+        if "ssh" in kwargs:
+            params["ssh"] = json.dumps(kwargs.get("ssh"))
```
Honny1: Non-blocking:  What if `kwargs.get("ssh")` returns an empty array?

podman/domain/images_build.py
sowmya-sl: Added

inknos:
@sowmya-sl thanks

@Honny1 I think a change is needed in podman like [this](https://github.com/inknos/podman/commit/751ff55d1b3454aa1f7b2059c7da0787794e765e). it's just a POC, what do you think?

Honny1:
> @sowmya-sl thanks
> 
> @Honny1 I think a change is needed in podman like [this](https://github.com/inknos/podman/commit/751ff55d1b3454aa1f7b2059c7da0787794e765e). it's just a POC, what do you think?

I think yes.

podman/domain/images_build.py
```diff
@@ -70,6 +70,9 @@ def build(self, **kwargs) -> tuple[Image, Iterator[bytes]]:
             manifest (str) - add the image to the specified manifest list.
                 Creates manifest list if it does not exist.
             secrets (list[str]) - Secret files/envs to expose to the build
+            ssh (list[str]) - SSH agent socket or keys to expose to the build.
+                Format is the same as ``podman build --ssh``, e.g.
+                ``["default"]`` or ``["src=/path/to/key"]``.
```
Honny1, line 75:
```suggestion
                ``["default"]`` or ``["mykey=/path/to/key"]``.
```

https://docs.podman.io/en/latest/markdown/podman-build.1.html#ssh-default-id-socket
specifies the format as `default | id[=socket]`, so valid examples are `"default"`, `"default=/path/to/socket"`, or `"mykey=/path/to/key"`.

Honny1 (approved):
LGTM, Only docstring nit.

Honny1:
@inknos Any update on podman side?

PR #643, opened 2026-04-09
fix: auto-pull image in run() when Podman returns 500 'image not known'

### LINKED ISSUE #594: [BUG] run doesn't pull image automatically
client.containers.run and create are behaving differently.

example code:
```python
self.client.containers.run(image="docker.io/alpine/openssl:3.5.3", name="registry_cert_gen",
                            remove=True,
                            command=command)

self.client.containers.create(image="docker.io/alpine/openssl:3.5.3", name="registry_cert_gen",
                            command=command)

```

When run is used, I get the following:
```
    output = self.client.containers.run(image="docker.io/alpine/openssl:3.5.3", name="registry_cert_gen",
                                volumes=volumes,
    ...<2 lines>...
                                user="root",
                                command=command)
  File "/home/pixii/.local/lib/python3.13/site-packages/podman/domain/containers_run.py", line 74, in run
    container = self.create(image=image, command=command, **kwargs)
  File "/home/pixii/.local/lib/python3.13/site-packages/podman/domain/containers_create.py", line 382, in create
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/pixii/.local/lib/python3.13/site-packages/podman/api/client.py", line 82, in raise_for_status
    raise APIError(cause, response=self._response, explanation=message)
podman.errors.exceptions.APIError: 500 Server Error: Internal Server Error (docker.io/alpine/openssl:3.5.3: image not known)

```

When create is used, the image is pulled automatically without any issues. 

Podman 4.3.1
podman             5.6.0

rahulkini31:
When an image is not present locally, Podman returns HTTP 500 with "image not known" rather than HTTP 404. This caused run() to raise APIError instead of triggering the existing ImageNotFound catch block, so the auto-pull never happened.

Extend the except clause to also catch APIError, re-raising it unless the message indicates a missing image. Add a unit test covering this path.

Closes #594

### DIFF
--- a/podman/domain/containers_run.py
+++ b/podman/domain/containers_run.py
@@ -8,7 +8,7 @@ from collections.abc import Generator, Iterator
 
 from podman.domain.containers import Container
 from podman.domain.images import Image
-from podman.errors import ContainerError, ImageNotFound
+from podman.errors import APIError, ContainerError, ImageNotFound
 
 logger = logging.getLogger("podman.containers")
 
@@ -72,9 +72,19 @@ class RunMixin:  # pylint: disable=too-few-public-methods
         if isinstance(command, str):
             command = [command]
 
+        _needs_pull = False
         try:
             container = self.create(image=image_id, command=command, **kwargs)  # type: ignore[attr-defined]
         except ImageNotFound:
+            _needs_pull = True
+        except APIError as e:
+            # Podman may return HTTP 500 with "image not known" instead of 404
+            # when an image is missing locally; re-raise any unrelated API errors.
+            if "image not known" not in str(e):
+                raise
+            _needs_pull = True
+
+        if _needs_pull:
             self.podman_client.images.pull(  # type: ignore[attr-defined]
                 image_id,
                 auth_config=kwargs.get("auth_config"),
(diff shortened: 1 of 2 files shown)


inknos:
@rahulkini31 
thanks for the pr. please, add your signature to the commit with `git commit -s`. Also, address pre-commit. Happy to help with questions.

podman/domain/containers_run.py
```diff
@@ -74,7 +74,9 @@ def run(
 
         try:
             container = self.create(image=image_id, command=command, **kwargs)  # type: ignore[attr-defined]
-        except ImageNotFound:
+        except (ImageNotFound, APIError) as e:
+            if isinstance(e, APIError) and "image not known" not in str(e):
+                raise
```
inknos: I don't understand this code. here you are ignoring all the ApiErrors, unless they contain "image not known". would you help me understand what's the goal?

inknos (changes requested):

podman/domain/containers_run.py
rahulkini31: Thanks for reviewing so quickly. Podman returns HTTP 500 (not 404) when an image is missing locally, so raise_for_status raises APIError instead of ImageNotFound. The original 'ImageNotFound' block never fired in this case. I've refactored to use separate except clauses with a comment explaining why APIError needs to be handled — hopefully clearer now. The mypy failure in utils.py is pre-existing and unrelated to this change.

TomSweeneyRedHat:
@rahulkini31 this one needs a rebase

Honny1:
@rahulkini31 Any updates on this?

PR #644, opened 2026-05-13
Fix exec_run() returning empty output over SSH connections

### LINKED ISSUE #506: [BUG] empty stdout/stderr for `exec_run` on macOS
Hi, I'm running Podman Machine on a MacBook and everything works fine except the `exec_run`, the output is always empty. Below an example which works on Linux machines:

```python
>>> import podman
>>> client = podman.PodmanClient()
>>> client.images.list()
[<Image: 'docker.io/library/alpine:latest'>]
>>> client.version()
{'Platform': {'Name': 'linux/arm64/fedora-41'}, 'Components': [{'Name': 'Podman Engine', 'Version': '5.4.0-dev-234c41c80', 'Details': {'APIVersion': '5.4.0-dev-234c41c80', 'Arch': 'arm64', 'BuildTime': '2025-01-22T00:00:00Z', 'Experimental': 'false', 'GitCommit': '', 'GoVersion': 'go1.23.4', 'KernelVersion': '6.12.7-200.fc41.aarch64', 'MinAPIVersion': '4.0.0', 'Os': 'linux'}}, {'Name': 'Conmon', 'Version': 'conmon version 2.1.12, commit: ', 'Details': {'Package': 'conmon-2.1.12-3.fc41.aarch64'}}, {'Name': 'OCI Runtime (crun)', 'Version': 'crun version UNKNOWN\ncommit: b143d9ea7cfea9de44222d7fcfb7f55c21ba8060\nrundir: /run/user/501/crun\nspec: 1.0.0\n+SYSTEMD +SELINUX +APPARMOR +CAP +SECCOMP +EBPF +CRIU +LIBKRUN +WASM:wasmedge +YAJL', 'Details': {'Package': 'crun-1.19.1-1.20250121161049663014.main.23.ga2d140d.fc41.aarch64'}}], 'Version': '5.4.0-dev-234c41c80', 'ApiVersion': '1.41', 'MinAPIVersion': '1.24', 'GitCommit': '', 'GoVersion': 'go1.23.4', 'Os': 'linux', 'Arch': 'arm64', 'KernelVersion': '6.12.7-200.fc41.aarch64', 'BuildTime': '2025-01-22T00:00:00Z'}
>>> container = client.containers.create("alpine", command=["sleep", "1000"])
>>> container.start()
>>> ret, out = container.exec_run("echo 42", demux=True)
>>> ret
0
>>> out
(b'', b'')
```

Any ideas?

maxammann:
SSHSocket.send()/recv() were writing to subprocess pipes instead of the real forwarded unix socket established by connect(). Additionally, exec_run() now sends Connection: Upgrade headers so the server returns a 101 response, allowing multiplexed output to be read directly from the raw socket. Without the upgrade, http.client closes the connection before the body can be read because no Content-Length or Transfer-Encoding is set.

Fixes #506

This debugging was heavily Claude supported. In the end after it figured it out I manually provided the correct and minimal fixes.

### DIFF
--- a/podman/api/client.py
+++ b/podman/api/client.py
@@ -460,3 +460,22 @@ class APIClient(requests.Session):
             )
         except OSError as e:
             raise APIError(uri.geturl(), explanation=f"{method.upper()} operation failed") from e
+
+    def _get_raw_response_socket(self, response):
+        """Extract the raw socket from an HTTP response.
+
+        After a 101 Upgrade, data flows directly on the socket outside HTTP
+        framing. This method navigates the urllib3/http.client internals to
+        reach the underlying transport socket.
+        """
+        if self.base_url.scheme == "http+ssh":
+            sock = response.raw._fp.fp.raw._sock
+        else:
+            sock = response.raw._fp.fp.raw
+        try:
+            # Keep a reference to the response to prevent garbage collection
+            # from closing the socket prematurely.
+            sock._response = response
+        except AttributeError:
+            pass
+        return sock
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -2,7 +2,9 @@
 
 import json
 import logging
+import select
 import shlex
+import struct
 from collections.abc import Iterable, Iterator, Mapping
 from contextlib import suppress
 from typing import Any, Optional, Union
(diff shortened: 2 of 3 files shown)


maxammann:
@jwhonce tyi this should fix podman on macOS as its using SSH connections there

podman/api/client.py
```diff
@@ -460,3 +460,22 @@ def _request(
...
+
+    def _get_raw_response_socket(self, response):
+        """Extract the raw socket from an HTTP response.
+
+        After a 101 Upgrade, data flows directly on the socket outside HTTP
+        framing. This method navigates the urllib3/http.client internals to
+        reach the underlying transport socket.
+        """
+        if self.base_url.scheme == "http+ssh":
+            sock = response.raw._fp.fp.raw._sock
```
Honny1, line 472: I don't like to see using that many private attributes.

podman/domain/containers.py
```diff
@@ -19,6 +21,65 @@
...
+    data = b""
+    while len(data) < n:
+        chunk = _read_from_socket(sock, n - len(data))
+        if not chunk:
+            return data
+        data += chunk
+    return data
+
+
+def _frames_iter(sock):
```
Honny1, line 49: `_frames_iter()` and `_consume_frames()` duplicate logic already in `podman/api/output_utils.py` (`demux_output()`) and `api.stream_frames()`.

podman/domain/containers.py
```diff
@@ -19,6 +21,65 @@
 logger = logging.getLogger("podman.containers")
 
 
+def _read_from_socket(sock, n=4096):
```
Honny1, line 24: `_read_from_socket()`, `_read_exactly()`, `_frames_iter()`, `_consume_frames()` are transport/protocol utilities. They belong in `podman/api/output_utils.py` or a similar shared module, not in the domain model file.

podman/domain/containers.py
```diff
@@ -19,6 +21,65 @@
...
+    """Read at most n bytes from a socket, using select to wait for data."""
+    if not hasattr(select, "poll"):
+        ready, _, _ = select.select([sock], [], [], 5.0)
+        if not ready:
+            return b""
+    else:
+        poll = select.poll()
+        poll.register(sock, select.POLLIN | select.POLLPRI)
+        if not poll.poll(5000):
```
Honny1, line 33: I don't like a hardcoded 5-second timeout.

podman/domain/containers.py
```diff
@@ -211,21 +272,48 @@ def exec_run(
         response.raise_for_status()
         exec_id = response.json()['Id']
         # start the exec instance, this will store command output
+        # Upgrade headers are required for SSH connections: without them the
```
Honny1, line 275: Changes in `exec_run()` always send Upgrade headers and always expect a 101 response. even for non-SSH connections where the existing `start_resp.content` path works fine today. The fallback else branch (lines for non-101) is good defensive coding, but the default behavior should not change for connections that work correctly today. After rebasing, the Upgrade+raw-socket path should be gated on self.client.base_url.scheme == "http+ssh" for the non-socket case.

Honny1 (changes requested):
Thanks for your contribution. I have some comments. Please rebase onto the main upstream.

Honny1:
Any update on this, @maxammann?

PR #647, opened 2026-06-08
Fix healthcheck kwargs for container.create()

inknos:
The health_* kwargs were being passed as snake_case in the JSON payload, but /libpod/containers/create endpoint expects nested objects.

Assemble kwargs into proper structures:
- healthconfig: {Test, Interval, Timeout, Retries, StartPeriod}
- startupHealthConfig: {Test, Interval, Timeout, Retries, Successes}
- healthLogDestination, healthMaxLogCount, healthMaxLogSize (camelCase)
- health_on_failure string mapped to health_check_on_failure_action int

Also, it's possible to pass duration values as go-style strings ("30s", "1m", "500ms") and integers as nanoseconds.

Fixes: https://redhat.atlassian.net/browse/RHEL-140716

---

keeping as draft since I want to sleep on the code before saying it's ready

### DIFF
--- a/podman/api/parse_utils.py
+++ b/podman/api/parse_utils.py
@@ -3,6 +3,7 @@
 import base64
 import ipaddress
 import json
+import re
 import struct
 from datetime import datetime, timezone
 from typing import Any, Optional, Union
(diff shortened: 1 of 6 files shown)


giuseppe (approved):
LGTM

podman/api/parse_utils.py
```diff
@@ -11,6 +12,65 @@
...
+    pos = 0
+    while pos < len(remaining):
+        match = pattern.match(remaining, pos)
+        if not match:
+            raise ValueError(
+                f"Invalid duration format: {value!r}. "
+                f"Use Go-style durations like '30s', '1m', '500ms', '1h30m'."
+            )
+        num_str, unit = match.groups()
+        total_ns += int(float(num_str) * DURATION_UNITS_NS[unit])
```
Honny1: Well, using `float` feels odd. Why not `int`?

Honny1: (`"1"` -> `1.0000000001` could happen)

podman/api/parse_utils.py
```diff
@@ -11,6 +12,65 @@
...
+    "none": 0,
+    # value 1 is "invalid" in the Go enum (HealthCheckOnFailureActionInvalid)
+    # and should not be used directly
+    "kill": 2,
+    "restart": 3,
+    "stop": 4,
+}
+
+
+def prepare_duration_ns(value: Union[int, str, None]) -> Union[int, None]:
```
Honny1, line 35: What about `prepare_duration_ns("0")`?

podman/domain/containers_create.py
```diff
@@ -823,15 +835,90 @@ def parse_host_port(_container_port, _protocol, _host):
...
-            params["no_healthcheck"] = args.pop("no_healthcheck", None)
-        else:
-            for hc in set(args) & health_commands_data:
-                params[hc] = args.pop(hc, None)
+        present_health_kwargs = set(args) & health_kwargs
+        no_healthcheck = args.pop("no_healthcheck", None)
+
+        if no_healthcheck and present_health_kwargs:
+            raise ValueError(
+                f"Cannot set {sorted(present_health_kwargs)[0]} when no_healthcheck is True"
```
Honny1: Non-blocking: I don't like indexing here. But it is okay.

podman/tests/integration/test_container_create.py
```diff
@@ -262,6 +262,39 @@ def test_container_healthchecks(self):
         container = self.client.containers.create(self.alpine_image, **parameters)
         self.containers.append(container)
 
+    def test_container_healthcheck_fine_grained_kwargs(self):
```
Honny1, line 265: I would also add a test for default values: `health_on_failure="none"`.

Honny1 (commented):
I have some comments. Moust of them are small improvments.

podman/api/parse_utils.py
inknos: I was thinking of "1.5s" being valid but we could do integers only. I see the validation is wrong anyway so doing ints seems ok. WDYT?

podman/api/parse_utils.py
inknos: nice I'll fix it

podman/tests/integration/test_container_create.py
inknos: right

Honny1 (approved):
LGTM

Honny1:
@inknos, can you please try to rebase on upstream main? CI failure seems to be unrelated.

PR #648, opened 2026-07-13
Implement socket=True in Container.exec_run via connection hijacking

### LINKED ISSUE #421: exec_run(cmd, socket=True) isn't implemented: blocks and only returns resulting stdout/stderr bytes
As per [docs](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.exec_run) and docker API examples, `c.exec_run(cmd, socket=True)` should return immediately with socket to be used for process' stdin/out/err, . However, podman-py waits for the process to finish and it only returns final stdout/err bytes as the output.

Fiddling with `stream` and `tty` parameters doesn't make any change in the behaviour.

Given that googling for `"PodmanClient" "exec_run" "socket=True"` or `"PodmanClient" "exec_run" "stdin=True"` yields zero results in both cases, maybe it's fine that `socket=True` isn't implemented, but then `socket`, `stdin` and `stream` parameters to `exec_run()` shouldn't be mentioned in the docs at all?

### versions
* Fedora 40: python3-podman-5.2.0-1.fc40.noarch
* RHEL 9: python3-podman-5.2.0-1.el9.noarch

#### just outputs
```python
>>> pc = PodmanClient(base_url=podman_sock)
>>> c = pc.containers.get(cont_name)
>>> res = c.exec_run(["bash", "-c", "echo -n 'hello'; sleep 10; echo 'world!'"], socket=True); type(res[1])
```
Actual results:
* `type(res[1])` output only appears after 10 s
* the type of `res[1]` is `<class 'bytes'>`

Expected result: `res` is returned right away and its `output` field is socket object for further I/O

#### with stdin
```python
>>> res = c.exec_run("bash", socket=True, stdin=True); type(res[1])
exit<enter>
```
Expected results: `bash` receives the exit command and ends itself

Actual results:
* bash never receives the `exit` command
* API waits for data that can't come
* no clean way to get out of the wait, keyboard interrupt in an interactive intepreter produces this traceback (on RHEL 9):
    ```python
    >>> sock = c.exec_run("bash", stdin=True, socket=True, tty=False)
    exit

    ^CTraceback (most recent call last):
      File "<stdin>", line 1, in <module>
      File "/usr/lib/python3.9/site-packages/podman/domain/containers.py", line 198, in exec_run
        start_resp = self.client.post(
      File "/usr/lib/python3.9/site-packages/podman/api/client.py", line 327, in post
        return self._request(
      File "/usr/lib/python3.9/site-packages/podman/api/client.py", line 425, in _request
        self.request(
      File "/usr/lib/python3.9/site-packages/requests/sessions.py", line 544, in request
        resp = self.send(prep, **send_kwargs)
      File "/usr/lib/python3.9/site-packages/requests/sessions.py", line 699, in send
        r.content
      File "/usr/lib/python3.9/site-packages/requests/models.py", line 831, in content
        self._content = b''.join(self.iter_content(CONTENT_CHUNK_SIZE)) or b''
      File "/usr/lib/python3.9/site-packages/requests/models.py", line 753, in generate
        for chunk in self.raw.stream(chunk_size, decode_content=True):
      File "/usr/lib/python3.9/site-packages/urllib3/response.py", line 628, in stream
        data = self.read(amt=amt, decode_content=decode_content)
      File "/usr/lib/python3.9/site-packages/urllib3/response.py", line 567, in read
        data = self._fp_read(amt) if not fp_closed else b""
      File "/usr/lib/python3.9/site-packages/urllib3/response.py", line 533, in _fp_read
        return self._fp.read(amt) if amt is not None else self._fp.read()
      File "/usr/lib64/python3.9/http/client.py", line 463, in read
        n = self.readinto(b)
      File "/usr/lib64/python3.9/http/client.py", line 507, in readinto
        n = self.fp.readinto(b)
      File "/usr/lib64/python3.9/socket.py", line 704, in readinto
        return self._sock.recv_into(b)
    KeyboardInterrupt
    ````

oligiochi:
The libpod REST API supports upgrading the exec start request to a raw bidirectional stream (101 Switching Protocols). Send the Upgrade headers, keep the response unconsumed (stream=True) and hand the caller the underlying UDS socket, anchoring the response to keep the hijacked connection out of the urllib3 pool.

Fixes #421

Signed-off-by: Giovanni Oliverio oligiovi2@gmail.com

Fixes #421

## Problem

`Container.exec_run()` accepts a `socket` parameter (documented as "Return
the connection socket to allow custom read/write operations") but the
parameter is silently ignored: the call always blocks until the command
exits and returns the buffered output. This makes interactive exec
sessions (a shell with live stdin/stdout, terminal emulators, REPLs)
impossible with the native bindings, while the same use case works with
docker-py against Podman's compatibility socket.

The limitation is client-side only: the libpod REST API fully supports
hijacking the exec start request into a raw bidirectional stream.

## Solution

When `socket=True`, send the `POST /exec/{id}/start` request with
`Connection: Upgrade` / `Upgrade: tcp` headers and `stream=True` (so
`requests` does not consume the body). The server replies
`101 Switching Protocols` and the underlying connection stops being HTTP:
it becomes a raw bidirectional byte channel attached to the exec
session's stdin/stdout/stderr (raw stream with `tty=True`, multiplexed
frames otherwise). The socket is extracted from the response via
`response.raw._connection.sock` — `_connection` here is podman-py's own
`UDSConnection` (podman/api/uds.py), which keeps a reference to its
socket, so the only private attribute traversed is urllib3's
`HTTPResponse._connection`.

The response object is anchored on the returned socket
(`sock._hijacked_response`) so the hijacked connection cannot be released
back to the urllib3 pool while the caller is using it; a later request on
the same client would otherwise write HTTP into the exec session.

`Detach` is forced to `False` in this branch: a socket to a detached
session is meaningless (consistent with docker-py behaviour).

Return contract matches docker-py: `(None, socket)`; the caller is
responsible for reading, writing and closing the socket.

## Testing

- Unit test: verifies that `socket=True` issues the start request with
  the Upgrade headers, `stream=True` and `Detach: false`, and returns the
  extracted socket (extraction helper mocked, since the hijack cannot
  happen against requests_mock).
- Integration test: opens an interactive `/bin/sh` via
  `exec_run(socket=True)`, writes a command computing a marker and reads
  it back from the live socket.
- Manually validated against Podman 4.9.3 (rootless, crun, UDS) with
  podman-py 5.8.0: upgrade returns 101, the channel is bidirectional
  while the command is still running, terminal resize via
  `/exec/{id}/resize` is honoured, exit codes are reported by
  `/exec/{id}/json`, and concurrent requests on the same client open a
  separate pool connection without corrupting the hijacked stream.

### DIFF
--- a/podman/domain/containers.py
+++ b/podman/domain/containers.py
@@ -145,7 +145,7 @@ class Container(PodmanResource):
         user=None,
         detach: bool = False,
         stream: bool = False,
-        socket: bool = False,  # pylint: disable=unused-argument
+        socket: bool = False,
         environment: Union[Mapping[str, str], list[str]] = None,
         workdir: str = None,
         demux: bool = False,
@@ -182,9 +182,12 @@ class Container(PodmanResource):
                 If ``stream``, then a generator yielding response chunks.
                 If ``demux``, then a tuple of (``stdout``, ``stderr``).
                 Else the response content.
+                If ``socket``, then ``(None, socket)`` where ``socket`` is the
+                    raw bidirectional connection to the exec session
+                    (``stdin``/``stdout``/``stderr``). The caller is responsible
+                    for reading, writing and closing it.
 
         Raises:
-            NotImplementedError: method not implemented.
             APIError: when service reports error
         """
         # pylint: disable-msg=too-many-locals
@@ -210,6 +213,37 @@ class Container(PodmanResource):
         response = self.api.post(f"/containers/{self.name}/exec", data=json.dumps(data))
         response.raise_for_status()
         exec_id = response.json()['Id']
+        if socket:
+            # Hijack the connection: request a protocol upgrade on the exec
+            # start endpoint. After the "101 Switching Protocols" response the
+            # underlying UDS connection stops being HTTP and becomes a raw
+            # bidirectional byte channel attached to the exec session's
+            # stdin/stdout/stderr (raw stream with tty=True, multiplexed
+            # frames otherwise).
+            if self.client.base_url.scheme == "http+ssh":
+                raise NotImplementedError("exec_run(socket=True) is not supported over SSH")
+            start_resp = self.api.post(
+                f"/exec/{exec_id}/start",
+                data=json.dumps({"Detach": False, "Tty": tty}),
+                headers={"Connection": "Upgrade", "Upgrade": "tcp"},
+                stream=True,  # keep requests from consuming the stream body
+            )
+            start_resp.raise_for_status()
+            conn = start_resp.raw.connection
+            if conn is None or conn.sock is None:
+                raise APIError(
+                    f"/exec/{exec_id}/start",
+                    explanation="Unable to extract socket from hijacked connection",
+                    response=start_resp,
+                )
+            sock = conn.sock
+            # Anchor the response on the socket: if start_resp were released
+            # or garbage collected, urllib3 could return the hijacked
+            # connection to the pool and a later request would write HTTP
+            # into the exec session.
+            sock._hijacked_response = start_resp
+            return None, sock
+
         # start the exec instance, this will store command output
         start_resp = self.api.post(
             f"/exec/{exec_id}/start", data=json.dumps({"Detach": detach, "Tty": tty}), stream=stream
--- a/podman/tests/integration/test_container_exec.py
+++ b/podman/tests/integration/test_container_exec.py
@@ -120,3 +120,26 @@ class ContainersExecIntegrationTests(base.IntegrationTest):
             output,
             b'\n',
         )
+
+    def test_container_exec_run_socket(self):
+        """exec_run(socket=True) returns a live bidirectional socket."""
+        container = self.client.containers.create(self.alpine_image, command=["top"], detach=True)
+        self.containers.append(container)
+        container.start()
+
+        rc, sock = container.exec_run("/bin/sh", stdin=True, tty=True, socket=True)
+        self.assertIsNone(rc)
+
+        try:
+            sock.settimeout(5)
+            sock.sendall(b"echo $((100+11))\n")
+            data = b""
+            while b"111" not in data:
+                # the echo contains "100+11", never the literal "111":
+                # a computed marker distinguishes output from input echo
+                chunk = sock.recv(4096)
+                if not chunk:
+                    self.fail("socket closed before the marker was seen")
+                data += chunk
+        finally:
+            sock.close()
--- a/podman/tests/unit/test_container.py
+++ b/podman/tests/unit/test_container.py
@@ -2,6 +2,7 @@ import base64
 import io
 import json
 import unittest
+from unittest.mock import MagicMock
 
 try:
     # Python >= 3.10
@@ -16,6 +17,7 @@ from podman import PodmanClient, tests
 from podman.domain.containers import Container
 from podman.domain.containers_manager import ContainersManager
 from podman.errors import APIError, NotFound
+from unittest.mock import patch
 
 FIRST_CONTAINER = {
     "Id": "87e1325c82424e49a00abdd4de08009eb76c7de8d228426a9b8af9318ced5ecd",
@@ -472,6 +474,64 @@ class ContainersTestCase(unittest.TestCase):
             self.assertDictEqual(response, actual)
 
         self.assertTrue(adapter.called_once)
+
+    def test_exec_run_socket(self):
+        """exec_run(socket=True) sends the upgrade request and returns the hijacked socket."""
+        exec_resp = MagicMock()
+        exec_resp.raise_for_status.return_value = None
+        exec_resp.json.return_value = {"Id": "exec1"}
+
+        fake_sock = MagicMock()
+
+        start_resp = MagicMock()
+        start_resp.raise_for_status.return_value = None
+        start_resp.raw.connection.sock = fake_sock
+
+        container = Container(attrs=FIRST_CONTAINER, client=self.client.api)
+
+        with (
+            patch.object(container.api, "post") as post,
+            patch.object(container.api, "get") as get,
+        ):
+            post.side_effect = [exec_resp, start_resp]
+
+            rc, sock = container.exec_run(
+                "/bin/bash",
+                stdin=True,
+                tty=True,
+                socket=True,
+            )
+
+        # return contract: (None, socket), matching docker-py
+        self.assertIsNone(rc)
+        self.assertIs(sock, fake_sock)
+        # the response is anchored on the socket to keep the hijacked
+        # connection out of the urllib3 pool
+        self.assertIs(fake_sock._hijacked_response, start_resp)
+
+        # exactly two POSTs (exec create + exec start), nothing else:
+        # the socket branch must return immediately, without the
+        # blocking start/inspect path
+        self.assertEqual(post.call_count, 2)
+        get.assert_not_called()
+
+        # first call: exec create
+        self.assertEqual(
+            post.call_args_list[0].args[0],
+            f"/containers/{container.name}/exec",
+        )
+
+        # second call: start with connection upgrade, unconsumed body
+        self.assertEqual(post.call_args_list[1].args[0], "/exec/exec1/start")
+        self.assertEqual(
+            post.call_args_list[1].kwargs["headers"],
+            {"Connection": "Upgrade", "Upgrade": "tcp"},
+        )
+        self.assertTrue(post.call_args_list[1].kwargs["stream"])
+        self.assertEqual(
+            json.loads(post.call_args_list[1].kwargs["data"]),
+            {"Detach": False, "Tty": True},
+        )
 
 
 if __name__ == '__main__':


podman/domain/containers.py
```diff
@@ -210,6 +213,29 @@ def exec_run(
...
+            # frames otherwise).
+            start_resp = self.api.post(
+                f"/exec/{exec_id}/start",
+                data=json.dumps({"Detach": False, "Tty": tty}),
+                headers={"Connection": "Upgrade", "Upgrade": "tcp"},
+                stream=True,  # keep requests from consuming the stream body
+            )
+            start_resp.raise_for_status()
+
+            sock = start_resp.raw._connection.sock
```
inknos: I believe you need `raw.connection` instead of `_connection`

inknos: also, I believe we need a `None` guard here in case the response is consumed or the socket is `None` for some reason

```pyhton
conn = start_resp.raw.connection
if conn is None or conn.sock is None:
    raise RuntimeError("unable to extract socket from hijacked connection")
sock = conn.sock
```

not sure about the message, but I would avoid throwing a python error.

inknos (commented):
thanks for the PR. I saw you converted to draft so I'll stop reviewing, but here are a couple of comments.

I am also thinking if it will break if the socket is passed via ssh. The reason is, when you hijack the socket over http you can get a clean file descriptor for read/write operations. But when the connection is handled via ssh the python code overrides the read/write methods and you pipe to different ssh subprocess. This will give you different file descriptors and it will break.

So the SSH case needs to be handled with care. A quick solution would be to raise `NotImplementedError` for ssh connections to keep it simple.

podman/domain/containers.py
oligiochi: Thanks for the review! I've addressed all three points:

1. Fixed raw._connection → raw.connection (public property instead of the private urllib3 attribute).
2. Added a None guard before extracting the socket:
```python 
conn = start_resp.raw.connection
if conn is None or conn.sock is None:
    raise APIError(
        f"/exec/{exec_id}/start",
        explanation="Unable to extract socket from hijacked connection",
        response=start_resp,
    )
sock = conn.sock
```
Used APIError instead of a bare Python exception, consistent with the rest of the codebase.

3. Added an explicit guard for SSH connections, raising NotImplementedError:
```python 
if self.client.api.base_url.scheme == "http+ssh":
    raise NotImplementedError("exec_run(socket=True) is not supported over SSH")
```
You're right that hijacking wouldn't work cleanly there since the socket ends up piped through an SSH subprocess rather than being a direct fd. Happy to look into proper SSH support as a follow-up if there's interest, but wanted to keep this PR focused on the UDS case.
Ready for another look whenever you have time.

inknos: > Used APIError

Makes sense

> if self.client.api.base_url.scheme == "http+ssh":
    raise NotImplementedError("exec_run(socket=True) is not supported over SSH")

Should be `if self.client.base_url.scheme`

> Happy to look into proper SSH support as a follow-up if there's interest, but wanted to keep this PR focused on the UDS case

definitely, it might be/not be so trivial, but let's make it another PR

I'll give it another round of review tomorrow

oligiochi: >Should be `if self.client.base_url.scheme`

Yes, I saw that and fixed it in the next commit, but thanks anyway.

oligiochi:
The testing-farm:fedora-rawhide-x86_64:distro-fedora-all job fails deterministically (retriggered via /packit test, same result), but I'm unable to reproduce it locally:

Full unit test suite passes on my machine (274 passed), with both pytest and unittest discover, including with the latest urllib3/requests/requests-mock
It also passes inside a fedora:rawhide container (Python 3.15), again with both runners

All other testing-farm targets (fedora 43/44, centos-stream 14/15, epel, unittest-coverage) pass. The truncated failure log only shows urllib3 exhausting retries (raise reraise(...) in urllib3/util/retry.py).
Is this a known issue with the rawhide plan, or is there something environment-specific I should look at? Happy to dig further if you can point me at the relevant log.

oligiochi:
# CI failure analysis: `test_ssh_ping` — unrelated to this PR

## TL;DR

The only failing test in the CI run is
`podman/tests/integration/test_adapters.py::AdapterIntegrationTest::test_ssh_ping`,
failing identically in all six tox environments (`coverage`, `py39`–`py313`).
The failure originates in the SSH port-forwarding layer of the CI runner
(sshd cannot connect to the podman service's UNIX socket), not in Python code,
and it occurs **before any code touched by this PR is ever executed**. All
331 other tests pass, including the two new tests added by this PR. Below is
the full evidence.

## What the failing run looks like

Environment: Podman **6.0.1** on the runner, `tox -e coverage,py39,py310,py311,py312,py313`.

Every environment reports the same single failure:

```
=========================== short test summary info ============================
FAILED podman/tests/integration/test_adapters.py::AdapterIntegrationTest::test_ssh_ping
=== 1 failed, 327 passed, 4 skipped, 88 subtests passed in 67.08s (0:01:07) ====
```

For comparison, an earlier run of the same tox matrix on a runner with Podman
**6.0.0** (July 1) passed everything, including `test_ssh_ping`:

```
======== 362 passed, 4 skipped, 74 subtests passed in 74.57s =========
  coverage: OK ... py313: OK
  congratulations :)
```

The only relevant difference between the two runs, besides the source tree,
is the runner image and its Podman version (6.0.0 → 6.0.1).

## Anatomy of the failure

`test_ssh_ping` starts a `podman system service` on a fresh UNIX socket under
`/tmp`, then reaches it through an SSH tunnel:

```python
def test_ssh_ping(self):
    with PodmanClient(
        base_url=f"http+ssh://{getpass.getuser()}@localhost:22{self.socket_file}"
    ) as client:
        self.assertTrue(client.ping())
```

Passing run (Podman 6.0.0):

```
Launching(6.0.0) podman ... system service --time=0 unix:///tmp/tmp35fi.../...
Starting new HTTP connection (1): localhost:22
Waiting on /run/user/0/podman/podman-forward-....sock   (x2)
http://localhost:22 "HEAD /v5.8.0/libpod/_ping HTTP/1.1" 200 0
PASSED
```

Failing run (Podman 6.0.1) — identical up to the request, then:

```
Launching(6.0.1) podman ... system service --time=0 unix:///tmp/tmp03r8.../...
Starting new HTTP connection (1): localhost:22
Waiting on /run/user/0/podman/podman-forward-....sock   (x2)
send: HEAD /v5.8.0/libpod/_ping ...
[ssh stderr]  channel 1: open failed: connect failed: open failed
FAILED
```

Resulting exception chain:

```
ConnectionResetError(104, 'Connection reset by peer')
  → requests.exceptions.ConnectionError ('Connection aborted.')
    → podman.errors.exceptions.APIError:
      http://root@localhost:22/v5.8.0/libpod/_ping (HEAD operation failed)
```

The key line is emitted by the **ssh client**, not by Python:

```
channel 1: open failed: connect failed: open failed
```

The client-side forward socket
(`/run/user/0/podman/podman-forward-*.sock`) is created and accepts the
connection exactly as in the passing run. The failure happens on the remote
end of the tunnel: when the HEAD request traverses the channel, **sshd fails
to connect to the target UNIX socket** (the podman service socket under
`/tmp`), closes the channel, and the client observes ECONNRESET. Note that
the `podman system service` process itself starts and exits cleanly
(`Command return Code: 0`), so the service side is healthy — it is the
sshd → UNIX-socket hop that breaks.

Plausible causes, in order of likelihood:

1. A behavior change between Podman 6.0.0 and 6.0.1 in `system service`
   socket creation (timing, path, or permissions) as seen by sshd.
2. A change in the runner image's sshd configuration (e.g. filesystem
   sandboxing such as `PrivateTmp`, which would make a socket under `/tmp`
   invisible to sshd-spawned channels).
3. A startup race — least likely, since the launch-to-request interval
   (~1 s) is the same in both runs.

## Why this PR cannot be the cause

1. **Execution order.** `test_adapters.py` sorts first, so `test_ssh_ping` is
   the very first test executed in each pytest process — it fails at `[ 0%]`,
   before any test added by this PR runs and before `exec_run()` is called
   even once in that process. There is no runtime state originating from this
   PR at the moment of failure.

2. **The changed code is unreachable from this test.** The production diff is
   confined to `Container.exec_run()`, and within it the new code runs only
   when `socket=True`. `test_ssh_ping` calls `client.ping()` →
   `HEAD /libpod/_ping`; `exec_run()` is never invoked. The PR adds no new
   imports and no module-level code, so there are no import-time side effects
   either.

3. **The error originates outside Python.** The `channel 1: open failed`
   message comes from sshd's failure to open a UNIX socket; no podman-py code
   (patched or otherwise) executes in that process.

4. **Deterministic across six fresh virtualenvs.** The failure is identical
   in `coverage` and `py39`–`py313`, each a freshly built environment, each
   failing at its first test — including the first environment of the run,
   i.e. a completely virgin process on the runner.

5. **No cross-process channel.** Each test launches its own
   `podman system service` on a fresh socket path; a hypothetical leaked
   resource from another test could only reference a different, already-dead
   socket and cannot affect sshd connecting to a new one. The new tests also
   clean up after themselves (`sock.close()` in a `finally` block,
   `remove(force=True)` in `tearDown`).

6. **Not a stale-branch artifact.** The SSH test and its infrastructure have
   not changed upstream since well before this branch's merge base
   (`test_adapters.py` last touched Nov 2024, `podman/api/ssh.py` Sep 2024,
   `tests/integration/{utils,base}.py` Mar 2026), so the branch is not
   missing any upstream fix relevant to this test.

Meanwhile, the tests that *do* exercise the PR pass everywhere:

- `podman/tests/unit/test_container.py::ContainersTestCase::test_exec_run_socket` → PASSED (all envs)
- `podman/tests/integration/test_container_exec.py::ContainersExecIntegrationTests::test_container_exec_run_socket` → PASSED (all envs), with the hijack visible in the logs:

```
POST /v5.8.0/libpod/containers/.../exec HTTP/1.1" 201
POST /v5.8.0/libpod/exec/.../start HTTP/1.1" 101 0    ← 101 Switching Protocols
```

## Suggested verification

Running `pytest podman/tests/integration/test_adapters.py` from a clean
checkout of `main` on the same runner (Podman 6.0.1) should reproduce the
same `test_ssh_ping` failure, confirming it is a runner/environment
regression independent of this branch. Happy to help bisect between Podman
6.0.0 and 6.0.1 or against the runner image if useful.

inknos:
your pr looks good, thanks!

also, **great** investigation on the failures, I will go and take a look at podman 6.0.1 changes.

inknos (approved):

inknos:
> I will go and take a look at podman 6.0.1 changes.

on second thought, I believe it could be even infrastructure related. I couldn't reproduce locally on my system. I'll try a rawhide box to check if it's rawhide related

oligiochi:
> on second thought, I believe it could be even infrastructure related. I couldn't reproduce locally on my system. I'll try a rawhide box to check if it's rawhide related

If it turns out to be infrastructure-related, I'd be happy to help investigate further.

inknos:
> If it turns out to be infrastructure-related, I'd be happy to help investigate further.

it is and it's related to SELinux policies in ssh. I was checking, and I am almost at the bottom of it, I'll share what I found.

I reproduced on a rawhide box, with SELinux enabled. If you set it to permissive you'll get the test passing with this line in the logs (or similar)

```
type=AVC msg=audit(1784552999.028:754): avc:  denied  { connectto } for  pid=3768 comm="sshd-session" path="/tmp/tmpu4f66z2l/72a732c0fd6244f1bbdfe32c3a4eed03" scontext=system_u:system_r:sshd_session_t:s0-s0:c0.c1023 tcontext=unconfined_u:unconfined_r:container_runtime_t:s0-s0:c0.c1023 tclass=unix_stream_socket permissive=0
```

I am not that fluent in SELinux, but I believe ssh forwardings run in the sshd-session binary, which Fedora's SELinux policy assigns to the new `sshd_session_t` domain. The [`containers/container-selinux`](https://github.com/containers/container-selinux) policies should be updated to grant `sshd_session_t` the connectto permission on `container_runtime_t` sockets.

oligiochi:
> approved these changes

The test_ssh_ping CI failure has been root-caused as an SELinux issue unrelated to this PR. The OpenSSH 9.8+ split of sshd into sshd/sshd-session means SSH forwarding now runs as sshd_session_t, which lacked connectto on container_runtime_t sockets. Tracked and fixed in containers/container-selinux#478 and containers/container-selinux#479.

Since the failure is confirmed environmental and independent of these changes, would you prefer to merge this PR now, or wait for the policy fix to land on the Testing Farm runners?

inknos:
ok, since we are gating podman-py again before the next release it can be merged

oligiochi:
> ok, since we are gating podman-py again before the next release it can be merged

thank you so much

inknos:
> thank you so much

sure and thanks a lot for your contribution!

Issue #654, opened 2026-09-01
from_env() ignores CONTAINER_CONNECTION and silently uses the local socket


michaelm-devel:
### Summary

`from_env()` does not consult `CONTAINER_CONNECTION`, which the `podman` CLI honours. The failure is silent: it connects to the local socket and returns entirely plausible data from the wrong machine.

### Reproducer

With a named connection to a remote host that has containers running, and a local socket that does not:

```
local socket, explicit                          ->  0 containers
explicit ssh client to the remote host          ->  7 containers
from_env(), CONTAINER_CONNECTION=<remote name>  ->  0 containers   <-- used the local socket
from_env(), CONTAINER_HOST=<remote uri>         ->  7 containers   <-- correct
```

Both machines ran podman 5.8.4, so the version string does not reveal the mistake either. `from_env()` reads `CONTAINER_HOST` / `DOCKER_HOST` but not `CONTAINER_CONNECTION`.

### Why it matters

The natural way to write a deployment tool is to reuse what the CLI uses:

```bash
export CONTAINER_CONNECTION=production
```

```python
client = podman.from_env()     # silently the local machine
```

A tool written this way builds, cleans up and reports success against the wrong host. There is no error to notice.

### Expected

Either `from_env()` resolves `CONTAINER_CONNECTION` through `PodmanConfig.services`, or the docstring states plainly that it does not and points at `PodmanClient(connection=...)`.

### Environment

podman-py 5.6.0 (Fedora `python3-podman-5.6.0-4.fc44`), podman 5.8.4 client and server, rootless, Python 3.14. Re-checked against current `main` before filing.

PR #656, opened 2026-09-21
fix: make service connection identity optional

### LINKED ISSUE #652: ServiceConnection.identity raises TypeError when a connection has no identity
### Summary

`ServiceConnection.identity` raises `TypeError` for any connection that has no identity configured — every `unix://` connection, and any `ssh://` connection added without `--identity`.

### Code (current `main`)

```python
@cached_property
def identity(self):
    """Path: Returns Path to identity file for service connection."""
    if self.attrs.get("identity"):
        return Path(self.attrs.get("identity"))
    return Path(self.attrs.get("Identity"))
```

When neither key is present this is `Path(None)`.

### Reproducer

Any host with a local `unix://` connection in `podman system connection list`:

```python
from podman.domain.config import PodmanConfig

cfg = PodmanConfig()
for name, svc in cfg.services.items():
    print(name, svc.identity)
```

```
TypeError: argument should be a str or an os.PathLike object where
__fspath__ returns a str, not 'NoneType'
```

Merely enumerating the configured connections is enough to hit it — the caller does not have to be doing anything unusual.

Same result with a hand-written `containers.conf` containing an ssh destination with no `identity =` line, so it is not specific to unix sockets.

### Expected

`identity` returns `None` when no identity is configured, and callers treat it as optional.

### Environment

podman-py 5.6.0 (Fedora `python3-podman-5.6.0-4.fc44`), podman 5.8.4 client and server, rootless, Python 3.14. Code path re-checked against current `main` before filing.

### Note

This also breaks `PodmanClient(connection=...)`; filed separately since the fix there is different.

### LINKED ISSUE #653: PodmanClient(connection=...) unusable without an identity; explicit identity= is ignored
### Summary

`PodmanClient(connection=...)` cannot be used with a connection that has no identity, and passing an explicit `identity=` does **not** work around it, because the fallback is evaluated eagerly.

### Code (current `main`, `client.py`)

```python
api_kwargs["identity"] = kwargs.get("identity", str(connection.identity))
```

`str(connection.identity)` is a default *argument*, so Python evaluates it before `kwargs.get` runs — even when the caller supplied `identity`. With no identity configured, `connection.identity` is `Path(None)` and raises (see the separate report on `ServiceConnection.identity`).

The same pattern is used on the `active_service` path, so a bare `PodmanClient()` fails on a system whose active connection has no identity.

### Reproducer

```python

PodmanClient(connection="remote-with-key")                      # OK
PodmanClient(connection="local")                                # TypeError
PodmanClient(connection="local", identity="/dev/null")          # TypeError  <-- explicit identity ignored
```

Measured against a host with a `unix://` connection named `local`.

### Expected

The explicit `identity=` is honoured, and a connection without one simply omits the argument.

### Suggested fix

Compute the fallback lazily:

```python
identity = kwargs.get("identity") or connection.attrs.get("Identity") or connection.attrs.get("identity")
if identity:
    api_kwargs["identity"] = identity
```

### Environment

podman-py 5.6.0 (Fedora `python3-podman-5.6.0-4.fc44`), podman 5.8.4 client and server, rootless, Python 3.14. Re-checked against current `main` before filing.

pratham8499:
Before the fix:
a) PodmanConfig().services['local'].identity
  -> failed: TypeError expected str, bytes or os.PathLike object, not NoneType
b) PodmanClient(connection='local')
  -> failed: TypeError expected str, bytes or os.PathLike object, not NoneType
c) PodmanClient(connection='local', identity='<dummy>')
  -> failed: TypeError expected str, bytes or os.PathLike object, not NoneType
d) PodmanClient(connection='production')
  -> success

After the fix:
a) PodmanConfig().services['local'].identity
  -> success: None
b) PodmanClient(connection='local')
  -> success
c) PodmanClient(connection='local', identity='<dummy>')
  -> success
d) PodmanClient(connection='production')
  -> success

The explicit `identity=` argument was ignored because the default fallback `str(connection.identity)` was evaluated eagerly, raising a TypeError before the fallback was even used. This fix lazily computes the fallback.

Fixes: #652
Fixes: #653
Signed-off-by: Pratham <patidarpratham40@gmail.com>

#654 (from_env ignoring CONTAINER_CONNECTION) is a separate fix and depends on this one; happy to follow up.


podman/domain/config.py
```diff
@@ -54,10 +54,12 @@ def url(self):
 
     @cached_property
     def identity(self):
-        """Path: Returns Path to identity file for service connection."""
+        """Optional[Path]: Returns Path to identity file for service connection."""
         if self.attrs.get("identity"):
             return Path(self.attrs.get("identity"))
-        return Path(self.attrs.get("Identity"))
+        if self.attrs.get("Identity"):
```
Honny1, line 60: What about `identity=""` case?

podman/client.py
```diff
@@ -68,13 +68,21 @@ def __init__(self, **kwargs) -> None:
             api_kwargs["base_url"] = connection.url.geturl()
 
             # Override configured identity, if provided in arguments
-            api_kwargs["identity"] = kwargs.get("identity", str(connection.identity))
+            identity = kwargs.get("identity") or connection.identity
+            if identity:
+                api_kwargs["identity"] = str(identity)
+            elif "identity" in api_kwargs:
+                del api_kwargs["identity"]
```
Honny1, line 75: Well, I see this part as uplicate with rows 81-85. Maybe put that into a separate function?

podman/domain/config.py
```diff
@@ -54,10 +54,12 @@ def url(self):
 
     @cached_property
     def identity(self):
```
Honny1, line 56:
```suggestion
    def identity(self) -> Optional[Path]:
```

Honny1 (commented):
Thanks, I have some comments.

Honny1 (changes requested):
Also linter complains:

```
 F841 Local variable `client` is assigned to but never used
     --> podman/tests/unit/test_podmanclient.py:101:60
      |
   99 |     def test_connect_no_identity(self, mock_api_init, mock_api_close):
  100 |         with mock.patch.multiple(Path, open=self.mocked_open, exists=MagicMock(return_value=True)):
  101 |             with PodmanClient(connection="no_identity") as client:
      |                                                            ^^^^^^
  102 |                 mock_api_init.assert_called_once()
  103 |                 kwargs = mock_api_init.call_args[1]
      |
  help: Remove assignment to unused variable `client`
  
  F841 Local variable `client` is assigned to but never used
     --> podman/tests/unit/test_podmanclient.py:111:84
      |
  109 |     def test_connect_explicit_identity(self, mock_api_init, mock_api_close):
  110 |         with mock.patch.multiple(Path, open=self.mocked_open, exists=MagicMock(return_value=True)):
  111 |             with PodmanClient(connection="no_identity", identity="/custom/key") as client:
      |                                                                                    ^^^^^^
  112 |                 mock_api_init.assert_called_once()
  113 |                 kwargs = mock_api_init.call_args[1]
      |
  help: Remove assignment to unused variable `client`
  
  F841 Local variable `client` is assigned to but never used
     --> podman/tests/unit/test_podmanclient.py:120:81
      |
  118 |     def test_connect_explicit_identity_override(self, mock_api_init, mock_api_close):
  119 |         with mock.patch.multiple(Path, open=self.mocked_open, exists=MagicMock(return_value=True)):
  120 |             with PodmanClient(connection="testing", identity="/custom/key2") as client:
      |                                                                                 ^^^^^^
  121 |                 mock_api_init.assert_called_once()
  122 |                 kwargs = mock_api_init.call_args[1]
      |
  help: Remove assignment to unused variable `client`
  
  F841 Local variable `client` is assigned to but never used
     --> podman/tests/unit/test_podmanclient.py:136:36
      |
  135 |         with mock.patch('podman.client.PodmanConfig', return_value=mock_config):
  136 |             with PodmanClient() as client:
      |                                    ^^^^^^
  137 |                 mock_api_init.assert_called_once()
  138 |                 kwargs = mock_api_init.call_args[1]
      |
  help: Remove assignment to unused variable `client`
  
  E501 Line too long (147 > 100)
     --> podman/tests/unit/test_podmanclient.py:154:101
      |
  152 | … / "podman.sock")
  153 | …
  154 | …l(), expected.replace('%5C', '\\') if '\\' in client.api.base_url.geturl() else expected)
      |                                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  155 | …
  156 | …or a user
      |
  
  Found 5 errors.
  No fixes available (4 hidden fixes can be enabled with the `--unsafe-fixes` option).
  
  ruff format..............................................................Failed
```

podman/domain/config.py
pratham8499: When the config has identity="", self.attrs.get("identity") returns an empty string "", which is False in python. It falls through the if condition and safely returns None. This correctly states that no identity file is given.

PR #657, opened 2026-09-22
fix: preserve list-valued filters

### LINKED ISSUE #542: `ContainersManager.list()` with label list filter doesn't perform filtering correctly
(not very familiar with podman or even python, feel free to close this if this is just a big misunderstanding)

Using package version `5.4.0.1`

### Expected Behavior
According to [the documentation](https://podman-py.readthedocs.io/en/latest/podman.domain.containers_manager.html#podman.domain.containers_manager.ContainersManager.list), `ContainersManager.list()` should support a `filters` keyword argument where you can filter on multiple labels, like so:
```python
filters = {
    "label": [
        "my.example.label=value1",
        "my.other.example.label=value2"
    ]
}

containers_manager.list(filters=filters)
```

Running this I'd expect to get a list of all containers that have the specified tags.

### Observed Behaviour
Listing containers with such a filter on my local environment returned no containers. But when I manually sent an HTTP request to the socket I got the expected result with the expected containers.

I can provide a more thorough example or logs if needed, just let me know.

### Theory?
After some debugging it seems like the function `_format_dict` in `api/http_utils.py` is a bit too eager to convert everything into a string. ([permalink](https://github.com/containers/podman-py/blob/2a29132efab4a183580a30d9c18b9bdb61b5ab49/podman/api/http_utils.py#L46))
```python
# in api/http_utils.py

def _format_dict(filters, criteria):
    for key, value in filters.items():
        if value is None:
            continue
        str_value = str(value) # list turns into a string here

        if key in criteria:
            criteria[key].append(str_value)
        else:
            criteria[key] = [str_value]
```

This then leads to a request being with a label filter with only one label, and that label is the string representation of the list of labels I provided to the filter. This is obviously only a problem with a label list, which makes filtering on a single label work fine.

Shubham-Padkonde:
Fixes #542: dictionaries containing list-valued filters now encode each element as a separate filter value rather than serializing the entire Python list as one string. Scalar conversion remains unchanged; empty lists and None values contribute no criteria.

Tests cover multiple labels, mixed scalar/list filters, numeric elements, empty/None entries, and the actual ContainersManager.list HTTP query without mutating the caller's filter dictionary.

Validation:
- Seven formatter subcases and the HTTP regression fail before and pass after locally.
- Hosted Fedora 44 unittest-coverage check passes: 257 tests run, one skipped, 85% reported total coverage. [Test log](https://artifacts.dev.testing-farm.io/5273dd77-2174-478d-ae29-e2b52e157542/work-unittest_coverage1nvsftqf/plans/distro/unittest_coverage/execute/data/guest/default-0/tests/coverage_unittest-1/output.txt).
- Local Windows/Python 3.13 broader run: 257 passed, 2 skipped, 13 failed; original production code has the same 13 failures. test_utils.py collection needs /etc/os-release and a Podman executable and was excluded from that local comparison.
- Ruff lint/format, targeted mypy and git diff --check pass locally.

Remaining hosted checks are not all green: distro-sanity errors before lint with `/bin/sh: - : invalid option`; Fedora 43 and rawhide RPM builds failed (build logs could not yet be retrieved), and several distro integration checks are pending. The passing Linux unit/coverage check resolves the original local validation blocker; these remaining checks still need investigation before merge.

Prepared with OpenAI Codex assistance. No local Podman daemon or container data was modified.


Shubham-Padkonde:
Fixed the test-case type inference reported by the sanity mypy job in 177df45. The explicit case annotation preserves all test inputs and runtime behavior.

Validation: `tox -e py -- podman/tests/unit/test_api_utils.py -q` passed 13 tests and 7 subtests; Ruff formatting and lint passed; `mypy --platform linux --package podman` passed for all 85 source files. The type check was run on Windows with the Linux target because native Windows checking reports the existing POSIX-only os.getuid/geteuid attributes.

Prepared with OpenAI Codex assistance; the contributor personally signed off this commit.

podman/api/http_utils.py
```diff
@@ -45,12 +45,10 @@ def _format_dict(filters, criteria):
     for key, value in filters.items():
         if value is None:
             continue
-        str_value = str(value)
-
-        if key in criteria:
-            criteria[key].append(str_value)
-        else:
-            criteria[key] = [str_value]
+        values = value if isinstance(value, list) else [value]
```
Honny1, line 48: What if the value is a tuple?

Honny1 (commented):
Please sqoush comiints in to one. I have just one qestion about type. 


PTAL @timcoding1988

